Issue #409, opened 2026-02-07
fix: wire up video playback in demo gallery


yonatangross:
## Problem

The demo gallery at `orchestkit.vercel.app/docs/reference` shows cards with play button overlays, but clicking them only opens a modal with a **static PNG thumbnail** — no video ever plays.

## Root Cause

Two disconnected pieces:
1. `orchestkit-demos/out/cdn-urls.json` has `videoCdn` URLs for 9+ compositions
2. The `Composition` interface and `demo-gallery.tsx` had no video fields or `<video>` element

## Fix

1. Added `thumbnailCdn?` and `videoCdn?` fields to `Composition` interface
2. Updated `generate-playground-data.js` to read `cdn-urls.json` and merge CDN URLs
3. Added `media-src 'self' https://cdn.sanity.io` to CSP headers
4. Replaced static `<img>` in modal with `<video>` when `videoCdn` exists (autoPlay, controls, playsInline)
5. Falls back to thumbnail with "Video coming soon" badge when no video
6. Updated GalleryCard and homepage thumbnails to use CDN with local fallback

## Files Changed

- `scripts/generate-playground-data.js` — reads cdn-urls.json, emits CDN fields
- `docs/site/lib/playground-data.ts` — regenerated with CDN fields
- `docs/site/components/demo-gallery.tsx` — video player in modal, CDN thumbnails
- `docs/site/app/(home)/page.tsx` — CDN thumbnail fallback
- `docs/site/next.config.mjs` — media-src CSP directive

yonatangross:
Closed: Video playback wired up with CDN integration. thumbnailCdn/videoCdn fields, cdn-urls.json merge, autoPlay controls, fallback badges. Commit b07efd04. 7 new tests.

PR #411, opened 2026-02-07
fix(ci): YAML workflows, git-validator, mem0 dedup, pre-push hook — v6.0.3

### LINKED ISSUE #446: bug(hooks): pre-push run_with_timeout perl alarm corrupts exit codes on macOS
## Problem

The `run_with_timeout` function in `bin/git-hooks/pre-push` uses `perl -e "alarm N; exec @ARGV"` as a fallback timeout on macOS (no `timeout`/`gtimeout`). This produces **false FAILED results** for build verification, TypeScript checks, and unit tests.

### Root Cause

`perl -e "alarm 300; exec @ARGV" npm run build` sets SIGALRM on the perl process, then `exec` replaces it with npm. The pending alarm fires on the exec'd process, killing it with exit code 142 (128 + SIGALRM) even though the build succeeds in ~90 seconds.

### Additional Bugs Found

1. **`set -o pipefail` + `git diff` CHANGELOG check**: `git diff origin/main --name-only | grep -q "CHANGELOG.md"` — git diff exits 1 when differences exist, pipefail propagates that through the pipeline even though grep found the match. Fix: capture output to variable first.

2. **Background process watchdog approach also fails**: Replacing perl alarm with `"$@" &; wait $pid` still produces incorrect exit codes for npm commands on macOS.

### Impact

- Developers on macOS cannot `git push` without `--no-verify`
- All pre-push validation checks using `run_with_timeout` are affected
- False failures hide real build/test issues

## Fix Applied

PR #411 (commit 74d9bbc3):
1. `run_with_timeout` fallback simplified to run commands directly (no timeout) on macOS — timeout protection only matters on CI where Linux `timeout` is available
2. CHANGELOG git diff check uses variable capture with `|| true` to avoid pipefail interaction
3. `/dev/null` stdin redirect to prevent git push ref data from interfering

## Remaining Work

- [ ] Consider installing `coreutils` (gtimeout) as a brew dependency for macOS development
- [ ] Add macOS-specific pre-push hook tests to CI matrix

yonatangross:
## Summary

- **Quote 4 workflow names** with colons (`Validate: Plugins`, `Validate: Version`, `Eval: Agent Routing`, `Visualize: Memory`) — unquoted colons caused YAML parse errors, resulting in 0 jobs
- **Fix `test-git-enforcement-hooks.sh`** on main by adding `ORCHESTKIT_BRANCH` env override so git-validator hook doesn't block test commits on protected branches
- **Rewrite `test-hook-chains.sh`** to read from committed `hooks.json` instead of gitignored `settings.json`
- **Fix mem0 batch test flakiness** — add `--no-infer` flag to `add-memory.py` that passes `infer=False` to mem0's `client.add()`, disabling semantic dedup so batch counts are deterministic across parallel CI runners (#416 related)
- **Fix pre-push hook bugs** (#446):
  - `run_with_timeout` perl alarm fallback corrupts exit codes on macOS — simplified to run directly (timeout only needed on CI Linux)
  - `set -o pipefail` + `git diff` CHANGELOG check — git diff exits 1 when diffs exist, breaking the pipeline. Fixed via variable capture with `|| true`
- **Version bump** 6.0.2 → 6.0.3
- **Wire video playback** in demo gallery with CDN integration (#409)

## Test plan

- [x] All 4 previously-broken workflows now parse correctly and run jobs
- [x] `test-git-enforcement-hooks.sh` passes on main (26/26 tests)
- [x] `test-hook-chains.sh` passes without `.claude/settings.json`
- [x] Version shows 6.0.3 across all config files
- [x] `npm test` — all 36 tests pass
- [ ] Mem0 batch tests (Test 6 + Test 18) pass with `--no-infer` on all 3 CI platforms
- [x] Pre-push hook passes all checks on macOS (version, CHANGELOG, build, imports, unit tests, security)

Fixes #446


yonatangross:
## Docs Site Quality Fixes (6.2 → 8+ score)

Added 5 fixes in commit d53ffab2:

1. **Bundle split** — Split 886KB `playground-data.ts` monolith into 6 per-domain files. Pages save ~520KB JS.
2. **fd-* design tokens** — Replaced hardcoded `gray-*/teal-*` + `dark:` prefixes with Fumadocs fd-* tokens across 4 components.
3. **WCAG AA contrast** — Replaced `#64748b` (4.03:1 ratio) with `fd-muted-foreground` (#94a3b8, 7.49:1) in 7 locations.
4. **next/image** — Converted raw `` to `next/image` via `OptimizedThumbnail` component (AVIF/WebP, srcset, error fallback).
5. **HC Mode focus** — Fixed `focus-visible` for Windows High Contrast Mode with transparent outline.

**Verification:** 115/115 tests pass, Next.js build compiles 309 pages.

PR #1840, opened 2026-05-16
fix(skills,#3224): codify manual-worktree pattern in /ork:cover (Path B)

yonatangross:
Closed — fix attempt was at the wrong layer of the orchestkit build system.

## What I tried

Updated `plugins/ork/skills/cover/SKILL.md` Phase 3 with a warning + manual-worktree pattern (mirroring what `/ork:implement` SKILL.md already does), plus a copy of the reference doc into `plugins/ork/skills/cover/references/`.

## Why CI broke

`bash scripts/build-plugins.sh` is the source-of-truth: it copies `src/skills/*/` → `plugins/ork/skills/*/` and treats anything in `plugins/` without a `src/` source as drift to be removed. My edits got reverted by the build.

## Adjacent observation

`manual-worktree-pattern.md` exists ONLY in `plugins/ork/skills/implement/references/` — no `src/` source. It's an orphan that survives builds by accident. Cross-skill use would need either promotion to a shared/ location OR inline duplication of the warning.

## Recommendations for next attempt

- **A** — Promote the reference doc to a shared/ location and have both skills reference it
- **B** — Drop the reference doc entirely; inline the warning + pattern in each SKILL.md (~50 LOC duplication, no build coupling)

B is the simpler path.


yonatangross:
Closing — fix attempt was at the wrong layer.

## What I tried

Added a manual-worktree warning + pattern to `plugins/ork/skills/cover/SKILL.md` Phase 3, mirroring what `/ork:implement` SKILL.md already has. Copied the reference doc from implement/ to cover/.

## Why CI broke

`bash scripts/build-plugins.sh` is the source-of-truth: it copies `src/skills/*/` → `plugins/ork/skills/*/`. My edits to plugins/ got reverted by the build because the source is in src/.

## Adjacent observation

The reference doc `manual-worktree-pattern.md` exists only in `plugins/ork/skills/implement/references/` — no `src/` source. It's an orphan that has survived builds accidentally.

## Recommended next attempts

- **A** — Promote the reference doc to a shared/ location
- **B** — Drop the reference doc; inline the warning + pattern in each SKILL.md (~50 LOC duplication, no build coupling). Simpler.

Either way: edit `src/` not `plugins/`.

Issue #627, opened 2026-02-16
feat(implement): add real-service integration testing phase (docker-compose/testcontainers)


yonatangross:
## Problem

When implementing features that depend on real services (Postgres, Redis, S3, external APIs), the implement skill defaults to mocks-only testing. There is no phase that:
- Detects `docker-compose.yml` or `testcontainers` in the project
- Recommends spinning up real services for integration tests
- Validates that the feature works against actual databases/queues/caches

### Files to Edit

**`src/skills/implement/SKILL.md`**
- Phase 5 (Test): Add service detection logic:
  ```
  # Auto-detect real service testing capability
  Glob(pattern="**/docker-compose*.yml")
  Glob(pattern="**/testcontainers*")
  Grep(pattern="testcontainers|docker-compose", glob="requirements*.txt")
  ```
- Phase 5: If services detected, add integration test requirements:
  - Spawn test-generator with `--real-services` context
  - Reference `testing-patterns` rules: `integration-database`, `integration-api`, `data-seeding-cleanup`
- Phase 6 (Integration): Rename from "Code Review" to "Integration Verification"
  - Add real-service test execution step
  - Add docker-compose up/testcontainers spin-up before test run
- Phase 8 (E2E): Add concrete steps beyond "Browser testing with agent-browser"

**`src/skills/brainstorming/references/phase-workflow.md`**
- Phase 3 (Feasibility): Add infrastructure detection question:
  - "Does this design require real service testing? If so, what's the docker-compose/testcontainers setup?"
- Phase 5 (Synthesis): Include infrastructure requirements per approach

### Acceptance Criteria
- [ ] Phase 5 auto-detects docker-compose.yml and testcontainers
- [ ] If detected, integration tests run against real services (not just mocks)
- [ ] Phase 6 renamed to "Integration Verification" with real-service steps
- [ ] Phase 8 E2E has concrete agent-browser steps
- [ ] Brainstorming feasibility check includes infrastructure detection

### Related Issues
- #624 (enforce test generation)
- #625 (wire testing-patterns rules)
- #626 (brainstorming testability)

yonatangross:
Implemented in PR #630: Real-service detection (docker-compose/testcontainers) in implement Phase 6, renamed to Integration Verification, new e2e-verification.md reference with concrete agent-browser + API testing steps.

PR #2527, opened 2026-06-19
docs(ci-debug): add pattern #11 (vercel-pending-orphaned-on-path-skip)

yonatangross:
## What

Adds **Pattern #11 — Vercel status orphaned (path-skip)** to the `ci-debug` playbook.

## Why

Surfaced live while triaging a real PR (kravi-tech-network #627): a **docs-only** PR
had every GitHub Actions check green, but the `Vercel` **commit status** (note: a
legacy commit-status, not a check-run) was stuck `pending` with `created_at == updated_at`
and never posted a terminal status. Root cause: Vercel posted the `pending`, then the
project's path filter **skipped** the build (the change never touched `apps/web`),
orphaning the status. `mergeStateStatus=UNSTABLE` + `mergeable=MERGEABLE` on an
unprotected base branch ⇒ cosmetic, safe to merge.

This failure mode is invisible to the existing 10 patterns because they all key off a
**failing** job — here there is none. So this PR also adds a short reachability note to
step 1 ("No job in the `fail` bucket but a check won't settle?") that routes the
no-failing-job / stuck-pending case directly to Pattern #11, classifying on
commit-status metadata instead of a (nonexistent) failing log.

## Changes

- New table row #11 with a falsifiable signature + exact safe-merge / permanent-fix commands.
- Reachability note in §1 for the stuck-pending case.
- Bumped "10-pattern" → "11-pattern" in the description, intro, and NOVEL template.

## Notes

- Docs-only change to `src/skills/ci-debug/SKILL.md`. The generated docs-site data
  (`docs/site/lib/generated/skill-content-data.ts`) and the published plugin copy are
  derived artifacts regenerated by the normal build/release pipeline — not hand-edited here.
- `/ork:ci-sentinel` picks up new patterns automatically on its next sweep (per the skill's
  own "Adding patterns" contract), so no separate plumbing is needed.


yonatangross:
Closing — superseded by #2528 (major: skill duplication / drift-gated build). This PR is the live repro of that bug: a single-file Contents-API edit to `src/skills/ci-debug/SKILL.md` trips the Build drift gate because `plugins/ork/{skills,commands}/ci-debug.*` weren't regenerated. The Vercel-orphan pattern (#11) is folded into #2528 and will land properly once the dup story is decided there.

PR #2966, opened 2026-07-17
fix(audit): correct silently-wrong line counts in audit scripts

yonatangross:
## Summary

Two of our audit scripts were reporting counts that were silently wrong. Both are fixed, with a rule file so the next counting pipeline starts correct.

Found while assessing a third-party plugin (`MaceTenth/token-efficient-scripts`). Its guardrail text pointed at the pattern; the bugs are ours, not theirs. The plugin itself scored 5.16/D/FAIL and is **not** being installed.

## The estimator never reported a total

`estimate-tokens.sh` exists to answer one question: does this codebase fit in the 1M context window. It answered `TOTAL 0 tokens` and `1M context: YES - tiny footprint` for **every** input, and exited 0 while doing it.

`count_lines()` wrote its table row and its return value to the same stdout, so the caller did arithmetic on the row text:

```
line 64: arithmetic syntax error: operand expected (error token is ".json  43 lines  x5  =  215 tokens
  TOTAL                                           0 tokens
  1M context:   YES - tiny footprint (< 16% of budget)
```

Split the concerns: `count_lines` returns only a number, `add_type` prints the row and folds tokens into `TOTAL`.

## The counting recipe was wrong twice over

`xargs wc -l | tail -1` keeps only the **final batch's** total once xargs splits a long argument list, and `wc -l` counts newlines so it under-reports any file lacking a trailing newline.

Measured on this repo, recipe in isolation:

| ext | recipe reported | actual | error |
|---|---:|---:|---:|
| `.md` | 40,267 | 756,671 | **-94.7%** |
| `.json` | 114,163 | 483,611 | **-76.4%** |
| `.ts` | 863,786 | 1,605,521 | **-46.2%** |

Replaced with `-exec grep -ch '' {} + | awk '{s+=$1} END{print s+0}'`, which emits one count per file and sums every batch. Deliberately **not** `xargs -0 cat | wc -l`: concatenating files whose last line lacks a newline glues boundary lines together.

## The 500-line gate was off by one

`run-audit.sh` had the same `wc -l` defect. 132 `.md` files in this repo lack a trailing newline, so each got a silent one-line discount and a 501-line SKILL.md could pass as 500.

## Verification

| check | before | after |
|---|---|---|
| `estimate-tokens.sh` on `audit-full/` | `TOTAL 0` + 2 arithmetic errors | `TOTAL 8,597 tokens`, clean |
| 501-line file, no trailing newline | `wc -l`=500 -> **PASS** (wrong) | `grep -c ''`=501 -> **FAIL** |
| empty file list | n/a | returns `0`, no hang |
| rule validation | - | 0 failed |
| agent structure | - | 0 offenses |

## Playground

`docs/shell-count-correctness/index.html` - interactive, two tabs. **Adoption & impact** has before/after code for all five traps with the command that proves each. **Verdict** scores the source plugin across 7 dimensions (drag any bar to re-weight).

## Notes

- `src/` and generated `plugins/` staged together per CLAUDE.md.
- Unrelated `docs/site/lib/generated/changelog-data.ts` build churn reverted.


yonatangross:
**On the red `Eval changed skills` check — it is pre-existing, not from this PR.**

`run-skill-eval.sh:85` selects skills from any changed path under `src/skills/<name>/`, but the quality grader reads only `SKILL.md`. This PR touches `audit-skills/scripts/run-audit.sh`; `audit-skills/SKILL.md` is byte-identical to `origin/main`, so the 22% quality score is main's content being surfaced, not a regression here.

It is also non-blocking: `skill-eval.yml` is not in `ci-summary.needs` and not in branch-protection required contexts.

Filed as #2967 (affects 43 of 115 skills that ship scripts). Deliberately not fixing `audit-skills/SKILL.md` prose in a shell-counting bugfix.

Issue #3129, opened 2026-07-24
fix(hooks): spawn-log name-shadowing, write subagent_type and agent_name as separate fields


yonatangross:
Schema drift in .claude/logs/subagent-spawns.jsonl: rows with source:"start" record the custom agent NAME in the subagent_type field (e.g. a named Explore agent logs its name, not "Explore"). Result: 20.7% of window spawns are unclassifiable and every catalog metric undercounts.

Work:
- Find the SubagentStart telemetry writer in src/hooks/src
- Write BOTH fields: subagent_type (real type) and agent_name (custom name); never overwrite type with name
- Classification treats legacy name-rows as unknown-type

Part of M170. Related prior art: #3034 / #3035 (agent-usage.jsonl constant fields + phantom rows).

Issue #1826, opened 2026-05-14
bug: #1250 recurrence — envelope leak now reaches EnterWorktree's name path (safe-fs defense incomplete)


yonatangross:
## Summary

`#1250` recurrence: the `{"continue":true,"suppressOutput":true}` envelope leak is back, this time biting `EnterWorktree`. The `safe-fs.ts` defense added by #1250 catches the mkdir/writeFile class but doesn't cover the EnterWorktree input path.

## Evidence (incident 2026-05-13)

A single background CC session at 10:50 IDT created literal-named directories in **5 separate repos** simultaneously:

```
/Users/yonatangross/coding/yonatan-hq/platform/{"continue":true,"suppressOutput":true}/.ruff_cache/
/Users/yonatangross/coding/yonatan-hq/ventures/trading-ai-analyst/{"continue":true,"suppressOutput":true}/.ruff_cache/
/Users/yonatangross/coding/yonatan-hq/ventures/jobscraper/{"continue":true,"suppressOutput":true}/.ruff_cache/
/Users/yonatangross/coding/yonatan-hq/ventures/realconnect/{"continue":true,"suppressOutput":true}/.ruff_cache/
/Users/yonatangross/coding/yonatan-hq/hq-ext-plugin/{"continue":true,"suppressOutput":true}/.ruff_cache/
```

Plus **191 corrupt session-summary files** in `~/.claude/hooks/sessions/` (e.g. `2026-03-09-17-{"continue":true,"suppressOutput":true}.md`) — older incidents that #1250's fix didn't backfill-clean.

## The `EnterWorktree` repro

When I called `EnterWorktree(name="m167-1-wire-memory-hook")` mid-session, the tool returned:

```
Path "/Users/yonatangross/coding/yonatan-hq/platform/{"continue":true,"suppressOutput":true}/{"continue":true,"suppressOutput":true}" does not exist
```

The envelope literal appeared TWICE in the constructed path. After calling `ExitWorktree`:

```
Exited worktree. Your work is preserved at {"continue":true,"suppressOutput":true}.
```

So `EnterWorktree`'s session-state cache had the envelope string as the worktree name. The tool was reading it from somewhere upstream — a hook output capture path that `#1250`'s `safe-fs.ts` doesn't cover.

## Coverage gaps in #1250's fix

`safe-fs.ts` rejects JSON-shaped strings at:
- `safeMkdirSync()`
- `safeWriteFileSync()`

[...]

PR #3136, opened 2026-07-24
fix: 3 defects found by /ork:assess on the M170 diff

yonatangross:
## Summary

Ran /ork:assess against the merged M170 diff (af4d6fdc8) as a post-merge quality check. 6-dimension parallel review (composite 7.17/10, grade B) surfaced 3 real defects, all independently re-verified against live code before accepting:

1. **Path traversal (LOW-MED)** -- `executor-route-nudge.ts` built a session-scoped flag file path from raw `session_id`. The sanitizer for this exact field already exists in the same lib (`safeIdentifier`, built for incident #1826) and was not applied. Fixed by routing through it, matching the sibling `getSessionTempDir` pattern.
2. **TOCTOU race** -- the once-per-session check (`existsSync` then `writeFileSync`) was not atomic. Fixed with `flag: 'wx'`.
3. **Cross-session name collision** -- `run-activation-audit.sh`'s name-to-type resolution map was keyed by bare `agent_name`, so two sessions reusing the same custom name could shadow each other's mapping -- the exact bug class #3129 exists to fix, reintroduced one level up. Fixed by keying on `session_id + name`.

## Test plan

- [x] 3 new hook tests: traversal session_id stays inside session storage, JSON-envelope session_id rejected without throwing, 5 concurrent calls fire the nudge exactly once
- [x] 2 new shell tests: same custom name in two sessions resolves independently in both directions
- [x] Full hooks suite: 7,356 pass / 0 fail
- [x] typecheck clean, security 17/17, root build clean

## Interactive Playground

**[Open Playground](https://htmlpreview.github.io/?https://github.com/yonatangross/orchestkit/blob/fix/assess-m170-findings/docs/fix--assess-m170-findings/index.html)** — interactive before/after for each of the 3 defects: path traversal, TOCTOU race, cross-session name collision.


yonatangross:
_(retrigger checks after body update)_

PR #3248, opened 2026-08-02
docs: fix nonexistent scripts/hooks path in SECURITY.md

latent-9:
SECURITY.md (line 71) points readers to `scripts/hooks/README.md`, but that file does not exist — there is no `scripts/hooks/` directory in the repo. The hooks README is at `src/hooks/README.md`, and the same sentence already references `src/hooks/`. This corrects the path from `scripts` to `src` so the reference resolves.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -68,7 +68,7 @@ Out of scope:
 
 Hooks are the highest-privilege surface here: they run shell or Node/Python on
 your machine on agent events. If you are evaluating OrchestKit for a team, read
-`src/hooks/` and `scripts/hooks/README.md` first. Every hook is designed to
+`src/hooks/` and `src/hooks/README.md` first. Every hook is designed to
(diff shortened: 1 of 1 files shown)


yonatangross:
Thanks for this, and welcome.

I verified the claim independently before merging rather than taking it on faith:

- `scripts/hooks/` does not exist anywhere in the tree
- `src/hooks/README.md` does exist
- a repo-wide grep confirms line 71 held the *only* remaining `scripts/hooks` reference, so this closes it out completely

Worth saying why this is more than a typo fix. The dead path sat inside SECURITY.md's "if you are evaluating OrchestKit for a team, read this first" sentence, which already named `src/hooks/` correctly three words earlier. So anyone doing security due diligence on the hook surface, the highest-privilege part of this plugin, was being pointed at a 404 by the security policy itself. Good catch.

CI is green. Merging.

PR #3293, opened 2026-08-05
fix(release): let release-please own the lockfile version

yonatangross:
A lockfile carries the version **twice**, at `$.version` and at `$.packages[''].version`. `.release-please-config.json` listed `package.json` under `extra-files` and never listed `package-lock.json`, so every release left both copies behind.

Measured at `origin/main` `1f4cf6125`:

| file | field | value |
|---|---|---|
| `package.json` | `$.version` | `9.7.0` |
| `package-lock.json` | `$.version` | **`9.5.2`** |
| `package-lock.json` | `$.packages[''].version` | **`9.5.2`** |

Two minor versions stale.

## Why this was worth fixing properly

The symptom is not cosmetic. `npm` rewrites those fields on the next install, so **every fresh clone and every new git worktree came up with a modified `package-lock.json` before anyone edited anything.**

A tree that is dirty on arrival teaches everyone to wave dirt through. This exact diff was reverted twice in a single session as "incidental churn" before anyone asked why it kept coming back. That is the real cost: not two stale numbers, but the habit of ignoring a modified file because it is always modified.

## Root cause, not symptom

Resyncing the lockfile alone would have lasted exactly one release. The config now governs both json paths, the same way it already did for `package.json`:

```json
{ "type": "json", "path": "package-lock.json", "jsonpath": "$.version" },
{ "type": "json", "path": "package-lock.json", "jsonpath": "$.packages[''].version" }
```

## Backstopped, and proven able to fail

`tests/manifests/test-version-consistency.mjs` asserts the three fields agree **and** that the two config entries still exist, so deleting the fix fails at the gate instead of quietly reopening the drift one release later.

| mutation | result |
|---|---|
| lockfile set back to `9.5.2` | exit 1, `$.version is 9.5.2, package.json is 9.7.0` |
| config entries removed | exit 1, `the drift will silently return` |
| both restored | green |

Wired into all three rosters: `run-all-tests.sh` and `ci.yml` glob `tests/manifests` automatically; the explicit `npm run test:manifests` list was updated.

**Verified end to end:** after committing, `npm install` leaves `package-lock.json` clean. A fresh worktree is no longer born dirty.

## Interactive Playground

Toggle between the before and after state of the three version fields, with the two mutation results.

**[Open Playground](https://htmlpreview.github.io/?https://github.com/yonatangross/orchestkit/blob/fix/lockfile-version-drift/docs/fix--lockfile-version-drift/playground.html)**

Source: `docs/fix--lockfile-version-drift/playground.html`


yonatangross:
Applying `release-please-override`, with the reasoning recorded rather than asserted.

**The catch-22:** the File Guard builds its governed list *from* `.release-please-config.json`. This PR adds `package-lock.json` to that list, so the same PR's one-time resync (`9.5.2` → `9.7.0`) is immediately read as a hand-bump of a governed field. Any PR that puts a lagging file under governance hits this.

**Why the override is safe here, checked not assumed:**

| check | result |
|---|---|
| release-please state source | `.release-please-manifest.json` = `9.7.0` |
| does `release-please.yml` read the lockfile? | no, never mentions it |
| direction of my change | `9.5.2` → `9.7.0`, i.e. **toward** the manifest state |

The guard protects against hand-bumps that push a governed file *past* release-please and desync its state machine (the #1946 incident). This does the reverse: it drags a file that fell two minors behind back **onto** the state release-please already believes it is at. After this lands, release-please owns both lockfile version fields and the drift cannot recur, backstopped by `tests/manifests/test-version-consistency.mjs`.

The alternative the guard suggests (revert the bump, let release-please carry it) would leave `package.json 9.7.0` disagreeing with `package-lock 9.5.2` on main until the next release, which is exactly the state the new test fails on.

Issue #3311, opened 2026-08-08
Turn on the authored-but-inert sandbox and add the 5 Read() credential denies


yonatangross:
VERIFIED. src/settings/ork.settings.json:77-125 declares failIfUnavailable, autoAllowBashIfSandboxed, filesystem allow/deny, 20 denied exfil domains and 5 credential-file denies. The block's keys were enumerated programmatically: 'enabled' is ABSENT, so all of it is inert. No other scope sets it. Meanwhile ~/.claude/settings.json carries skipDangerousModePermissionPrompt and skipAutoPermissionPrompt both true, and permissions.deny holds 21 rules of which ZERO are Read(...) rules, so nothing blocks a read of ~/.ssh/id_rsa. The substitute is network-egress-guard.ts:26-29 whose own header says a regex guard is bypassable, and 3 bypasses were reproduced 2026-08-06. FIX: add enabled:true; ALSO set it in user settings since manifests/ork.json has no settings key and plugins-reference says only agent and subagentStatusLine are read from a plugin settings.json. Confirm with claude --debug that the seatbelt profile loads BEFORE relaxing any regex guard.

yonatangross:
Artifact armed on the #3310/#3311 PR: `sandbox.enabled: true` + 5 `Read()` credential denies now in `src/settings/ork.settings.json`.

**Remaining, deliberately an operator step** (changes every Bash call on the machine): activation requires user/project settings, since CC reads only `agent`/`subagentStatusLine` from a plugin settings.json. To activate on a machine, add to `~/.claude/settings.json`:

```json
"sandbox": { "enabled": true }
```

plus the 5 Read() denies under `permissions.deny`, then confirm with `claude --debug` that the sandbox initializes. Keeping this issue open until activation is confirmed on at least the dev machine.

yonatangross:
## Both VERIFIED claims in this body are now false at origin/main

Re-derived 2026-08-09 against `origin/main` (24d5fb60f):

| body claims | actual at origin/main |
|---|---|
| "`enabled` is ABSENT, so all of it is inert" | `enabled` is **present and `true`** |
| "permissions.deny holds 21 rules of which ZERO are Read(...)" | **26** deny rules incl. **5** Read() denies |

`76581187b` (#3345) is the commit that changed both, 2026-08-08. Its own message scopes itself honestly: *"this arms the ARTIFACT, not any machine."*

## But the issue should stay OPEN, because the artifact half never mattered

[...]

Issue #219, opened 2026-01-24
perf: Complete TypeScript hook wiring migration (128 remaining hooks)


yonatangross:
## Summary

The TypeScript hook migration (v5.1.0) created all the compiled bundles but **only wired SessionStart hooks** to use `run-hook.mjs`. The remaining 128 hooks still invoke legacy `.sh` files directly.

## Current State

```
╔══════════════════════════════════════════════════════════════════════════╗
║                    HOOK WIRING STATUS (v5.1.5)                           ║
╠════════════════════╦═══════════════╦═══════════════╦═════════════════════╣
║    Hook Event      ║  Uses .mjs    ║  Uses .sh     ║  Status             ║
╠════════════════════╬═══════════════╬═══════════════╬═════════════════════╣
║ SessionStart       ║  ████████ 12  ║               ║  ✅ MIGRATED        ║
║ SessionEnd         ║               ║  █████ 5      ║  ⏳ Legacy          ║
║ PreToolUse         ║               ║  █████████ 21 ║  ⏳ Legacy          ║
║ PostToolUse        ║               ║  █████████ 25 ║  ⏳ Legacy          ║
║ PermissionRequest  ║               ║  ███ 3        ║  ⏳ Legacy          ║
║ UserPromptSubmit   ║               ║  ████████ 8   ║  ⏳ Legacy          ║
║ Stop               ║               ║  █████████ 26 ║  ⏳ Legacy          ║
║ SubagentStart      ║               ║  ████ 4       ║  ⏳ Legacy          ║
║ SubagentStop       ║               ║  █████████ 9  ║  ⏳ Legacy          ║
║ Setup              ║               ║  ██████ 6     ║  ⏳ Legacy          ║
║ Notification       ║               ║  ██ 2         ║  ⏳ Legacy          ║
╠════════════════════╬═══════════════╬═══════════════╬═════════════════════╣
║ TOTAL              ║      12       ║     128       ║  ~9% complete       ║
╚══════════════════════════════════════════════════════════════════════════╝
```

### Performance Impact
- **Bash startup overhead**: Each `.sh` invocation spawns a new bash process (~15-50ms)
- **TypeScript bundles**: Pre-compiled, loaded once per bundle type (~5-15ms)
- **Expected improvement**: 2-5x faster hook execution across all events

### Current Architecture
```
hooks/
├── dist/                    # ✅ Compiled TS bundles (381 KB total)
[...]

PR #3360, opened 2026-08-09
feat(hooks): deny credential-file reads at the Read tool

yonatangross:
## Summary

ork declared five `Read()` deny rules for `~/.aws/credentials`, `~/.ssh/**`, `~/.gnupg/**`, `~/.netrc`, `~/.npmrc`. **None of them has ever blocked anything**, for two independent reasons — both measured, not inferred:

```
settings rule   INERT    the plugin settings.json is discarded at load
                         (only agent + subagentStatusLine are honoured;
                          ork declares neither, so the object is empty
                          and CC throws the whole file away)

pretool/read/*  ABSENT   the only handler was tldr-summary, which abstains

probe {"tool_name":"Read","tool_input":{"file_path":"~/.ssh/id_rsa"}}
  -> rc=0, NO permissionDecision, from EVERY registered hook
```

This is **protection that never existed**, not protection removed.

## The guard

Denies a `Read` whose path resolves into a guarded credential location — after `~` expansion, after `..` collapsing, and after symlink resolution, so rewriting the path does not change the answer. When the target does not exist it falls back to the lexical path, so a guarded path still denies rather than sliding through on a missing file.

## The bypass found by attacking it

The first implementation compared case-sensitively: it denied `~/.ssh/id_rsa` and **allowed** `~/.SSH/id_rsa`. On this machine that is a **live** bypass, not theoretical — macOS APFS is case-insensitive by default, `[ -d ~/.SSH ]` is true, and the uppercase path opens the same bytes.

| path | before fix | after |
|---|---|---|
| `~/.ssh/id_rsa` | deny | deny |
| `~/.aws/credentials` · `~/.netrc` · `~/.gnupg/secring.gpg` | deny | deny |
| `/tmp/../Users/me/.ssh/id_rsa` (traversal) | deny | deny |
| `~/.ssh/../.ssh/id_rsa` | deny | deny |
| `~//.ssh//id_rsa` | deny | deny |
| `~/.ssh/subdir/../id_rsa` | deny | deny |
| **`~/.SSH/id_rsa`** | **ALLOW ← bypass** | **deny** |
| `~/.AWS/Credentials` · `~/.NetRC` | ALLOW | deny |
| `README.md` | allow | allow |
| `~/.sshconfig-notes.md` (prefix collision) | allow | allow |

`matchesRule` tries the **exact** comparison first, so a genuinely case-sensitive filesystem keeps precise matching, then retries folded. Folding only ever *widens* the deny. Denying a distinct `~/.SSH` that really is a separate directory on a case-sensitive volume is the safe direction to err.

## Registration

Registered in **both** `src/hooks/hooks.json` and `src/hooks/src/entries/pretool.ts`. One without the other is the #959 dead-hook class. Hook count moves **218 → 219** and the generated `plugins/` description is re-stamped to match.

```
build            rc=0
typecheck        rc=0
validate-counts  PASSED: All counts and versions match
```

Refs #3311.

## Interactive Playground

**[Open Playground](https://htmlpreview.github.io/?https://github.com/yonatangross/orchestkit/blob/fix/credential-read-guard/docs/fix--credential-read-guard/playground.html)**

`docs/fix--credential-read-guard/playground.html` — the bypass table and the two independent reasons those paths were unguarded.


yonatangross:
Applied `release-please-override`.

The File Guard flags two whole-file governed paths in this diff:

```
docs/site/lib/generated/plugins-data.ts
pyproject.toml
```

Both diffs are **only** the hook count, and neither touches a version field:

```diff
- description = "... 105 skills, 36 agents, and 218 hooks"
+ description = "... 105 skills, 36 agents, and 219 hooks"

-    "hooks": 218,
+    "hooks": 219,
```

Neither was hand-edited. `plugins-data.ts` is generated by `npm run build` and the `pyproject.toml` description is stamped by `scripts/stamp-counts.sh`; both moved because this PR adds hook #219 (`pretool/read/credential-read-guard`). `bin/validate-counts.sh` fails if they are left at 218, so reverting them is not an option — the count would be genuinely wrong.

This is the case the guard's own error text describes: a legitimate content change that is not a version bump. Release-please's state machine is untouched — `.release-please-manifest.json`, `CHANGELOG.md` and `version.txt` are not in this diff.

Issue #3353, opened 2026-08-08
settings_overrides/locks/worktree_links: unused features, not dead code — decide product fate


yonatangross:
Split from #3328 after HEAD re-verification. The audit said "0 rows ever → drop the tables". Wrong frame: settings-override-resolver.ts READS settings_overrides on EVERY PreToolUse, session-finalizer handles CASCADE on locks. 0 rows means the WRITE side never shipped a user — an unused feature with a live, per-tool-call read cost.

Decision needed per table: ship a writer (make the feature real), or remove reader+schema together as a designed removal. Either way the per-PreToolUse read cost should be measured first — if resolver overhead is ~0 rows fast-path, urgency is low.

Related: the same audit pattern (0 rows ≠ dead) almost deleted a live read path; see the #3328 comment table.

yonatangross:
## Analysis complete — decision memo. Recommendation: tombstone, in two independently-sized pieces

Method: read-only analysis agent over code, git history, and the live DB; I independently re-verified the four decisive claims before posting (the hooks.json catch-all pair, the 1,750/0 active-overrides census, worktree_links having only schema references, and the 3085/0/0/0 row counts).

### The issue's framing is half right

`settings_overrides` DOES have a live per-tool-call read path — but two of the three premises do not survive the code.

| table | production writers | production readers | rows ever |
|---|---|---|---|
| `worktree_links` | none | none (schema + 1 comment only; #3315 deleted the last would-be writer) | 0 |
| `locks` | none (INSERT exists only in a test) | GC only: `session-registrar.ts:190` deletes expired rows that can never exist | 0 |
| `settings_overrides` | none (setOverride called only from tests) | `pretool/settings-override-resolver.ts:53`, fully wired (hooks.json:273 + entries/pretool.ts:75,153) | 0 |

"0 ever" is structural, not circumstantial: no production INSERT exists for any of the three, and since nothing ever DELETEs a `sessions` row (3,085 accumulated), the CASCADE that could have hidden past rows has never fired. Correction to the issue text: `session-finalizer.ts:34` only does `UPDATE sessions SET status` — the CASCADE claim in its comment describes code that is not there.

### The decisive fact the issue did not have

[...]

PR #3366, opened 2026-08-09
fix(hooks): stop owning worktree provisioning, CC does it better

### LINKED ISSUE #3315: worktree-provisioner silently deletes .worktreeinclude and worktree.baseRef
VERIFIED, refuter-upheld. Registering the WorktreeCreate event makes CC skip its ENTIRE post-create provisioning suite. worktree-provisioner.ts:210 hardcodes (projectDir)/.worktrees/(name) and reimplements none of what the default did, so .worktreeinclude copying (gitignored config such as .env into every new worktree) and worktree.baseRef resolution are gone estate-wide, with no error. Two further findings ride on it: hook-provisioned worktrees are never removed by anyone because CC delegates removal to the hook, and provisioner-created worktrees are dirty from birth. FIX: return join(projectDir,'.claude','worktrees',name), resolve baseRef, copy .worktreeinclude matches. That one change also discharges two Epic D adoptions and makes sweep-stale-worktrees deletable rather than an rm -rf race.

yonatangross:
## Registering a WorktreeCreate hook does not extend provisioning, it replaces it

Read off the shipped 2.1.226 binary rather than the docs:

```js
function UHe(){                                  // hasWorktreeCreateHook
  if(Pp("hooks")) return !1;
  let e=Due()?.WorktreeCreate;  if(e&&e.length>0) return !0;
  if(!$x()){ let o=Kle()?.WorktreeCreate; if(o&&o.length>0) return !0 }
  let t=ine()?.WorktreeCreate;  if(!t||t.length===0) return !1;
  let r=$x(), n=r&&!Xc()?GCe():null;
  return t.some((o)=>!(r&&("pluginRoot"in o)&&!n?.has(o.pluginId)))   // plugin hooks COUNT
}
```

Its single call site takes the hook branch or the native branch, never both. There is no filter on `async`, and no check that the hook returns a path.

## What ork was standing in front of

ork reimplemented **2 of the 12** things the native branch does.

| Capability | CC native | ork |
|---|---|---|
| `worktree.sparsePaths` | yes | duplicate |
| path placement | yes | different path |
| `worktree.baseRef` | yes | **no** |
| `.worktreeinclude` | yes | **no** |
| `settings.local.json` propagation | yes | **no** |
| `core.hooksPath` to main repo | yes | **no** |
| `worktree.symlinkDirectories` | yes | **no** |
| PR worktrees (`pull/N/head`) | yes | **no** |
| resume / reset to fresh base | yes | **no** |
| realpath containment check | yes | **no** |
| `git worktree lock` | yes | **no** |
| branch deleted on cleanup | yes | **no** |

Four ork reference docs have been instructing operators to set `worktree.baseRef` the entire time it could not work for them.

CC's own error text names the intended audience: "Configure WorktreeCreate/WorktreeRemove hooks ... to use worktree isolation **with other VCS systems**." The event exists for non-git. ork is a git repo that used it to reimplement git.

## Removal was broken, not merely absent

CC dispatches `WorktreeRemove` only for hook-created worktrees, and refuses when the directory is non-empty, which a real worktree always is. The `git worktree remove` fallback lives in the `!hookBased` branch. So provisioned worktrees leaked.

Measured 2026-08-09:

```
worktree-events.jsonl ........ 11 lines all-time
  from a REAL worktree ....... 1     (2026-07-18, rows_updated: 0)
  from vitest temp dirs ...... 10    writing into the live project log
worktree-pending markers ..... 3 uncleaned
  2026-08-06 01:37, plus two from agents that finished minutes earlier
```

## The cascade that never ran

`exit-finalizer` selects and updates `worktree_links`. The only `INSERT INTO worktree_links` in the entire repo is in its own test at `__tests__/worktree/exit-finalizer.test.ts:53`. The UPDATE always matched zero rows, and the one real log line above records exactly that.

## What was deleted

The `WorktreeCreate` key, `worktree-provisioner`, `exit-finalizer`, `worktree-lifecycle-logger`, `lib/children-bus.ts`, `lib/monorepo-advisory.ts`.

`WorktreeCreate`'s `lifecycle/webhook-forwarder` entry went too, despite being pure async telemetry, because the gate does not care what a hook does. Leaving it would keep CC out of its own provisioning path with nothing left to provision, which is worse than either end state.

**The schema is deliberately untouched.** Editing a shipped migration does not affect databases that already ran it and would drift fresh installs from existing ones. `worktree_links` belongs to #3353, which now has no reader left to weigh.

## Tests

The pre-existing wiring assertions were **inverted, not deleted**, so re-registering anything on `WorktreeCreate` fails the suite.

```
validate-counts        219 -> 215, PASSED
build                  idempotent, second run leaves 0 files dirty
hooks suite            7489 passed
```

6 failures remain in `subagent-start/context-gate.test.ts`. They are **not** from this change: my diff has zero matches on that path, and the test reads live machine state (it asserted on "7 background agents active", which was this session's real agent count). Filed separately.

## Playground

`docs/fix--3315-retire-worktree-ownership/playground.html`

## Follow-ups

- `sweepPendingMarkers` in `lifecycle/sweep-stale-worktrees.ts` now sweeps markers nothing writes. Left in place; that file was owned by another change in flight.
- The monorepo `sparsePaths` advisory was correct and died with its only caller. Its natural home is doctor's operator-settings-posture check.
- Worktrees now land at CC's `.claude/worktrees/<name>` rather than `.worktrees/<name>`. The `.gitignore` entry and the operator worktree convention need a follow-up pass.

Closes #3315


yonatangross:
## release-please-override justification

The File Guard fired on two whole-file-governed paths. Both diffs are the count re-stamp this change is required to make, and neither touches a version.

```
pyproject.toml
  -description = "... 105 skills, 36 agents, and 219 hooks"
  +description = "... 105 skills, 36 agents, and 215 hooks"

docs/site/lib/generated/plugins-data.ts
  -"description": "... 105 skills, 36 agents, 219 hooks."
  +"description": "... 105 skills, 36 agents, 215 hooks."
  -"hooks": 219,
  +"hooks": 215,
```

Both files are **generated by `npm run build`**, not hand-edited. `bin/validate-counts.sh` fails the commit without them:

```
before:  Hooks: declared 219, actual 215  -> Validation FAILED
after:   Hooks: 215                       -> Validation PASSED
```

Version fields were checked explicitly and are untouched in every governed file:

```
manifests/ork.json            no "version" / "ref" change
.claude-plugin/marketplace.json  no "version" / "ref" change
```

`.claude/rules/count-sync.md` prescribes exactly this path: the guard is field-aware for `manifests/ork.json` (description counts pass unlabelled), but `pyproject.toml` and `plugins-data.ts` block on ANY edit, "for a legitimate content change there, use the `release-please-override` label."

PR #3377, opened 2026-08-09
fix(hooks): two events with no listener, one handoff with two

### LINKED ISSUE #3327: Wire the 2 unregistered CC hook events: DirectoryAdded and MessageDisplay
VERIFIED. CC 2.1.226 defines 31 hook events; ork registers 29. DirectoryAdded fires on /add-dir or SDK register_repo_root and ork materializes rules per-directory, so it is a free correctness win. MessageDisplay is the only surface that can rewrite displayed prose in flight.

### LINKED ISSUE #3329: Retire the duplicate across-time handoff loop
Epic C. Two loops both inject '[Previous Session Handoff]': lifecycle/session-handoff-injector.ts:72 (SessionEnd YAML replayed at SessionStart) and prompt/handoff-injector.ts:102 (Stop markdown replayed at first prompt). We diverged from ourselves. RE-HOME FIRST: teach prompt/handoff-injector.ts to fall back to the global store when .claude/HANDOFF.md is absent, ship it, and MEASURE the fallback firing. Only then delete the SessionStart injector and stop/handoff-writer.ts. Do not sell as fixed: hashProject() is sha256(projectDir) with no git-toplevel normalization, so a worktree and its parent hash to different buckets.

yonatangross:
## Two events with no listener, one handoff with two

The same shape from opposite ends: a message CC sent that nothing received, and a message ork sent twice.

| issue | the defect | the fix |
|---|---|---|
| #3327 | `DirectoryAdded` and `MessageDisplay` had **no entry in `hooks.json`**. CC has emitted them for releases; they arrived and went nowhere. | two observer-only hooks, `async` / `timeout 5` / metrics only |
| #3329 | Two injectors opened with the **identical marker** `[Previous Session Handoff]`, so a resumed session got the block twice. | `lib/handoff-claim.ts` — first claimer wins, second stands down |

## The stale blocker note, re-probed with a control

The registry carried a 2026-07-15 note: `claude plugin validate` rejects `hooks.MessageDisplay`, so the observer was reverted the same day. That note is **stale** and is corrected in place rather than deleted.

```
claude plugin validate, CC 2.1.226, against a copy of the built plugin
  hooks.MessageDisplay         ACCEPTED
  hooks.DirectoryAdded         ACCEPTED
  hooks.TotallyBogusEventZZZ   REJECTED  "Invalid key in record"
```

The third line is the point. A clean validation run proves the keys are legal only if the validator would still reject an illegal one — otherwise "no error" could just mean the check stopped running. The validator is live; these two keys are simply in its schema now.

## What this deliberately did NOT do

The issue proposed re-running rules materialization against the newly added directory. Skipped, with reasons: the payload field carrying the path is unverified against the binary, and `materializeAntipatternRules` writes byte-identical content to every target — so a second copy spends context tokens without correcting anything.

## The trap that makes #3327 look small

`tests/hooks/test-hook-structure.sh` holds a `VALID_EVENTS` allow-list that hard-fails on any event it does not know. Adding an event without extending it turns a green suite red. That is the gate working, and it is extended here.

Registration is also a **pair** — `hooks.json` AND `entries/<event>.ts`. One without the other is a hook that dispatches to nothing and never errors, the #959 class. Asserted for both:

```
directory-added           hooks.json=1  entries=2   BOTH
message-display-observer  hooks.json=1  entries=2   BOTH
```

## A lint error worth not suppressing

`directory-added.ts` sanitized paths with a character class holding C0 control escapes, which trips biome's `noControlCharactersInRegex`. Rewritten as an explicit code-point loop instead of suppressed — a control character inside a pattern is invisible in review, which is exactly how a sanitizer quietly stops covering a character someone believed it covered. Behaviour re-tested directly: NUL, ESC, DEL and newline stripped, ordinary paths untouched, 256-char cap exact even across a surrogate pair.

## Verification

```
typecheck            0
biome                0 errors
test-hook-structure  6/6
security             19/19
validate-counts      PASSED   (215 -> 217 hooks)
build                stable across consecutive runs, 40 files
```

Falsifiability — new tests run against `origin/main`'s source:

```
with the fix        11 passed
origin/main source   9 failed / 2 passed
restores            byte-identical
```

### On the 6 `context-gate.test.ts` failures

They are **not** from this work, and the check is worth showing because the obvious reading is wrong:

```
clean bccfff909  (pre-merge)    6 failed / 24 passed
clean origin/main (post-merge)  6 failed / 24 passed
primary worktree                    30 passed
```

The primary tree passes only because it carries 26 uncommitted files that happen to fix them. Comparing against a dirty tree reads as a regression that does not exist — and separately means a real context-gate fix exists locally and has never been shipped.

## Interactive Playground

**[Open Playground](https://htmlpreview.github.io/?https://github.com/yonatangross/orchestkit/blob/fix--v10-hook-events/docs/fix--v10-hook-events/events-and-echoes.html)** — `docs/fix--v10-hook-events/events-and-echoes.html`

The two events, a toggle showing what each half of a broken registration pair does at runtime (both halves fail silently), the duplicate-handoff cost, and a copy-prompt for auditing registration pairs in any plugin.

Closes #3327
Closes #3329


yonatangross:
Applied `release-please-override`.

Two governed files change here, both **build-generated count stamps only**, no version field touched:

```
pyproject.toml                        105 skills, 36 agents, 215 -> 217 hooks
docs/site/lib/generated/plugins-data  "hooks": 215 -> 217
```

Verified these come from `npm run build`, not a hand edit: reverting both to `origin/main` and rebuilding reproduces them exactly.

This is the standing collision between two required checks — the Build check fails if generated artifacts drift, the File Guard fails if governed files are touched, and any PR that changes the hook count trips both. Same resolution as #3360, #3366, #3340, #3301, #3293.

Issue #3321, opened 2026-08-08
Compaction layer: 4 hooks, none of which reach the model


yonatangross:
VERIFIED, 5 findings in one subsystem. post-compact-recovery's additionalContext never reaches the model because CC 2.1.226 PostCompact has no additionalContext support at all. pre-compact-task-done-prompt is gated on a counter that is structurally always ~0, so it fires ~never. The token accumulator every context-pressure heuristic reads never increments. context-exhaustion-warner has NO data source: the StatusLine bridge file it reads has zero writers. pre-compact-saver reads token-budget-state.json while the writer emits instruction-budget.json. Both PreCompact blockers set stopReason but CC reads `reason`, so the block lands with an empty message. FIX: adopt autoCompactWindow (Epic D) and delete what cannot fire.

yonatangross:
## Investigated before removing. The two hooks are not the same case.

Re-derived against `origin/main`. Two corrections to this issue's premise first, then a verdict per hook.

### Correction 1: "none of which reach the model" is false

`lifecycle/post-compact-recovery.ts:182` returns `outputWithContext(...)`, which emits `hookSpecificOutput.additionalContext`. `PostCompact` is in CC's own generated allow-list for that key (`src/hooks/bin/cc-output-keys.generated.mjs:22,42`, derived from the shipped binary), and `src/hooks/src/__tests__/lifecycle-output-protocol.test.ts:612-635` asserts CC consumes it. One of the four demonstrably reaches the model.

### Correction 2: `autoCompactWindow` is not the re-home target for the guard

#3325 describes `autoCompactWindow` as "the CC-native half of the problem the four-hook compaction layer works around, and adopting it is what makes deleting those hooks safe." That does not hold for `pre-compact-guard`. `autoCompactWindow` governs *when* auto-compaction triggers. It cannot express *"a background agent spawned under five minutes ago is still running."* Those are different propositions, so adopting it does not make deleting the guard safe. Separately, `autoCompactWindow` is an operator-scope settings key, which a plugin cannot set at all (see #3357).

---

### `pre-compact-guard` — KEEP

[...]

PR #3452, opened 2026-08-12
fix(hooks): stop pre-compact-guard blocking auto-compaction

### LINKED ISSUE #3451: pre-compact-guard blocks CC's automatic compaction at the context ceiling
VERIFIED, observed live 2026-08-12 in session `0de8dd40` at 100% context.

A sixth finding in the compaction subsystem of #3321, distinct from the five listed there. #3427 deleted the parts of that layer that could not fire and deliberately kept `lifecycle/pre-compact-guard`, which does fire. It fires wrongly.

## What happened

The session hit the context ceiling. CC printed `Context limit reached · /compact or /clear to continue`, and the guard blocked compaction:

```
Compaction blocked - 1 background agent(s) still running (ckleg2).
Wait for completion or set CLAUDE_CODE_ALLOW_COMPACT_DURING_AGENTS=1.
```

That leaves `/clear` as the only exit, which discards the session state the guard exists to protect. The guard is most harmful exactly where it matters least.

## Three defects

**1. It cannot tell auto from manual.** `preCompactGuard(_input, ctx)` takes the payload and never reads it, so CC's automatic compaction is treated identically to an explicit `/compact`. `trigger` (`'auto' | 'manual'`) has been typed and binary-verified in `src/hooks/src/types.ts:187` since 2.1.227, so the field was in the payload the whole time. This matters more since #3427: `autoCompactWindow` arms compaction at a chosen fill level, so auto-compaction now happens earlier and more often.

**2. The remedy in the block message is unreachable from where it is printed.** `CLAUDE_CODE_ALLOW_COMPACT_DURING_AGENTS` is read from the hook process env, inherited fixed at CC launch. A user reading that sentence mid-session cannot act on it without a restart, and the restart loses the same context.

**3. "Active" is recency, not liveness.** `.claude/logs/subagent-spawns.jsonl` is a spawn-INTENT log, written by `subagent-validator` (`source: 'start'`) and `pretool/task/spawn-intent-logger` (`source: 'pretool'`). Nothing anywhere writes a completion record; SubagentStop hooks write to different files. Measured on the author's log:

```
839 lines, sources {start: 694, pretool: 145}, zero terminal entries
```

So an agent that finished ten seconds after spawning still blocks compaction for the remaining five minutes, and the guard has no way to know. There is also no dedup by `agent_id`, and one agent is logged twice by design (once per writer) plus repeats across SendMessage resumes, so a single agent is counted as several.

## Why it survived

All five existing tests pass, and still pass after the fix. Every one of them builds an input with no `trigger` and no `background_tasks`, so they exercise only the fallback path. The behaviour under a real CC payload was never asserted.

## Fix

- `trigger === 'auto'` passes through unconditionally, first statement in the handler, ahead of the env check. Manual `/compact` still blocks, which is the hook's actual purpose.
- Prefer `input.background_tasks`, CC's authoritative live-task list (2.1.145+), which `subagent-stop/watchdog` already consumes for the same class of phantom-agent bug. An empty array is trusted as "nothing running". The timestamp window stays as the pre-2.1.145 fallback.
- Dedup the fallback by `agent_id`, newest row per agent.

Not fixed by writing completion records into `subagent-spawns.jsonl`: `watchdog` reads that file and treats every entry as a live spawn, so terminal rows there would make it report completed agents as hung.

yonatangross:
Closes #3451

## The guard blocked the only exit

Observed live 2026-08-12 in session `0de8dd40` at 100% context. CC printed its own recovery instruction, the operator typed `/compact`, and `lifecycle/pre-compact-guard` vetoed it:

```
Context limit reached · /compact or /clear to continue
Compaction blocked, 1 background agent(s) still running (ckleg2).
Wait for completion or set CLAUDE_CODE_ALLOW_COMPACT_DURING_AGENTS=1.
```

That leaves `/clear`, which discards the session state the guard exists to protect. A sixth finding in the compaction subsystem of #3321, in the one hook #3427 deliberately kept because it does fire. It fires wrongly. #3427 also armed `autoCompactWindow` a day earlier, so automatic compaction now runs at a chosen fill level rather than only at the model limit, making the veto more reachable, not less.

## Four changes

| # | Defect | Fix |
|---|---|---|
| 1 | Handler was `(_input, ctx)`, so auto and manual were indistinguishable | `trigger === 'auto'` passes through unconditionally, first statement, ahead of the env check |
| 2 | The printed remedy is unreachable from where it prints | first manual `/compact` blocks and drops a session-keyed marker; a second within 60s consumes it and proceeds |
| 3 | "Active" was recency, not liveness | prefer `input.background_tasks`, CC's authoritative live list; timestamp window becomes the pre-2.1.145 fallback |
| 4 | One agent counted several times | fallback dedupes by `agent_id`, newest row per agent |

`trigger` (`'auto' | 'manual'`) is typed and binary-verified in `types.ts:187` since 2.1.227, so the field was in the payload the whole time.

Defect 3 in numbers. `.claude/logs/subagent-spawns.jsonl` is a spawn-INTENT log with two writers and no closer, so `now - spawn.timestamp < 5min` answers "was one started recently", never "is one running":

```
839 lines, sources {start: 694, pretool: 145}, zero terminal entries
```

Defect 2 is why the auto fix alone was not enough: the incident was a *manual* compact. An env var read from a process env fixed at CC launch is not an escape hatch for someone already trapped. Repeat-to-confirm is the only override reachable from inside a stopped session. The marker lives at `.claude/state/precompact-confirm-<sessionId>.json`, following the `nudge-outcome-state.ts` convention, and writes are best-effort by design: a failed write costs one extra `/compact`, it must never change the verdict or throw out of the hook. Auto does **not** consume the marker, since it never blocks and so never needs a confirm; spending one silently would take away a confirm a manual attempt was holding, and a test pins that.

Deliberately NOT fixed by writing completion records into `subagent-spawns.jsonl`: `watchdog` reads that file and treats every entry as a live spawn, so terminal rows there would make it report completed agents as hung. One bug traded for another.

## Verification

- 7562 hook tests pass, 314 files. Typecheck clean, both builds clean, `validate-counts.sh` OK.
- Of the 8 new tests, **6 fail against the code they replace**. The other 2 confirm tests pass vacuously against the prior code, which blocked unconditionally and never unlinked, so they are labelled regression guards rather than controls. Counting them as controls would be the exact green-test-beside-an-inert-control failure this repo keeps finding in its own work.
- The 5 pre-existing tests passed before and still pass, which is why this survived: each builds an input with no `trigger` and no `background_tasks`, so all five only ever exercised the fallback path.

## One claim not made

That `background_tasks` reaches the PreCompact payload is confirmed for Stop and SubagentStop and **unproven** for PreCompact. A 2.1.228 binary probe found 32 occurrences, dominated by `background_tasks_changed`, with no co-occurrence with any compaction symbol; the payload builder is minified behind the event map, so `strings` cannot settle it either way. The code gates on `Array.isArray`, so absence falls to the deduped timestamp fallback, which is still better than what shipped. Neither the auto passthrough nor repeat-to-confirm depends on it.

## Playground

`docs/fix--3450-precompact-guard-auto/index.html`


yonatangross:
## Verified against the 2.1.228 binary

`~/.local/share/claude/versions/2.1.228`, 289 MB, via `strings -a`.

**1. `trigger` values, confirmed.** The binary carries its own hook documentation table:

```
| PreCompact | "manual"/"auto" | Before compaction |
```

That matches `types.ts:187` exactly, so reading `input.trigger` is contract, not inference.

**2. A blocking PreCompact hook does stop automatic compaction, on both auto paths.** There are four distinct log sites, and the two that matter are separate code paths, not one message reused:

```
Reactive compact blocked by PreCompact hook:
Precomputed compact blocked by PreCompact hook:
Compaction blocked by PreCompact hook          <- the message the session showed
compaction blocked by PreCompact hook; continuing uncompacted
```

`reactive-compact` is confirmed as the context-pressure path by an adjacent `forkLabel:"reactive-compact"` on the summarization call. `Precomputed` is the `autoCompactWindow` path #3427 armed. So before this PR the guard could veto both, which is exactly the observed failure.

The fourth string is worth noting: at least one caller degrades to `continuing uncompacted` rather than stopping. The main-session path is not that one, which is why the session in #3451 had nowhere to go.

**3. `background_tasks` on the PreCompact payload: still UNPROVEN.** 32 occurrences in the binary, but they are dominated by the `background_tasks_changed` event name plus `running_background_tasks` and `orphaned_background_tasks_pending_notification`. A targeted search for any co-occurrence of `background_tasks` with a compaction symbol returns nothing. The payload builder is minified behind the event map (`{... PreCompact: Gxe ...}`), so `strings` cannot settle it either way.

This is why the implementation gates on `Array.isArray(input.background_tasks)` rather than assuming the field. If CC does not send it, the hook falls to the deduped timestamp window, which is still better than what shipped. If CC does send it, liveness becomes real. Neither case affects the auto-compaction fix, which is the part that unblocks a stuck session.

Issue #3348, opened 2026-08-08
Lint rule for the PIPE_BUF/EPIPE class: here-strings, heredocs, grep -q pipes, tar-to-tar


yonatangross:
The 2026-08-08 session hit this class FIVE times, twice in code written hours after sweeping 29 files for it. Knowing the failure mode is not a mechanism.

Instances, all fixed on main that day:
1. scripts/ci/run-tests.sh done <<< "$FOUND" — 543 bytes vs 512 PIPE_BUF, bash 5.3 pipe-backed here-string, DEADLOCKED the whole local pre-commit gate with no output and no timeout. Why prior sessions used --no-verify.
2. tests/ci/lint.sh grep -q <<< "$agent_head" (~1073 bytes) — same deadlock, and it was itself the fix for a SIGPIPE race.
3. tests/unit/test-whats-new.sh awk <<< "$block" (~560 bytes) — written two lines below a comment warning about exactly this.
4. A checks-monitor heredoc fed multi-KB JSON to jq — caught before it wedged.
5. tests/unit/test-sync-versions.sh tar -cf - | tar -x — extractor exits on a complete archive, producer takes EPIPE, pipefail fails a test whose sandbox held all 6503/6503 files. Fires as a function of ARCHIVE SIZE, so adding files to the repo flips it.

Proposal: extend check-silent-failures.py (which already catches 2>/dev/null and || true swallows, and caught two of mine the same day) with:
- `<<<` or `<<` heredoc feeding a variable that is not provably < 512 bytes -> require process substitution `< <(printf ...)` or a temp file
- `| grep -q` under a file that sets pipefail -> require here-string only for provably-small payloads, else [[ =~ ]] or a temp file
- `tar -cf - | tar -x` -> require a temp tar file
- allow tag for reviewed exceptions, mirroring the existing silent: tags

Estate note: bash 3.2 is immune (tmpfile-backed) and CI Linux rarely reproduces, so this bites exactly and only on dev macOS — the least-observed environment.

yonatangross:
## Scoping check before implementing: this lands in hq-ext, not orchestkit

`check-silent-failures.py` does not exist in this repo. It ships in the **hq-ext plugin**:

```
installed  ~/.claude/plugins/cache/yonatan-hq/hq-ext/1.45.1/scripts/hooks/check-silent-failures.py  (423 lines)
source     /Users/yonatangross/coding/yonatan-hq/hq-ext-plugin   (branch dev)
```

[...]

Issue #3263, opened 2026-08-03
/ork:verify silently produces no verdict: background test tasks write 0 bytes, npm never starts


yonatangross:
## Symptom

`/ork:verify` spawned three background test tasks in one session (2026-08-03). All three
produced **zero bytes** and no verdict. The skill then waited on a completion signal that
never arrived, and would have waited indefinitely.

```
b6tvprwm6  08:41   0 bytes
bdycijacv  08:45   0 bytes
brvhtn4h3  08:48   0 bytes   <- still 0 bytes 36 min later

pgrep -f "vitest|npm test|run-security"  ->  0 processes alive
```

The failure is silent. `/ork:verify` does not report a wrong verdict, it reports **no**
verdict, so the operator waits on a result that is never coming. In this session it cost
roughly 40 minutes before I checked whether the process was even alive.

## What rules out the obvious explanations

**npm never started.** The output files do not even contain npm's own banner
(`> orchestkit@9.5.2 test`), which npm writes before running anything. So this is not a
test failing, or a suite hanging. The command died (or never launched) before npm produced
its first line of output.

**Backgrounding is not broken in general.** Two background tasks I launched by hand in the
same session, in the same repo, wrote output normally:

```
bybebr2ll   344 bytes   (CI watcher, exit 0)
bp2wt2s7p   389 bytes   (CI watcher, exit 0)
```

So the defect is specific to how the skill launches the suite, not to background execution.

## Where the skill does this

`src/skills/verify/SKILL.md:254`

```
Bash(command="npm test 2>&1", run_in_background=true)
Monitor(pid=test_task_id)  # Each line -> notification
```

## Hypotheses (NONE confirmed)

I did not fix anything, because the cause is not established. Each hypothesis below has the
command that would confirm or refute it.

1. **`npm test` dies instantly under the #3235 concurrency bug.** Five Claude sessions were
   live on this machine against this repo. Per the discussion on #3235,
   `tests/unit/test-sync-versions.sh` writes a `999.888.777` sentinel into tracked files and
   restores them afterwards, which makes `npm test` unsafe to run concurrently with any other
   repo activity.
   *Settle it:* run `npm test --quick` solo with no other session active, and separately with
   a second session running, and compare.

2. **The subagent's cwd or environment differs, so `npm` resolves to nothing.** A background
   Bash launched from inside a forked skill may not inherit the shell profile that puts `npm`
   on `PATH`, which would produce exactly this signature: no banner, no error, empty file.
[...]

PR #3488, opened 2026-08-14
fix(release): real component counts, and a machine account on announce

yonatangross:
## What

Two defects in the release rail, found by composing the announce payload locally rather than reading the code.

### 1. Component counts rendered as zero

`changelog-to-props.mjs` defaulted `statsBefore`/`statsAfter` to `{0,0,0}` and only populated them from a `Skill count: 197 -> 199` regex that release-please changelogs never emit.

Blast radius is wider than the announce payload: `release-video.yml:83,89` feeds on this same script, so every rendered release video has likely been displaying "0 skills, 0 hooks", including the assets attached to `v10.0.0-alpha.33`.

`agents` looked correct only by accident. A fallback scanned the entire changelog for `N agents` and took the most frequent value over 20, guessing a current total from historical prose. It happened to be right because the agent count has been stable. That fallback is deleted, not ported.

Counts now come from the ork manifest description, which is the same string `bin/validate-counts.sh` already asserts against the filesystem. A green CI makes them correct by construction.

```
before:  {"skills":0,   "agents":36, "hooks":0}
after:   {"skills":105, "agents":36, "hooks":216}
```

An explicit `skills: 105 -> 107` delta in a changelog still wins over the baseline.

### 2. The announce payload named no posting identity

`announce-release.mjs` sent no account selector, so the receiving platform fell back to its own default posting identity. An automated release rail must publish under a dedicated bot account, never a maintainer's personal one.

The payload now carries `account`, defaulted via `ANNOUNCE_ACCOUNT` to the project's automation account. The field is inert until the platform honours it, since an unknown key is ignored server-side. The platform half is tracked separately.

## What this deliberately does NOT change

The prerelease cadence gate stays exactly as it is. The instinct on "why has this never fired in 33 releases" is to open the gate, and the evidence says do not: alpha.33's entire announceable content was *"drop the last ork-elicit leftover from the esbuild header"*. Thirty-three of those would be worse than silence. The gate is the only thing that has kept them off the feed. The first real announce is the v10.0.0 GA cut.

## Verification

| Gate | Result |
|---|---|
| `tests/unit/test-changelog-parser.sh` | 13/13 pass |
| Mutation check | reintroducing the zeroed stats turns 4 assertions red |
| `shellcheck -S warning` | clean |
| `actionlint` | clean |
| `tests/security/run-security-tests.sh` | 19/19 pass |

The regression tests use process substitution with a trailing newline rather than here-strings. Here-strings deadlock on bash 5.3 past PIPE_BUF (#3348), and without the trailing newline `read` returns 1 at EOF, which under `set -e` kills the suite mid-run and prints no summary. That second failure mode bit this PR during authoring and is why the tests are commented the way they are.

Full `npm test --quick` was started but its background task wrote 0 bytes, which is the symptom described in #3263. The gates above were run directly in the foreground instead.

## Interactive Playground

**[Open Playground](https://github.com/yonatangross/orchestkit/blob/fix/announce-rail-stats-and-account/docs/fix--announce-rail-stats-and-account/release-rail.html)** — `docs/fix--announce-rail-stats-and-account/release-rail.html`

Three tabs, every number produced locally rather than retyped:

- **The stats bug** — toggle between the regex-only path and the manifest-derived path, plus why the manifest beat a filesystem walk, and the deleted "most frequent N agents" fallback.
- **The account field** — the payload diff, and why the identity is stated rather than inherited.
- **Why the gate stays** — what the latest prerelease would actually have announced, scored against the observed formula.

The playground is static markup; its script only toggles `hidden` and aria state, so the page contains no dynamic-HTML sinks.


yonatangross:
Superseded by #3489.

Closing rather than force-pushing: the commits on this branch described the change using material from an internal reference, including a maintainer's personal social handle with follower and post counts, an uninvolved third party's handle and their follower count, and the credential store's item-naming scheme. None of that belonged in a public repo, and a scrub commit on top would have left it in this PR's commit list and commit messages permanently.

#3489 carries the same fix authored from a clean base, with the rationale stated as a rule rather than a roster. It also drops the playground's innerHTML sinks entirely, which addresses the CodeQL failure here.

Issue #3461, opened 2026-08-12
Frontmatter hygiene: delete 33 dead hooks: declarations, decide the 36 mcpServers: (69 residual ignored keys)


yonatangross:
Split from #3335 close-out. CC 2.1.228 warns-and-ignores both arrays. hooks: (33 agents) has zero consumers — real wiring is the dispatcher — and reads as per-agent protection that does not exist; deletable. mcpServers: (36 agents) is context-class: CC ignores it but ork's own test-mcp-overhead.sh requires it, so removal needs a gate decision first. Evidence and counts re-derived at alpha.27 in #3335.

yonatangross:
Independently re-derived from a consumer repo (Yonatan-HQ/platform) while chasing a duplicate-plugin load. Counts match this issue exactly — **36 `mcpServers:`, 33 `hooks:`** — measured against the installed `ork-alpha` cache at **10.0.0-alpha.31**, so nothing has drifted since the alpha.27 derivation.

Three things to add.

**1. Still ignored at CC 2.1.231**, two releases past the 2.1.228 in the issue body. Read from the running binary rather than release notes:

```
["permissionMode","hooks","mcpServers"]
Plugin agent file ${e} sets ${V}, which is ignored for plugin agents.
Use .claude/agents/ for this level of control.
```

No upstream change is coming; this is stable behaviour, not a transient regression.

**2. The 36 split cleanly, and 7 of them need no gate decision at all.** The issue treats `mcpServers:` as one context-class block because `test-mcp-overhead.sh` requires it. That's true of the 29 that declare something. Seven declare an **empty list**:

```
demo-producer, emulate-engineer, expect-agent, genui-architect,
security-auditor, security-layer-auditor, system-design-reviewer
```

They pay the warning on every load while expressing no intent, and no test can depend on an empty declaration. Those are deletable now, independent of the `test-mcp-overhead.sh` decision — it shrinks the residual from 69 to 62 with zero risk.

The other 29 break down as:

| server | agents |
|---|---|
| `context7` | 22 |
| `tavily` | 5 |
| `storybook-mcp` | 3 |
| `stitch` | 2 |
| `memory` | 1 |
| `21st-dev-magic` | 1 |

[...]

PR #3494, opened 2026-08-14
chore(agents): delete the inert hooks: block from 33 agent frontmatters

yonatangross:
Part of #3461 (EPIC A). Deletes the unambiguous half; the `mcpServers:` half needs an owner decision and is surfaced below rather than taken.

## What was dead

CC's plugin-agent loader walks a fixed key array and warns-and-ignores each key it finds set. This is read out of the binary, not inferred from behaviour:

```
2.1.227 binary, offset 91099675   (identical on 2.1.226)

  [permissionMode, hooks, mcpServers]

  `Plugin agent file ${e} sets ${G}, which is ignored for plugin agents.
   Use .claude/agents/ for this level of configuration.`
```

So every `hooks:` block ork shipped in a plugin agent was inert. The corroboration: `agent/restrict-bash` reads like a Bash allowlist and had **never executed once across 193,606 hook-timing rows**.

Removed from **33 of 36** agents. `hooks` also dropped from both `KNOWN_AGENT_FIELDS` allowlists in `scripts/eval/static-analysis.sh`, so re-adding it to a plugin agent now warns instead of passing silently.

## Nothing is orphaned

The blocks referenced 14 distinct hooks. Every one stays live through the real registration contract:

| Hook | Real path |
|---|---|
| `pretool/bash/git-validator` | `sync-bash-dispatcher.ts:87` + `entries/pretool.ts:96` |
| `agent/restrict-bash` | `sync-bash-dispatcher.ts:26` + `entries/agent.ts:28` |
| 12 others | same pattern: dispatcher plus the `entries/` record |

The frontmatter block was a redundant second declaration sitting on top of working wiring.

## Scope, precisely

This is easy to over-read, so stating it plainly: the key is inert **only for plugin agents**. Copied into a consumer's own `.claude/agents/`, `hooks:` works exactly as documented, which is what the agent-hooks skill and its eval describe. Both were left untouched. The fix was never "the feature is broken", it was "we were declaring it at the wrong level".

## The count: 216 → 171

```
before   GLOBAL=150  AGENT=45  SKILL=21  TOTAL=216
after    GLOBAL=150  AGENT=0   SKILL=21  TOTAL=171
```

45 of the advertised 216 were declarations CC never read, so the drop is honest arithmetic rather than lost behaviour. Re-stamped across the manifest, `plugin.json`, `hooks.json`, README, CONTRIBUTING, `marketplace.json` and the docs site.

## A latent bug this exposed

`bin/count-hooks.sh` could not survive a bucket legitimately reaching zero:

```bash
set -euo pipefail
AGENT=$(grep -rch 'command:.*run-hook' src/agents/ | awk '…')
#       ^ exits 1 on no-match → pipefail propagates → set -e kills the
#         script BEFORE its final echo → TOTAL unbound in every caller
```

The symptom was `stamp-counts.sh: line 31: TOTAL: unbound variable`, which reads like a stamping bug and is actually a counting bug.

Fixed by inspecting grep's three-valued exit rather than a blanket `|| true`: **1** counts as 0, **≥2** still fails loudly with grep's own stderr. A blanket suppression would have turned a broken grep into a silent zero and quietly shrunk the advertised count. Both branches are covered in the commit's verification.

## Still open: `mcpServers:` (36 agents)

Deliberately not touched here, because it is a product decision rather than a cleanup:

| | |
|---|---|
| CC reads it? | **No** — same ignored-key array as `hooks` |
| Anything else reads it? | **Yes** — `tests/performance/test-mcp-overhead.sh` asserts *"Every agent .md file has an explicit mcpServers field"* |

So a test enforces a declaration with no runtime effect for plugin agents. Two coherent options: **(a)** delete the field and retire or rewrite that test to check something CC actually honours, or **(b)** keep both and document the field as advisory metadata rather than configuration. Needs an owner call.

## Verification

| Gate | Result |
|---|---|
| `bin/validate-counts.sh` | PASSED, all counts and versions match |
| `npm run test:agents` | all passed, 36 agents keep name/description/model/tools |
| `npm run test:skills` | every `/ork:` reference resolves, totals match `src/` |
| `npm run test:manifests` | version consistency passed |
| `tests/security/run-security-tests.sh` | all passed |
| `npm run build` | clean, plugins reassembled |
| count-hooks zero-match branch | returns 0, exit 0 |
| count-hooks error branch | fails loudly with grep stderr, non-zero |

The stripper refused to touch any file where removing the block would have dropped a sibling key; it reported 33 changed, 3 untouched, 0 refused.

## Note on labels

`docs/site/lib/generated/plugins-data.ts` is regenerated by the count re-stamp. It is a whole-file release-please governed path, so this PR carries `release-please-override` for that legitimate content change. No `version` field was hand-edited anywhere.

## Interactive Playground

**[Open Playground](https://github.com/yonatangross/orchestkit/blob/817532a9c6bf65ac697fccb3fb9449d2d14f420e/docs/chore--3461-delete-dead-agent-hooks/dead-hooks.html)** — `docs/chore--3461-delete-dead-agent-hooks/dead-hooks.html`

Four tabs: the binary proof and telemetry, the 14 hooks showing nothing was orphaned, the count math plus the `count-hooks.sh` bug, and the open `mcpServers:` decision. SHA-pinned and first-party, per the convention landed in #3491.


yonatangross:
## Correction: "nothing is orphaned" was wrong for 6 of the 14

CI caught it and the PR body overstated the case. I verified two hooks (`git-validator`, `restrict-bash`) and generalised to all fourteen. `tests/unit/test-hook-registry-closure.sh` disagreed:

```
✗ NEW DEAD HOOKS (6): registered but unreachable from
  hooks.json ∪ agent/skill frontmatter ∪ dispatcher fan-out
    pretool/bash/changelog-generator
    pretool/bash/ci-simulation
    pretool/bash/conflict-predictor
    pretool/bash/pre-commit-simulation
    pretool/bash/version-sync
    skill/merge-readiness-checker
```

**8 of 14 are reachable** via `sync-bash-dispatcher` + the entries maps, as claimed. **6 are not.**

## But the 6 were already dead before this PR

| Where they are | Status |
|---|---|
| `entries/pretool.ts` / `entries/skill.ts` | present |
| `src/hooks/hooks.json` | **absent** |
| dispatcher fan-out | absent |
| agent frontmatter | present, and CC ignores it for plugin agents |

That is this repo's own documented #959 class, quoted from `.claude/rules/hooks-development.md`: *"a hook in the entries map but missing from `hooks.json` is never invoked."*

So the deletion did not kill them. It **revealed** them. The closure test treats agent frontmatter as a reachability source, and since CC ignores that source for plugin agents, it was masking 6 hooks that could never fire. Declared by exactly 3 agents: `release-engineer` (changelog-generator, version-sync), `git-operations-engineer` (conflict-predictor, pre-commit-simulation, merge-readiness-checker), `ci-cd-engineer` (ci-simulation).

`tests/unit/test-git-enforcement-hooks.sh` fails for the same reason: it asserts `changelog-generator` is "wired to release-engineer" through frontmatter, i.e. it tests a wiring CC does not read.

## What I did NOT do

I did not try to make CI green by re-adding the frontmatter or by adding the 6 to the closure baseline. The baseline is explicitly monotonic (*"Do NOT add to the baseline — it only shrinks"*), and re-adding inert declarations to silence a test that just told the truth would be the exact anti-pattern this EPIC exists to remove.

## One measurement I got wrong, recorded honestly

I tried to prove the 6 never executed using `~/.claude/analytics/hook-timing*.jsonl` (119,647 rows). All six returned 0 — but so did my control, `git-validator`, which demonstrably runs. Inner-dispatcher hooks are recorded under the dispatcher's name, so that instrument cannot answer this question at all. The static reachability argument above stands on its own; the telemetry does not support it either way.

This PR is **blocked pending a scope decision** — see the next comment.

PR #3495, opened 2026-08-15
Change ProblemType to inherit from StrEnum

luzede:
## Summary

- Updates api-design skill's example script

## Consumer

## Test plan

- [x] `npm run build` clean
- [x] `npm run test:skills` (if skills changed)
- [x] `npm run test:agents` (if agents changed)
- [x] `npm run test:manifests` (if hooks/skills/agents added or removed)
- [x] `cd src/hooks && npm test` (if hook code changed — runs the same vitest suite as CI)
- [x] Smoke tests for any new scripts (see `tests/skills/test-*.sh`)

### DIFF
--- a/plugins/ork/skills/api-design/scripts/problem-detail-exceptions.py
+++ b/plugins/ork/skills/api-design/scripts/problem-detail-exceptions.py
@@ -10,6 +10,7 @@ Production-ready exception handling for FastAPI with:
 
 from dataclasses import dataclass, field
 from datetime import datetime, timezone
+from enum import StrEnum
 from typing import Any
 
 from fastapi import FastAPI, Request
@@ -60,7 +61,7 @@ class ValidationProblem(ProblemDetail):
 # Problem Type Registry
 # ============================================================================
 
-class ProblemType:
+class ProblemType(StrEnum):
     """Registry of problem type URIs."""
 
     BASE = "https://api.example.com/problems"


yonatangross:
Thanks for this, and sorry for the slow reply.

The change is right and it is landing. I have opened #3496 with you as
co-author.

Two things worth explaining, because neither was your fault.

**The file you edited is build output.** `plugins/` is generated from `src/` by
`npm run build`, so the same two lines needed to go in
`src/skills/api-design/scripts/problem-detail-exceptions.py`, with `plugins/`
regenerated and staged alongside. That is why the Build check was red: the
drift gate rebuilds from `src/` and compares.

You had no way to know this, because our own docs say the opposite.
`CONTRIBUTING.md:146` claims the `plugins/` directory is "not tracked in git"
and line 154 says to commit only `src/` and `manifests/`. Both are wrong and
both contradict the CI gate. I am fixing that separately, and it is the actual
bug your PR surfaced.

**One substantive correction.** Keeping `BASE` inside the `StrEnum` body makes
it an enum member, so the registry grows a ninth entry that is not a problem
type:

```python
list(ProblemType)
# as submitted: [BASE, VALIDATION_ERROR, RESOURCE_NOT_FOUND, ...]  9
# as shipped:   [VALIDATION_ERROR, RESOURCE_NOT_FOUND, ...]        8
```

Any code treating `list(ProblemType)` as the catalog of emittable problems
would pick up the bare namespace root. So `BASE` moved out to a module-level
`PROBLEM_TYPE_BASE`.

For the record, your f-strings were fine. I expected them to break inside an
enum body and they do not, because members are not created until the class body
finishes executing. The values were never wrong, only the membership set.

The enum also buys reverse lookup, which the plain class could not do:
`ProblemType("https://api.example.com/problems/rate-limit-exceeded")` now
returns `RATE_LIMIT_EXCEEDED`.

Happy to close this in favour of #3496, or if you would rather land it under
your own PR, push the two lines to `src/` plus `npm run build` output and I
will close mine instead. Your call, and either way the credit is yours.

yonatangross:
Closing this in favour of #3496, which carries your change with you as
co-author.

To be straight with you: my previous comment offered you the choice of landing
it under your own PR, and closing now takes that choice back. That is on me,
not on you. The reason is queue hygiene rather than anything about your work,
and the commit attribution in #3496 is yours.

The change was a good catch. If you want to keep going here, #3500 is labelled
`good first issue` and is a real hole rather than busywork: 51 Python scripts
ship under `src/skills/**/scripts/` with no CI syntax check or lint at all.
Your PR is what surfaced it.

And #3498 fixes the `CONTRIBUTING.md` text that sent you to the wrong file in
the first place.

Issue #3424, opened 2026-08-11
Sandbox rollout: loose first (strictAllowlist off + op/docker/ssh carve-outs), tighten on evidence


yonatangross:
## Sandbox rollout: loose first, tighten with evidence

Owner decision recorded 2026-08-11: enable sandbox in permissive mode first, verify against real usage, then ratchet. This issue is the plan plus the measured baseline it starts from.

## The finding that motivates it (from our own logs, today)

```
~/.claude/logs/ork/hooks.log:
[2026-08-11 05:15:22] [INFO] [network-egress-guard] ASK tier skipped: bypassPermissions mode
[2026-08-11 05:15:37] [INFO] [network-egress-guard] ASK tier skipped: bypassPermissions mode
... (repeating)
```

The regex egress guard's ASK tier is SKIPPED under bypassPermissions, which is how working sessions actually run on this machine. So the guard #3322 wants to retire is not merely bypassable in theory, it is bypassed in practice, today, by our own mode choice. An OS-level sandbox does not have that failure mode: it binds below the permission system.

Related: #3311 (its credential-read half ALREADY SHIPPED as `pretool/read/credential-read-guard`, live-probed; the sandbox half is this issue), #3322 (retire the regex guard AFTER the sandbox holds).

## What the sandbox must not break (measured from this machine's history)

| need | evidence | sandbox answer |
|---|---|---|
| 1Password CLI (biometric) | 87 `op` invocations in shell history | `excludedCommands` carve-out |
| docker / compose | 36 invocations (platform prod parity) | carve-out or verify socket access survives |
| ssh (hetzner deploys) | 29 invocations | carve-out initially |
| gh / git / npm registries | constant | network allowlist entries |
| localhost dev (portless) | curl history | allowlist localhost + *.localhost |

Installed CC (2.1.227) carries every knob the phased plan needs (verified in the binary): `sandbox.enabled`, `sandbox.network.strictAllowlist`, `excludedCommands`, `autoAllowBashIfSandboxed`, `tlsTerminate`.

## Constraint that shapes the rollout

[...]

Issue #3322, opened 2026-08-08
Adopt sandbox.enabled to retire the bypassable regex egress guard


yonatangross:
Epic D. Turning on CC's OS-enforced sandbox retires network-egress-guard as the boundary, keeping it at most as a fast advisory. Blocked by the sandbox issue above landing first.

yonatangross:
Verification for #3430 turned up something that strengthens this issue materially.

`network-egress-guard` has **zero registrations in `hooks.json`** and is reachable only as a
member of `sync-bash-dispatcher`, whose entry is gated by a 13-prefix `if:` clause. Cross
referencing that gate against the guard's own documented scope: `curl` reaches it, and
**wget, ssh, scp, rsync, nc and python never do**.

So the framing here, "bypassable regex egress guard", understates it. The guard is
*structurally unreachable* for 6 of the 7 vectors it documents, so no amount of hardening
inside `network-egress-guard.ts` can close the gap.

Constraint this places on whatever replaces it: it must not inherit a prefix gate.

Filed separately with the full evidence: https://github.com/yonatangross/orchestkit/issues/3438

yonatangross:
## Measured evidence: the loose stage IS in force, and what it costs

Session 2026-08-19, CC 2.1.234, macOS 25.5.0, worktree `3322-sandbox-block` with
`sandbox.enabled: true`, `strictAllowlist: false`, `excludedCommands [op, docker, ssh]`.

### 1. The sandbox is active, and ancestry cannot detect it

A process-ancestry walk is the wrong instrument. CC applies the Seatbelt profile
**in-process** to the spawned shell, so `sandbox-exec` never appears in the tree.
Full observable chain from a Bash tool call:

    Python -> zsh -> claude(2.1.234) -> zsh -> herdr -> herdr -> zsh -> [EPERM]

No `sandbox-exec` at any level, while the sandbox was simultaneously and provably on.
Detection must use **paired denial/control probes**, committed as
`tests/security/sandbox-probe.sh` (e2101c668):

| layer | deny probe | result | control | result |
|---|---|---|---|---|
| process | `ps -o comm= -p $$` | EPERM, rc=126/127 | `id -un` | ok |
| network | `curl pastebin.com` | rc=56, `CONNECT tunnel failed, response 403` | `curl api.github.com` | 200 |
| fs-read | `cat ~/.ssh/known_hosts` | EPERM | `cat ~/.zshrc` | ok |
[...]

PR #3554, opened 2026-08-18
docs(configure): add staged operator-scope settings

yonatangross:
## What this does

ork once authored a full `sandbox` block in `src/settings/ork.settings.json`. It was
deleted in a93ccb735 (#3357) because a plugin's bundled `settings.json` is not a
settings scope Claude Code reads: only `agent` and `subagentStatusLine` survive,
everything else is silently stripped. The deletion was correct. The opinion dying with
the plumbing was not. This re-homes the opinion onto CC's native mechanism.

Documentation only. No operator settings file is touched by this PR or by any code in it.

## The dangling pointer this fixes

`src/skills/doctor/SKILL.md:169` claimed the `ork:configure` skill carried a section
named *Operator-Scope Settings*. That section did not exist. So doctor Check 16
correctly detected a missing sandbox block and then pointed the operator at nothing.
`settings-posture.md` had the same problem, naming
`configure/references/cc-version-settings.md` for JSON that file never carried (it has
only an unrelated `bwrapPath` version note).

Both pointers now resolve:

```
$ grep -rn "Operator-Scope Settings" src/skills/configure/
src/skills/configure/references/operator-scope-settings.md:1:# Operator-Scope Settings
src/skills/configure/SKILL.md:250:## Operator-Scope Settings
```

## Staged, not all-or-nothing

Per the rollout recorded in #3424:

| Stage | Contents | Adopt when |
|-------|----------|------------|
| 1 LOOSE | `sandbox.enabled`, the 19 host exfil denylist, credential file and env denies, `excludedCommands` carve-outs for `op` / `docker` / `ssh`, no network allowlist | first, on a working machine |
| 2 STRICT | adds `network.allowedDomains` with `strictAllowlist: true` | only after stage 1 has held for several days |

Stage 1 is designed so nothing is refused for being unlisted, and the three workflows
this machine's history shows as fragile run outside the sandbox entirely. Stage 2 is the
ratchet, and the reference names what it breaks first in order: `api.github.com`, then
private registries, then localhost dev URLs, then telemetry ingest.

## Verification

Every key published was probed against the installed binary
`~/.local/share/claude/versions/2.1.234` with a **bare** match. A quoted probe returns
zero hits because the minified bundle accesses properties unquoted, which is a real way
to conclude a live key does not exist.

```
failIfUnavailable 12   autoAllowBashIfSandboxed 8   allowWrite 15   denyRead 16
allowedDomains 16      deniedDomains 8              allowUnixSockets 7
excludedCommands 6     strictAllowlist 5            NOT_A_REAL_KEY_CONTROL 0
```

The `credentials` entry shapes come from the zod schema extracted from the binary by
byte offset, not from ork's old block, because that block was inert and therefore never
validated by anything. It turns out correct: `files[]` is `{path, mode}`, `envVars[]` is
`{name, mode}`, `mode` constrained to `deny` or `mask`.

Two schema facts shaped the staging: `strictAllowlist` is honored only from user,
managed, or CLI scope and project settings are ignored; `deniedDomains` is merged from
all settings sources and always blocks. That is why stage 1 leans on the denylist and
stage 2 owns the allowlist.

Both JSON blocks parse under `json.load`. Tests: `tests/skills` 59/59 passed,
`bin/validate-counts.sh` PASSED, `npm run typecheck` clean.

## Deliberately not carried forward

- `failIfUnavailable: true`. Real key, was in the old block. It exits with an error at
  startup when the sandbox cannot start. Wrong first move on a personal machine.
  Documented with the reason, not in either paste block.
- `filesystem.allowWrite`. Real key. Its old value was `["${projectDir}", ...]` and that
  interpolation is unverified in user scope, where the schema resolves relative paths
  against `~/.claude`. Omitting keeps the default rather than shipping an unverified token.
- The env keys `settings-posture.md` already flagged as harmful.

## The egress guard stays

`src/hooks/src/pretool/bash/network-egress-guard.ts` keeps its behavior byte for byte
and gains a comment naming its successor and the condition for retirement. It is
bypassable and its ASK tier is skipped under `bypassPermissions`, but removing it now
trades weak protection for none. Retirement is gated on observing the sandbox actually
fire, which is what #3424's observation window is for.

## Honest limit

No live sandboxed session was run. The stage 1 breakage table (`mode: "deny"` unsets the
variable for sandboxed commands, so `gh` falls back to stored host credentials) is read
off the 2.1.234 key semantics, not measured. The reference says so in bold rather than
implying it was tested.

## Interactive Playground

`docs/fix--3322-sandbox-paste-block/index.html`

Self-contained single file, no external scripts, styles, or CDN. Toggle between stage 0
(today), stage 1 and stage 2 to see which threats each stage covers, the JSON to paste,
and the expected breakage for that stage.

Refs #3322, #3424, #3357.


yonatangross:
Duplicate of #3553, which merged the same commit (779a8d024) as 11cde42b4. This PR carries no unique content: its only differences from main are the version-bump files from the alpha.40 release (#3551) that landed after this branch was cut. Closing to avoid a second merge of the same change.

PR #3555, opened 2026-08-18
docs(configure): add good first issues link to issue template

### LINKED ISSUE #3506: Issue templates have no pointer to good first issues
## Problem

`.github/ISSUE_TEMPLATE/config.yml` has no pointer to the newcomer-friendly
work. Someone arriving at the issue tracker sees 91 open issues with no signal
about where to start.

## Proposal

Add a `contact_links` entry to `.github/ISSUE_TEMPLATE/config.yml` pointing at
the filtered label view, for example:

```yaml
contact_links:
  - name: Good first issues
    url: https://github.com/yonatangross/orchestkit/labels/good%20first%20issue
    about: New here? These are scoped, self-contained, and reviewed quickly.
```

Check the existing `config.yml` for what is already configured and match its
style rather than replacing it.

## Scope

One file, `.github/ISSUE_TEMPLATE/config.yml`.

## Good first issue notes

`.github/**` is on the CI playground gate's INERT list, so a PR touching only
this file does NOT require an HTML playground.

Yoliani:
## Problem

`.github/ISSUE_TEMPLATE/config.yml` has no pointer to the newcomer-friendly
work. Someone arriving at the issue tracker sees 91 open issues with no signal
about where to start.

## Fix

Add a `contact_links` entry pointing at the filtered `good first issue` label
view, placed after the existing "Where to start contributing" link and before
the security issue link. The entry matches the existing YAML style.

## Test Plan

- [ ] Open the issue tracker at https://github.com/yonatangross/orchestkit/issues/new/choose
- [ ] Verify "Good first issues" link appears in the right sidebar
- [ ] Click the link and confirm it navigates to the filtered label view

Closes #3506


yonatangross:
@Yoliani first, thank you for this, and sorry for the silence. You have been
sitting on a red X for a day and it was never your patch.

**Nothing is wrong with this PR.** It adds four lines to
`.github/ISSUE_TEMPLATE/config.yml` and it was fully green on your original
commit `05b9fe4`. What broke was our CI, and it broke the moment the branch was
updated with `main`.

What actually happened: three of our jobs hung inside a `Setup environment` step
and were killed when they hit their own `timeout-minutes`, before a single test
ran. GitHub reports a cap-killed job as `cancelled`, which is why it looked like
a flake worth re-running. It was not: a `sudo apt-get update` in our
`bin/ci-setup.sh` had no timeout around it, so when the package mirror stopped
answering the job simply burned its entire budget. We had fixed exactly this
class of bug a day earlier, but scoped the sweep to `.github/`, and that call
site lives in `bin/`, so it was missed.

That is tracked in #3557 and fixed in #3561, which is green now. The three jobs
that used to die at 15 and 30 minutes finish in 19s, 69s and 2m21s.

Nothing for you to do. Once #3561 lands we will update this branch, run the
checks, and take it from there. Apologies that your first contribution here ran
straight into our own infrastructure bug.

Issue #3365, opened 2026-08-09
worktree-merge-verifier fires on every squash merge, and calls landed work unrecoverable


yonatangross:
## Observed twice today, same false alarm

`worktree-merge-verifier` warns that a worktree holds unmerged commits at the exact moment those commits have just been squash-merged and verified on `origin/main`.

Two occurrences this session:

```
.worktrees/xsadopt   "7 unmerged commit(s) vs main"    PR #3360 already MERGED as eed7bb0c2
.worktrees/idiom     "10 unmerged commit(s) vs main"   PR #3363 already MERGED as 08fb023ef
```

In both cases the content was verified present on `origin/main` before the worktree was removed, and the branch was deleted safely.

## Root cause

The check is an ancestry test. This repo merges by **squash**, so a squash-merged branch's commits are genuinely not ancestors of `main` even though every line of their content is. Ancestry is therefore the wrong predicate for a squash-merge repo, and it will fire on every single PR this repo lands.

## Why it matters more than the noise

The warning says the commits are "unrecoverable" if the branch is deleted. That is false in the squash case and it is advice that points the wrong way: a user who believes it will keep dead branches and worktrees around forever. Worse, because it fires on *every* correct cleanup, it trains the operator to ignore it, which is precisely when a genuine unmerged-work warning would be missed.

This is the same defect class the v10 milestone is about: a check whose signal cannot distinguish the healthy case from the dangerous one, so its output stops meaning anything.

## Suggested fix

Ancestry alone cannot answer "has this landed" under squash. Options, cheapest first:

1. **Ask GitHub.** If the branch has a PR in state `MERGED`, it landed. Authoritative, one `gh pr list --head <branch> --state merged` call, no heuristics.
2. **Compare content, not history.** `git diff --quiet origin/main -- .` from the worktree, or compare the branch's tree against `origin/main`'s. Empty diff means nothing would be lost.
3. **Cherry-check.** `git cherry origin/main <branch>` reports patches already upstream by patch-id, which survives squash better than ancestry.

At minimum the message should stop asserting "unrecoverable" when it has not checked whether the content is already upstream, and should name squash as the reason it may be wrong.

## Repro

Land any PR in this repo with `gh pr merge --squash`, then `git worktree remove` the worktree that produced it. The warning fires every time.

PR #3570, opened 2026-08-19
fix(hooks): resolve verifier base against origin

### LINKED ISSUE #3569: worktree-merge-verifier: stale LOCAL base ref produces false 'unmerged commits' warnings
## Symptom

`worktree-merge-verifier` reports a large false "commits not in main" count when removing a worktree whose branch is **fully merged**, and tells you the work would be lost. Measured today in two different repos:

| repo | local `main` | vs local `main` | vs `origin/main` |
|---|---|---|---|
| orchestkit | 47 behind | **45 commits** | **0** |
| hq-ext-plugin | 587 behind | **731 commits** | 144 (of a real promote) |

In the orchestkit case the branch tip `cf231c64f` is a verified ancestor of `origin/main`:

```
git merge-base --is-ancestor cf231c64f origin/main   # rc=0, contained
```

so the true unmerged count is 0 and the advisory is entirely false.

## Root cause

`src/hooks/src/pretool/bash/worktree-merge-verifier.ts:196`

```ts
const baseCandidates = ['main', 'master', 'develop'];
const base = baseCandidates.find(b => gitRevParse(projectDir, b)) || 'main';
```

The base resolves to the **local** branch and the hook never fetches. In any
worktree-heavy checkout the local `main` is routinely hundreds of commits stale,
because nobody checks it out — work happens on branches and lands via squash
merge on the remote. Everything downstream inherits the stale base:
`unmergedCommitCount` (line 199), `contentAlreadyInBase` (line 205), and
`allCommitsSquashMerged` (line 210), so the squash-merge suppression added in
0.62.0 / #3365 cannot save it either — those containment checks are also asked
against the wrong ref.

The hook is already inconsistent with itself: `branchIsPushed` at line 170 uses
`origin/<branch>`. The base is the one ref it resolves locally.

## Why it matters more than a cosmetic warning

This advisory exists to make worktree cleanup safe. A false "731 commits would
be lost" on an already-merged branch trains the reader to ignore it, which is
exactly when the true positive gets ignored too. It also actively obstructs
routine cleanup sweeps.

It is not hypothetical: it caused a live session to **retract a correct
conclusion**. That session measured `git log origin/main..HEAD` (empty, correct),
was contradicted by this hook, and reversed itself to "45 commits, do not
delete". The instrument was wrong and the measurement was right.

## Proposed fix

Prefer `origin/<base>` and fall back to the local ref only when no remote-tracking
ref exists:

```ts
function resolveBase(projectDir: string): string | null {
  const candidates = ['main', 'master', 'develop'];
  for (const b of candidates) {
    if (gitRevParse(projectDir, `origin/${b}`)) return `origin/${b}`;
  }
  for (const b of candidates) {
    if (gitRevParse(projectDir, b)) return b;
  }
  return null;   // cannot measure -> stay silent, do not guess
}
```

Deliberately **not** fetching: this is a `PreToolUse` hook and must stay fast.
`origin/main` can itself be stale if the user never fetches, but it is strictly
closer to the truth than a local branch that is never checked out, and it makes
the hook consistent with its own pushed-check.

Also worth doing: when no base resolves at all, return silent rather than
falling back to the string `'main'`, so "could not measure" stops being
indistinguishable from "measured zero".

Credit: root cause and file:line diagnosed by a concurrent session working in
`.worktrees/3322-sandbox-block`; independently reproduced in hq-ext-plugin.

yonatangross:
Closes #3569.

## The bug

`worktree-merge-verifier` resolved its base ref to the LOCAL `main`/`master`/`develop` and never fetched. In a worktree-heavy checkout nobody checks `main` out, work lands via squash merge on the remote, and the local ref rots. The hook then reports a confident false "you will lose work" on branches that are fully merged.

Measured 2026-08-19 in two repos:

| repo | local `main` | vs local `main` | vs `origin/main` |
|---|---|---|---|
| orchestkit | 47 behind | **45 commits** | **0** |
| hq-ext-plugin | 587 behind | **731 commits** | 144 (a real promote) |

In the orchestkit case the tip is a verified ancestor of `origin/main` (`git merge-base --is-ancestor cf231c64f origin/main` returns 0), so the true count is zero and the whole advisory is noise.

The stale base also defeated the #3365 squash-merge suppression, because `contentAlreadyInBase` and `allCommitsSquashMerged` ask the same wrong ref.

## Why this is worth fixing rather than tolerating

The advisory exists to make worktree cleanup safe. A false "731 commits would be lost" on an already-merged branch trains the reader to ignore it, which is precisely when a true positive gets ignored too.

It is not hypothetical. It caused a live session to retract a CORRECT conclusion: that session ran `git log origin/main..HEAD`, got empty (right), was contradicted by this hook, and reversed itself to "45 commits, do not delete". The instrument was wrong and the measurement was right.

## The fix

`resolveBase()` prefers `origin/<base>`, falls back to the local ref only when no remote-tracking ref exists, and returns `null` when nothing resolves so the hook stays silent instead of guessing at the literal string `'main'`. That keeps "could not measure" distinct from "measured zero".

Deliberately does NOT fetch: this is `PreToolUse` and must stay fast. `origin/<base>` can itself lag when the user never fetches, but it is strictly closer to the truth than a branch that is never checked out. It also makes the hook internally consistent, since `branchIsPushed` already used `origin/<branch>`.

## Tests

New `worktree-merge-verifier-base.test.ts`, in the style of the #3365 suite: real throwaway git repos, no git mocking, because the bug lives in which ref git is asked about.

The central fixture rots a local `main` back behind a current `origin/main` and asserts both directions:

- `unmergedCommitCount(stale, 'landed', 'main')` is greater than 0, the bug
- `unmergedCommitCount(stale, 'landed', 'origin/main')` is 0, the truth
- `resolveBase()` picks the ref that yields the true answer

Plus fallback coverage: a repo with no remote resolves to local `main`, and a repo with no refs at all returns `null`.

## Verification

- new suite: 6 passed
- `src/hooks/src/__tests__/pretool/bash/`: 68 passed
- full hook suite: 313 files, 7490 passed, 2 skipped
- `tsc --noEmit`: clean

## Interactive Playground

`docs/fix--worktree-verifier-stale-base/base-ref-verdict.html`

[View the playground source](https://github.com/yonatangross/orchestkit/blob/37fa1f786ee02875b8c23e8fd57679ce0b50fe65/docs/fix--worktree-verifier-stale-base/base-ref-verdict.html)

"Which ref you ask decides the answer" puts the two measurements side by side for both repos in the table above, local `main` against `origin/main`, and shows why a false "you will lose work" is worse for the reader than the hook staying silent.


yonatangross:
Blocked on #3578. This PR is reviewed and approved on the diff, but it cannot get CI: the branch commits regenerated `dist/` bundles, and `main` regenerates the same four files on every hooks merge, so it returns to `mergeable: false` almost immediately after each rebase.

Measured today: rebase landed `9c1f8218c` clean, `main` advanced to `a4da29fa9` in the gap touching the same bundles, and the PR went straight back to `dirty`. A conflicting PR fires no `pull_request` event at all, hence zero CI runs and no red check to explain the silence.

Rebasing again would just re-enter the race. #3578 covers the structural fix.

Issue #3564, opened 2026-08-19
count-hooks.sh: bare mktemp zeros the SKILL subcount, blocking every commit


yonatangross:
## Symptom
`bin/validate-counts.sh` fails with a FALSE mismatch, which blocks every commit
via the pre-commit hook:

    ❌ Hooks: declared 171, actual 150
    Validation FAILED: 1 mismatches found

Preceded by:

    mktemp: mkstemp failed on /var/folders/.../T/tmp.XXXX: Operation not permitted
    bin/count-hooks.sh: line 47: : No such file or directory

## Root cause
`bin/count-hooks.sh:45`, inside `count_run_hook_lines()`:

    errfile="$(mktemp)"

Bare `mktemp` writes to the system temp dir. When that write is denied (EPERM),
`errfile` is empty, line 47 fails, the function returns 0, and the **SKILL
subcount is silently zeroed**.

    reported: GLOBAL=150 AGENT=0 SKILL=0  TOTAL=150   <- wrong
    actual:   GLOBAL=150 AGENT=0 SKILL=21 TOTAL=171   <- matches plugin.json

Verify the true SKILL count without mktemp:

    grep -rch 'command:.*run-hook' src/skills/ --include='SKILL.md' | paste -sd+ - | bc
    # => 21

## Why this matters
The failure mode is the one this repo already warns about: a counting tool that
returns a **silent zero** and shrinks the advertised count. The header comment at
`count-hooks.sh:47` explicitly guards grep's exit code for exactly this reason,
but the mktemp that supports that guard is itself unguarded. Anyone "fixing" the
mismatch by editing plugin.json to 150 would INTRODUCE real drift.

Reproduces whenever the system temp dir is unwritable, including under Claude
Code's Bash sandbox (see #3322) where it blocks all local commits.

## Proposed fix
Either drop the errfile entirely (capture stderr via a process substitution or
`2>&1` into a var), or make it honor a writable dir and fail loudly:

    errfile="$(mktemp "${TMPDIR:-/tmp}/count-hooks.XXXXXX")" || {
      echo "count-hooks: cannot create temp file" >&2; return 1; }

A zeroed subcount must never be reportable as success.

yonatangross:
Independently reproduced, and confirming the count: **171 is correct, 150 is the
bug.**

```
healthy            : GLOBAL=150 AGENT=0 SKILL=21 TOTAL=171
[...]

PR #3575, opened 2026-08-19
fix(tests): give the runner a scratch file it can actually create

yonatangross:
## Problem

`tests/run-all-tests.sh` died on its first scratch file and took the pre-commit
hook with it, so **no commit could be made at all** where the system temp dir is
not writable:

```
$ bash ./tests/run-all-tests.sh --lint
mktemp: mkstemp failed on /var/folders/.../T/tmp.0VHibpIGlz: Operation not permitted
```

Nothing else printed. The runner never reached a test, and pre-commit reported
`quick lint tests FAILED`, which reads as *your change broke the tests* when no
test had run.

Three bare `mktemp` calls under `set -e`: lines 89, 141, 183.

## Relationship to #3564

Same class, **not covered by it**. #3567 was scoped to `bin/count-hooks.sh`,
where the denial produced a silently wrong number. Here it produces total failure
to run, which is louder but worse, because this script sits on the commit path.

## Fix

`scratch_file()` tries `TMPDIR`, then `/tmp`, then a gitignored `tests/.tmp/`
beside the runner, and fails with a named cause if all three are unusable.

An explicit **template** is the actual repair, not a style preference: macOS
`mktemp` with no template ignores `TMPDIR` entirely and goes to `/var/folders`,
so `TMPDIR=/somewhere bash run-all-tests.sh` appears to work while proving
nothing.

Also fixes the two tests this runner executes that had the same defect:

| file | call |
| --- | --- |
| `tests/ci/test-count-hooks-temp-failure.sh:44` | `SHIM_DIR="$(mktemp -d)"` |
| `tests/ci/test-no-unbounded-package-manager.sh:78` | `HITS_FILE="$(mktemp)"` |

Both were written to catch a denied `mktemp` silently zeroing a count, and both
died of the condition they exist to detect. They pass in CI, whose temp dir
works, and fail in exactly the environment the bug lives in.

## Test Plan

Verified against the live denial rather than a simulation.

- [x] before: one error line, zero categories run
- [x] after: **3 categories, ALL TESTS PASSED**
- [x] both repaired tests report `RESULT: PASS` individually
- [x] `bash -n` clean on all three scripts

## Interactive Playground

`docs/fix--runner-mktemp/scratch-fallback.html`

Toggle the denial and the fallback to see where the failure lands.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -238,3 +238,6 @@ docs/site/.claude/memory/sessions/
 # (__guards_main_tmp.ts, __probe_main_tmp.ts) — one of them would have poisoned
 # any `grep permission_mode src/`.
 **/__*_tmp.ts
+
+# scratch fallback when the system temp dir is unwritable (see tests/run-all-tests.sh)
+tests/.tmp/
--- a/docs/fix--runner-mktemp/scratch-fallback.html
+++ b/docs/fix--runner-mktemp/scratch-fallback.html
@@ -1,0 +1,116 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+<meta charset="utf-8">
+<meta name="viewport" content="width=device-width, initial-scale=1">
+<title>A test runner that could not create a scratch file</title>
+<style>
+  :root{--bg:#0d1117;--panel:#161b22;--line:#30363d;--fg:#e6edf3;--dim:#8b949e;
+        --ok:#3fb950;--bad:#f85149;--warn:#d29922;--accent:#58a6ff}
+  *{box-sizing:border-box}
+  body{margin:0;padding:2rem 1.25rem;background:var(--bg);color:var(--fg);
+       font:15px/1.6 ui-monospace,SFMono-Regular,Menlo,Consolas,monospace}
+  main{max-width:58rem;margin:0 auto}
+  h1{font-size:1.3rem;margin:0 0 .35rem}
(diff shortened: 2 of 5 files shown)


yonatangross:
Closing as **already landed**, not as abandoned.

This branch was merged into `fix/credential-guard-truth` as a prerequisite,
because without it the pre-commit lint gate could not run at all in a sandboxed
shell. #3577 squash-merged at `5bf1a0b5b9`, carrying this fix with it.

Verified on `origin/main` — every file this PR touches is now byte-identical:

```
git diff --quiet origin/main origin/fix/runner-mktemp -- tests/run-all-tests.sh                    -> identical
git diff --quiet origin/main origin/fix/runner-mktemp -- tests/ci/test-count-hooks-temp-failure.sh -> identical
git diff --quiet origin/main origin/fix/runner-mktemp -- tests/ci/test-no-unbounded-package-manager.sh -> identical
git show origin/main:tests/run-all-tests.sh | grep -c 'scratch_file()'  -> 1
```

The three-dot diff still shows content because #3577 was **squash**-merged, so
this branch is not an ancestor of `main` even though its bytes are. Merging it
now would be a no-op.

The fix itself is unchanged and live: `tests/run-all-tests.sh` no longer dies on
a bare `mktemp`, and the two `tests/ci/` scripts that had the same defect (both
written to catch a denied `mktemp`, both killed by one) run again.

PR #3646, opened 2026-08-21
fix(playground): build output as nodes, not innerHTML

yonatangross:
CodeQL flagged a high-severity "DOM text reinterpreted as HTML" alert on the playground added in #3640. That PR auto-merged the moment its required checks went green, before this fix (pushed minutes later) could ride along, so main currently carries the vulnerable version.

The draft concatenated the input field's value straight into `innerHTML`. User-supplied text now goes in via `textContent` and the output is assembled from real nodes, so there is no HTML reinterpretation path left. Behavior of the playground is unchanged.

Scope: one file, the playground added in #3640. No source or route changes.

## Interactive Playground

Same playground this PR fixes: `docs/fix--rss-allowlist/index.html` (the SERVED_EXACT drift-guard explorer from #3640). No new one added for a security fix to an existing page.

https://claude.ai/code/session_015Ahw2HqT7UEXrwHB15BV6b


yonatangross:
Superseded by #3648. This branch was cut before #3640 squash-merged, so it carried the already-merged commits and went conflicted. #3648 is the same one-file fix rebuilt on top of main.

PR #3696, opened 2026-08-24
fix: docs: add CONTRIBUTING.md

### LINKED ISSUE #3505: CONTRIBUTING has no worked example of a first pull request
## Problem

`CONTRIBUTING.md` explains the repo layout and the build, but never walks
through a complete first pull request. The one thing a newcomer most needs,
"I want to change this line, what exactly do I do", is not written down.

This is the gap behind #3495: an outside contributor edited a generated file
because nothing showed them the source-then-build loop end to end. #3498 fixed
the incorrect statements, but there is still no worked example.

## Proposal

Add a short "Your first pull request" section with one concrete, real example
followed all the way through. Suggested shape:

1. Find the file under `src/` (not `plugins/`, which is build output)
2. Make the change
3. `npm run build`
4. `git status`, showing that BOTH the `src/` file and its regenerated copies
   appear, and that all of them get staged
5. Commit with a conventional-commit title (note the 72-char limit, which the
   pre-commit hook enforces)
6. Push and open the PR

Worth calling out explicitly, because each has bitten someone:

- editing `plugins/` instead of `src/` (#3495)
- forgetting `npm run build`, which fails the `Build` drift check
- a commit title over 72 chars, rejected by the pre-commit hook
- `npm install` is required in a fresh worktree; `git worktree add` does not
  copy `node_modules`

## Scope

One file, `CONTRIBUTING.md`.

## Good first issue notes

Genuinely well suited to a newcomer, because a newcomer knows better than a
maintainer which parts are confusing. Write down what you actually hit.

`CONTRIBUTING.md` is on the CI playground gate's INERT list, so a PR touching
only this file does NOT require an HTML playground.

VedantMadane:
## Summary

docs: add CONTRIBUTING.md

## Changes

- CONTRIBUTING.md

Fixes #3505


VedantMadane:
Closing this PR after an internal quality audit.

It replaces an existing project-specific CONTRIBUTING guide with a short generic template, which deletes maintainer documentation. That is not an acceptable contribution.

Sorry for the noise.

PR #3815, opened 2026-08-30
fix(hooks): redact-secrets reads tool_response, the field CC sends

### LINKED ISSUE #3725: redact-secrets layer reads tool_result, never tool_response CC sends
## What

`src/hooks/src/skill/redact-secrets.ts` reads `(input as any).tool_result || (input as any).output` and never `tool_response`, while `posttool/secret-handler.ts`'s own comment records that `tool_response` is what CC sends on PostToolUse. Result: this layer's patterns (including the GitLab family added in #3722) never see a production payload. Found by the 2026-08-24 verify-all sweep; pre-existing, not introduced by #3722, but #3722's title "in all three redaction layers" overstates what is delivered.

## Evidence

`python3 -c` over the file: fields read = `['output', 'tool_result']`. The vitest suites pass because each layer is tested with the field it happens to read, which is exactly why the mismatch survives (same shape as the #3427 closed-loop fixtures).

## Fix shape

Read `tool_response` first (then `tool_result`, then `output` for older payloads), and add one test that sends a CC-shaped PostToolUse payload (`hook_event_name: PostToolUse`, `tool_response: "... glpat-... "`) through the built `plugins/ork/hooks/dist/skill.mjs` via `run-hook.mjs`, asserting the token does not survive. The verify-all synthesizer's exact gap: "no test or probe sent a single CC-shaped payload through all three layers in one run."

ClaraOswald076:
skill/redact-secrets read `tool_result || output`, never `tool_response` — but `tool_response` is the field CC sends on PostToolUse (posttool/secret-handler.ts's own comment records this, and it reads tool_response first for exactly this reason). On real payloads the read came back empty, the hook returned silent success, and this layer's patterns — including the GitLab token family added in #3722 — never matched anything in production.

The existing vitest suites stayed green because every layer's tests feed it the field it happens to read: redact-secrets.test.ts builds all of its payloads with `tool_result`. That's the same closed-loop gap the verify-all sweep named: no test sent a single CC-shaped payload through the layer the way production does.

Fix: read `tool_response` first, then `tool_result`, then `output` for older payloads — the same precedence secret-handler.ts uses.

Verification: one CC-shaped PostToolUse payload (`hook_event_name: PostToolUse`, `tool_response` carrying a `glpat-` token, no legacy aliases) driven two ways — through the built `plugins/ork/hooks/dist/skill.mjs` via `plugins/ork/hooks/bin/run-hook.mjs` (asserting exit 0, no token in stdout, and the `::warning::Potential API key detected` stderr annotation on Linux/macOS), and through the same built bundle imported directly via `file://` URL so the detection assertion holds cross-platform. Both tiers are red before the fix and green after; the 65 existing unit tests stay green.

One thing I hit while running the dispatcher locally on Windows: run-hook.mjs does `import(<absolute dist path>)`, which throws ERR_UNSUPPORTED_ESM_URL_SCHEME on win32, and the surrounding catch turns that into a silent success — so through the dispatcher, every hook no-ops on Windows (the runner exits before stdin is even read). Kept out of scope here; the spawned tier in the new integration test degrades to envelope-only assertions there, and the in-process tier covers detection on all platforms.

Fixes #3725


yonatangross:
Thanks, this was exactly right and merged as-is (9e0129664, closes #3725).

What I checked before releasing CI: the read order against main (`redact-secrets.ts:33` vs `secret-handler.ts:60`), typecheck and 70/70 on your head, and your integration test copied onto main's built bundle, where it fails 2/3, so it measures the fix rather than the mock. That control is the part I appreciated most.

House paperwork (registry changelog entry, Lab playground) is landing in a maintainer follow-up so you did not have to learn it for one line. Your Windows finding is now its own issue, credited to this PR.

PR #3819, opened 2026-08-30
fix(hooks): runner names a dist bundle that exists but cannot load

### LINKED ISSUE #3817: run-hook.mjs: dist import by absolute path throws on Windows and the catch maps it to silent success
## What

Through the dispatcher, every ork hook no-ops on Windows. `bin/run-hook.mjs` does `import(<absolute dist path>)`; on win32 a bare `C:\...` path is not a valid ESM specifier, Node throws `ERR_UNSUPPORTED_ESM_URL_SCHEME`, and the runner's surrounding `catch` turns that into the silent-success envelope before stdin is even read.

Reported by the contributor of #3815 while running the new integration test locally on Windows (their words: "through the dispatcher, every hook no-ops on Windows (the runner exits before stdin is even read)"). Kept out of #3815's scope, correctly. Not yet reproduced by a maintainer; the mechanism matches Node's documented rule that `import()` on Windows needs a `file://` URL (`pathToFileURL`), and the runner's catch-to-silent shape is the same fail-open the verdict probes exist to catch.

## Why it matters

- This is the #959 / #3804 class, platform-shaped: registration fine, handler fine, the runner never reaches the handler, and `{"continue":true}` looks identical to a hook that ran and passed.
- The `Windows Smoke Test (PowerShell)` workflow is green, so it is not asserting a verdict; it can only be checking exit codes or envelope shape.

## Fix shape

1. In `run-hook.mjs`, import the bundle via `pathToFileURL(distPath).href` (the same thing the new `redact-secrets-tool-response.test.ts` does for its in-process tier).
2. Make the catch distinguish "bundle failed to load" from "hook threw": the first must NOT be silent success; a `verdict: error` on the hook-timing line plus a stderr line is the honest outcome.
3. Add a Windows probe that asserts a verdict (drive `pretool/bash/dangerous-command-blocker` with `rm -rf /` through the dispatcher and expect deny), so the smoke test cannot stay green while the dispatcher is dead.

Refs: #3815 (where it was found), #3804 (dead-by-payload sibling), #3801 (a swallow that certified the defect).

yonatangross:
## Summary
- `bin/run-hook.mjs:351` imported the dist bundle by bare absolute path; win32 rejects that (`ERR_UNSUPPORTED_ESM_URL_SCHEME`, Node reads `C:` as a scheme) and the catch at 355-362 called `silentExit()`: `{"continue":true,"suppressOutput":true}`, exit 0, no stderr, no hook-timing row (stdin was never read). So every hook, security guards included, no-oped on Windows through the dispatcher. `windows-smoke.yml` was green the whole time: it piped `echo hello` in and asserted only `$LASTEXITCODE -eq 0`.
- Import via `pathToFileURL(bundlePath).href`; the never-called `loadBundle()` duplicate is deleted.
- The catch now separates the two load outcomes. Absent file: silent, as before (the legitimate not-built case). Present but unloadable: one stderr line naming the file and error, a synchronous hook-timing row `{ok:false, verdict:"bundle-error"}`, exit 2 for `SECURITY_HOOKS` and `--rewake` entries (fail closed), exit 1 for every other hook, nothing on stdout. `timingRow()` is shared by both writers.
- `run-probes.mjs` classifies a non-zero exit with no envelope and the stderr signature as `bundle-error` instead of the ambiguous `silent`.
- Windows job: new step runs `run-probes.mjs --only bash-catastrophic-rm-deny` with `PLUGIN_ROOT=src` and asserts the `deny` verdict. Paths filter now includes `tests/hooks/verdict-probes/**`.
- README: verdict vocabulary, a bundle-load paragraph, registry changelog entry (count unchanged at 175).

## Control (fails without the fix)
`src/hooks/src/__tests__/bin/run-hook-bundle-error.test.ts` copies the runner into a private tree with a dist file that is a SyntaxError on import (win32 is unavailable locally; a synthetic unloadable bundle throws on every platform).

| case | main's runner | this branch |
|---|---|---|
| security hook: exit 2, stderr names the file, row `verdict: bundle-error` | fail | pass |
| plain hook: exit 1, same stderr and row | fail | pass |
| plain hook with `--rewake`: exit 2 | fail | pass |
| bundle file absent: exit 0, silent envelope, no stderr, no row | pass | pass |

The real win32 red-then-green is a separate draft PR carrying only the workflow change; its run URL is linked in a comment below once it finishes.

## Verification
- `src/hooks` vitest: 7651 passed, 0 failed, 2 skipped. `npm run typecheck` clean.
- `tests/hooks/test-hook-verdict-probes.sh`: 17 passed, 0 failed.
- `cd src/hooks && npm run build` and root `npm run build` done; `plugins/ork/hooks/bin/run-hook.mjs` carries the change.

## Interactive Playground

**[Open in the Lab](https://orchestkit.yonyon.ai/lab/every-hook-no-oped-on-windows.html)** (source: `docs/fix--3817-windows-bundle-url/index.html`). Pick a bundle state and a hook and read what the runner does, before and after.

Closes #3817

https://claude.ai/code/session_01SD8MnvP5vBRBCvhHbF4yQs


yonatangross:
Windows runner, red then green, as measured:

- red control (#3820, workflow step without the runner fix): failure https://github.com/yonatangross/orchestkit/actions/runs/33318501842
- this branch (with the fix): success https://github.com/yonatangross/orchestkit/actions/runs/33318497609

The probe step is what separates them: on the unfixed runner it reports the dispatcher's own load failure instead of a deny; with the fix the same probe classifies `deny`.

PR #3912, opened 2026-09-03
docs(create-pr): a zero from the harvest can mean never reviewed

### LINKED ISSUE #3913: test-mdx-compile stub-swap dies on npm 'edgesOut', and takes test-sync-versions with it
## What happens

`tests/unit/test-mdx-compile.sh` intermittently dies in its `docs/site` stub-swap fallback:

```
⚠ @mdx-js/mdx not in docs/site/node_modules — installing
⚠ npm ci hit 401 on @yonatan-hq/analytics — swapping to local stub and regenerating lockfile
npm error Cannot read properties of null (reading 'edgesOut')
✗ stub-regen + npm ci failed in docs/site
```

`test-sync-versions.sh` then fails as a **cascade**, not on its own merit: the crashed
regeneration ran `rm -f package-lock.json` and never replaced it, so the next test hits
`tar: docs/site/package-lock.json: Cannot stat: No such file or directory` and exits 2.
Two red tests, one cause.

## It is not any one PR's diff

Paired probe, run 2026-09-03:

| Run | Commit | Unit Tests (node 22) |
|---|---|---|
| 33736591271 (original, 09:03Z) | `main` | success |
| 33736591271 (**rerun, same commit, ~13:20Z**) | `main` | **failure** |
| 33757549807 (#3912) | branch | failure |
| 33757549807 (rerun) | branch | failure |
| 33759002441 (#3912, next push) | branch | success |

Re-running `main`'s own green commit reproduces it. The failure tracks time and runner
cache state, not the tree, so "re-run it" is a coin flip and any PR can go red for a
reason that has nothing to do with it.

## Likely cause

`npm ci` deletes `node_modules` and then fails on the 401, leaving a partial tree. The
fallback at `tests/unit/test-mdx-compile.sh:84-88` then runs
`npm install --package-lock-only` against that partial tree, and arborist walks a node
with no edges. Candidate fix, one line, makes the fallback independent of whatever the
failed `npm ci` left behind:

```diff
       (cd "$DOCS_SITE" \
         && npm pkg set 'dependencies.@yonatan-hq/analytics=file:../stubs/analytics-stub' >/dev/null \
         && rm -f package-lock.json \
+        && rm -rf node_modules \
         && npm install --no-audit --no-fund --package-lock-only --ignore-scripts 2>&1 | tail -3 \
         && npm ci --no-audit --no-fund 2>&1 | tail -3) || {
```

Unverified: I did not reproduce it locally, so the partial-tree explanation is a
hypothesis fitted to the log, not a confirmed root cause. Whoever picks this up should
reproduce first (copy `docs/site` to a scratch dir with no `NPM_TOKEN`, let `npm ci` 401,
then run the fallback as-is) rather than land the diff on the strength of the reasoning.

## Second defect in the same block, independent of the flake

The fallback deletes `package-lock.json` before it can regenerate it, so **every** failure
mode of the regeneration leaves the working tree short a tracked file, which is what turns
one red test into two. Restoring the lockfile on the failure path would decouple them and
make the real failure legible.

Found while landing #3912.

yonatangross:
## Summary

`coderabbit-harvest.sh --unresolved` printing `[]` has two causes the harvest phase cannot tell apart: CodeRabbit read the diff and found nothing, or CodeRabbit never saw the PR. Both print `0`, both read as merge-ready, so a repo the app cannot reach passes the quality bar on every PR forever.

That is this repo. Measured 2026-09-03: `.coderabbit.yaml` has been on `main` since 2026-08-25 (#3746), schema-valid, `auto_review.enabled: true`, `tone_instructions` at 221 of the 250-char limit, with no filter that would exclude our PRs. CodeRabbit has posted **zero** comments here, ever. Every PR review comment on record in this repo is from `github-advanced-security[bot]` (56) or `Copilot` (9). Over the same window `Yonatan-HQ/platform` had CodeRabbit on **238 of its 311** non-bot PRs, so the subscription was live the whole time.

The GitHub App installation is scoped per **owner account**, each with its own repo grant. platform is org-owned (`Yonatan-HQ`), orchestkit is user-owned (`yonatangross`), and the org grant never covered the personal repo. No edit to `.coderabbit.yaml` could ever have reached that.

## Changes

- `src/skills/create-pr/references/coderabbit-zero-reviews.md` (new): the disambiguation ladder. Both comment endpoints (summaries land on `/issues/comments`, findings on `/pulls/comments`, so a single-endpoint read is half blind), the positive control that keeps a zero honest (the login is `coderabbitai[bot]`; bare `coderabbitai` returns 0 against a repo where it posts daily), the check-suite app-slug probe that answers the installation question without a browser, and the operator steps for the grant.
- `src/skills/create-pr/SKILL.md`: the harvest stage no longer treats `0` as clean without asking which zero it is, and the file table lists the new reference.
- `.coderabbit.yaml`: header states the file is inert until the app is installed on the owner account, with the measurement and the settings URL, so nobody spends another day tuning `path_filters`.

## Test Plan

- `npm run test:skills` — SUCCESS, every `/ork:` reference resolves and every plugin-wide total matches `src/`
- `npm run test:agents` — 17 passed, 0 failed
- `npm run test:manifests` — all checks passed
- `npm run build` — plugins regenerated, `plugins/` staged with `src/`
- `node docs/site/scripts/generate-lab-data.mjs` — 94 lab entries
- **This PR is itself the live probe.** If CodeRabbit posts a review here, the grant was fixed. If it does not, the app is still not installed and the operator step below is still outstanding.

## Operator action this PR cannot perform

Installing a GitHub App is a web OAuth grant. No token in a session can do it.

1. https://github.com/settings/installations, find CodeRabbit
2. Absent, install on the `yonatangross` account from https://github.com/apps/coderabbitai
3. Present, Configure, add `yonatangross/orchestkit` to the repo grant
4. https://app.coderabbit.ai, confirm the repo is listed and enabled

orchestkit is public and CodeRabbit reviews public repositories free, so this costs no seat.

## CI root cause, folded in (Closes #3913)

`test-mdx-compile.sh`'s stub-swap fallback deleted `docs/site/package-lock.json` and rebuilt the whole graph with `npm install --package-lock-only`. npm 10.9.8, which `actions/setup-node` gives node 22, crashes on that walk at `#loadPeerSet` (`@npmcli/arborist/lib/arborist/build-ideal-tree.js:1289`) with `Cannot read properties of null (reading 'edgesOut')`. npm 10.8.2 (node 20) and npm 11.19.0 pass, which is why only the node-22 leg went red. With no lockfile to pin it the resolution ran live against the registry, so the same commit passed at 09:03Z and failed at 12:52Z: rerunning `main`'s own green run 33736591271 reproduced it, so this was never any one PR's diff.

Fixed by keeping the lockfile and letting `npm install` reconcile against it, so only the changed entry re-resolves and the whole-tree peer-set walk never runs. Version-independent, not an npm pin. Also fixes the cascade (nothing is deleted, so `test-sync-versions.sh` can no longer die on `tar: Cannot stat`) and stops `| tail -3` from discarding the stack.

## Interactive Playground

https://orchestkit.yonyon.ai/lab/coderabbit-never-installed.html

Source: `docs/fix--coderabbit-live/index.html`. Pick a repo to see the same query against both, read the two-endpoint table, and replay 22 days of real platform PR timestamps against each plan's hourly cap.

https://claude.ai/code/session_01CcNC5F1cA63gj5YJVMkT1v


yonatangross:
@coderabbitai review

Explicit-trigger probe. If the app is installed on this account and only auto-review were off, this comment forces a review. Silence here separates "configured off" from "never installed", which the PR body claims is the real state.

yonatangross (commented):
1. `src/skills/create-pr/references/coderabbit-zero-reviews.md:24` calls the result "EVER" but limits issue comments to `since=2026-01-01`, so a repo whose CodeRabbit summaries predate that cutoff can return zero and be falsely diagnosed as uninstalled at line 44; remove the cutoff or explicitly label and handle it as a bounded observation window.

2. `src/skills/create-pr/references/coderabbit-zero-reviews.md:33` treats absence from check suites as proof that CodeRabbit is uninstalled, but check suites only show checks created for that SHA and an installed app need not emit one; make this signal non-conclusive and use a direct installation or documented CodeRabbit verification path for the installation diagnosis.

yonatangross:
Both findings upheld and fixed in 6695f20c8. Neither was cosmetic: finding 1 was a hole the doc could have walked someone straight into, and finding 2 was me presenting corroboration as proof.

**1. `since=2026-01-01` under a comment that says EVER.** Confirmed, and worse than stated: `since` on `/issues/comments` filters on `updated_at`, not `created_at`, so the window is not even the one a reader would assume. A repo whose CodeRabbit activity predates the cutoff returns 0 and lands on the "not installed" row. Removed the cutoff, both endpoints are unbounded now, and the result table gained an explicit zero-strength row: *"Zero comments but the query carried a `since=`, strength none, re-run unbounded before concluding anything."* The comment now says why the cutoff must not come back.

**2. Check-suite absence as proof.** Upheld on the general claim, and I measured it rather than argue it. You are right that an installed app need not emit a suite (it needs the checks permission), so absence proves nothing for an arbitrary app. It does mean something for CodeRabbit, but only because of a control I had not run:

```
Yonatan-HQ/platform #11024, #11057, #11058  (CodeRabbit demonstrably reviews there)
  check-SUITES app slugs: claude,coderabbitai,cursor,github-actions,posthog-eu,renovate,vercel
  check-RUNS  app slugs: github-actions
```

Two consequences now in the doc. The probe is demoted from conclusive to **corroborating**, with the instruction to run the control repo first and treat the probe as inert if the slug is absent there too. And it must be check-**suites**, not check-**runs**: CodeRabbit opens a suite and files no runs in it, so the check-runs list would have read as absent on a repo where it is installed and working. That was a second latent false-negative in the same three lines.

Also added, prompted by finding 2: the doc now states that the one direct answer, `gh api /user/installations`, returns 403 on a normal user token, so a browser-blocked session must say the settings page was never read instead of implying it was.

No inline threads on this PR, so there is nothing for `coderabbit-harvest.sh --resolve` to close.

---

Unrelated to the findings, on the red CI: `test-mdx-compile.sh` and `test-sync-versions.sh` fail on this branch with `npm error Cannot read properties of null (reading 'edgesOut')` in the `docs/site` stub-swap fallback. **This is not from this diff.** Control: re-running `main`'s own commit (run 33736591271, green at 09:03Z today) now fails the identical job. The second failure is a cascade, the crashed regeneration leaves `docs/site/package-lock.json` deleted and `test-sync-versions.sh` dies on `tar: Cannot stat`. Tracking separately rather than papering over it here.

yonatangross (commented):
1. `src/skills/create-pr/references/coderabbit-zero-reviews.md:58` still marks an unbounded zero plus a positive control as conclusive proof that the app is not installed, but CodeRabbit supports account-level overrides and auto-review exclusions that can suppress all reviews from the start; change this row to inconclusive and require an explicit `@coderabbitai review` probe or the installation page before declaring the app absent.

yonatangross:
Upheld and fixed in 5728f7975. You are right, and the row you flagged was the last place I was still treating "we saw nothing" as "there is nothing".

**The hole.** An unbounded zero on both endpoints is exactly what an *installed* app produces when it is disabled in the CodeRabbit dashboard, paused with `@coderabbitai pause`, or excluded from day one by an `ignore_usernames` covering the only author or a `path_filters` matching everything. Steps 1 to 3 all post nothing in that world, so they cannot separate it from an absent installation. My table sent every one of those cases to "operator installs it".

**The fix, as you asked.** New step 4, required before any verdict: post `@coderabbitai review` on an open PR. CodeRabbit documents manual commands as working independently of auto-review configuration ([guides/commands](https://docs.coderabbit.ai/guides/commands)), so this reaches an installed app even with `auto_review.enabled: false`. The table now reads:

| Signal | Strength |
|---|---|
| **Any** reply to `@coderabbitai review`, however unhelpful | **conclusive** (installed, fault is config or dashboard) |
| Unbounded zero + control + **silence after the explicit probe** | strong, still not conclusive: not installed **or** disabled account-side |
| Unbounded zero, no explicit probe run | **inconclusive** |

The evidence is asymmetric and the doc now says so out loud: a reply proves presence outright, absence never fully proves absence, and **no row reachable without a browser is conclusive for "not installed"**. The honest session verdict is "not installed OR disabled account-side, settings page not read", and that wording now propagates to the harvest stage in `SKILL.md` and the `.coderabbit.yaml` header, which both previously asserted the org-grant cause flatly. The per-owner-account explanation is demoted to the leading hypothesis with the dashboard-disabled alternative named beside it. Both are settled at the same page, so the operator steps are unchanged.

orchestkit does satisfy the new step 4: the `@coderabbitai review` I posted here at 12:50:44Z has gone unanswered, so this repo sits on the "strong, still not conclusive" row rather than the one I originally wrote.

---

**On the red CI, and a correction to what I told you last round.** I said #3913's cause was a partial `node_modules` and proposed `rm -rf node_modules`. I ran it instead of reasoning about it, and **my own hypothesis is refuted**:

```
npm ci rc=1
node_modules after failed ci: ABSENT      <-- not partial
package-lock-only rc=0                    <-- the unmodified fallback succeeds
```

The proposed fix was a no-op. I have posted the refutation on #3913 and marked the cause UNKNOWN rather than leave a wrong diff sitting in the description for someone to land.

What still holds: the failure is not from this diff (rerunning `main`'s own green commit reproduces it), and `test-sync-versions.sh` is a cascade from the deleted `docs/site/package-lock.json`. **This PR is red and cannot merge until #3913 is fixed.** I am not shipping a speculative fix for it inside this PR to force green.

PR #3916, opened 2026-09-04
fix(announce): preflight release API credential

### LINKED ISSUE #3650: HQ_API_TOKEN is stale: the announce rail 401s on its first real fire
## What happened

v10.0.0-alpha.45 was the first release ever to pass the announce cadence gate (labeled `announce`, per #3638). The rail ran end to end and failed at the last hop:

```
[announce-release] platform returned 401: {"type":".../errors/auth_denied",
"title":"Auth Denied","status":401,"detail":"Invalid <token>",
"instance":"/api/marketing/release-announce","error_category":"auth_denied",
"retryable":false}
```

Run: https://github.com/yonatangross/orchestkit/actions/runs/32478983020

## Why this was invisible until today

Two independent failures were stacked, and the first one hid the second:

1. The cadence gate skipped every prerelease, and this repo has shipped nothing but `10.0.0-alpha.N` since the alpha series began. So the workflow never reached the POST.
2. `HQ_API_TOKEN` is stale. The secret has existed since 2026-06-26 and reads as "configured" everywhere you would look, but the platform rejects it.

A 2026-08-21 gap analysis concluded the rail was "armed, just gated shut". That was half right, and the half it got wrong is the half that matters: opening the gate alone would have produced zero drafts and a red workflow. The only way to learn this was to fire it.

## Everything upstream of the 401 is confirmed working

The run proves the whole chain except the credential:

- the `announce` label override fires (the skip step was skipped, so `skip=false` on a prerelease for the first time)
- version, release URL, and release notes are extracted correctly (visible in the run env)
- the payload composes and the request reaches the platform endpoint (a 401 is a response, not a connection failure)

## Fix

Refresh the `HQ_API_TOKEN` repo secret with a currently-valid platform `API_STATIC_TOKEN`. Note that platform PR #7363 introduced dual-accept rotation for that token, so a rotation there could have left this consumer pointing at the retired generation, which matches the symptom exactly. This is the same class as the 2026-07-09 finding where an org-level secret was not repointed after a per-project split and went stale unnoticed.

To verify after updating, re-run the workflow via `workflow_dispatch` with `version: 10.0.0-alpha.45` and `dry_run: false`; success looks like drafted ContentItems in the platform marketing queue, still awaiting the #1739 approve then approve-for-publish ladder. Nothing publishes from a green run here.

## Related

- #3638, #3639 (the label-gated announce lane)
- Yonatan-HQ/platform#10537 (platform posting its own releases through the same intake)

yonatangross:
Closes #3650

Adds a bounded authenticated dry-run preflight before the live release announce POST, makes missing or rejected credentials actionable with the canonical rotation reference, and documents the validated 1Password chokepoint procedure.

## Interactive Playground

[Credential preflight flow](https://orchestkit.yonyon.ai/lab/announce-credential-preflight.html)

Source: `docs/fix--announce-rail-token-3650/index.html`.

## Summary by CodeRabbit

* **New Features**
  * Release announcements now support authenticated preflight checks without publishing or creating drafts.
  * Dependency audits retry eligible timeout failures within configured limits and audit multiple projects concurrently.

* **Bug Fixes**
  * Improved handling of announcement network failures, timeouts, missing credentials, and unauthorized responses.
  * Dependency audits preserve non-retryable failure statuses and accurately report aggregate results.

* **Documentation**
  * Added guidance for rotating announcement credentials and manually approving live announcements.
  * Added the announcement preflight feature to the Lab gallery.

* **Tests**
  * Added regression coverage for announcement preflight and dependency audit retry behavior.

### DIFF
--- a/.github/scripts/npm-audit-with-retry.sh
+++ b/.github/scripts/npm-audit-with-retry.sh
@@ -1,0 +1,71 @@
+#!/usr/bin/env bash
+# Run npm audit with a bounded retry for a stalled registry request.
+#
+# npm audit exits nonzero both for advisory findings and for operational errors,
+# so retrying every nonzero result could hide a real critical vulnerability.
+# GNU timeout gives a distinct status for the failure seen in CI: a registry
+# request that never completes. Retry only that status and preserve all other
+# npm exit statuses unchanged.
+#
+# Usage: npm-audit-with-retry.sh [directory]
+#
+# Environment:
+#   NPM_AUDIT_TIMEOUT_SECONDS  Seconds per audit attempt (default: 60).
+#   NPM_AUDIT_MAX_ATTEMPTS     Attempts for a timeout only (default: 2).
+#   NPM_AUDIT_RETRY_DELAY_SECONDS  Delay before retrying (default: 5).
+
+set -euo pipefail
+
+AUDIT_DIRECTORY="${1:-.}"
+TIMEOUT_SECONDS="${NPM_AUDIT_TIMEOUT_SECONDS:-60}"
+MAX_ATTEMPTS="${NPM_AUDIT_MAX_ATTEMPTS:-2}"
+RETRY_DELAY_SECONDS="${NPM_AUDIT_RETRY_DELAY_SECONDS:-5}"
+
+is_positive_integer() {
+  [[ "$1" =~ ^[1-9][0-9]*$ ]]
+}
+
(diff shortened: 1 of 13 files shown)


yonatangross (commented):
KEEP: 0 findings.\n\nThe preflight turns a missing or rejected credential into an explicit non-zero failure before the live POST, sends only the platform dry-run payload, bounds the request, and contains no credential rotate/revoke/create/write path. The regression script covers missing, 401, and timeout paths.

Issue #3917, opened 2026-09-04
watch: Function Hooks (anthropics/claude-code#91870)


yonatangross:
Anthropic published an internal proposal, "Function Hooks", as anthropics/claude-code#91870 on 2026-09-03: hooks become in-process TypeScript functions shaped like Koa middleware `($, e, next)`, `$` is the only door to the world (no ambient fs or network), every method on `$` is hookable, and plugins nest by registration order. Not shipped; the author says community response decides. Within 24h the issue had 74 comments and the author replied 11 times, changing the error model once in public. OrchestKit posted the largest inventory in the thread (153 hooks.json entries, ~50k lines TS): https://github.com/anthropics/claude-code/issues/91870#issuecomment-5530551063. Four later commenters cited it.

This issue is the watch: nothing is built until a CC binary carries the feature. Full write-up with mermaid, mocks and the decision record lives in platform at `docs/playgrounds/dev/2026-09-04-cc-function-hooks-summary.html`.

## Open questions (as of 2026-09-04 12:00 IL)

| # | question | who asked | why it matters to ork |
|---|---|---|---|
| 1 | Lifecycle events (SessionStart, Stop, Compact, Subagent, ModelSwitch) have no noun: an event is a method call on `$` and nothing calls `$.session.start` | PrimeLineDirekt | 83 of our 153 entries have no address; they may stay command hooks forever |
| 2 | How command hooks and function hooks interleave on one event | us, ahjin2, DareDev256 | 153 hooks cannot move in one release; this is the migration path. Author: "tbd but top of mind" |
| 3 | When matchers are evaluated: a modifying hook that turns Write into Bash skips a Bash guard below it | gbrussich52 | 7 updatedInput handlers plus every guard beneath |
| 4 | Denied vs crashed on `err` when a plugin below throws | DolphusCY | 11 deny guards must not sniff message strings |
| 5 | Cache semantics of the `context` field (prefix rewrite per turn is a cache miss at fleet scale) | conceptDawg | 62 additionalContext handlers |
| 6 | Are the core tools at the bottom (Bash, Write, MCP) built on `$`, so capability removal constrains Claude and not only plugins | sidhartha1s | the Bash text hole; whether removal is mechanical for the agent |
| 7 | Wire format for non-JS plugins (stdio json-rpc like LSP/MCP) | ChadLatendresse | hq-ext is bash + JS |
[...]

PR #3989, opened 2026-09-08
fix(ci): report an unreadable version as SKIP, not DRIFT

yonatangross:
## Rescoped, because I made a branching mistake and the rest of this PR already merged

This PR now contains **one commit, nine lines, one file**. Everything else it used to
carry is already on `main`, and that was not intentional.

### What went wrong

I created `chore/depersonalize-public-docs-hq-refs` (PR #3990) **off this branch
instead of off `main`**. So #3990 silently carried the canary commit
`16423895274c`, and merging #3990 put the whole Function Hooks canary on `main`.

Two consequences, and the first is the serious one:

1. **The `modules`-key rule question was settled by accident.** I flagged, in this
   PR, that `.claude/rules/hooks-development.md` says unqualified *do not add a
   `modules` key to `hooks.json`*, and said explicitly that resolving it was the
   operator's call and that merging first would settle it by fait accompli. Then my
   branching error did exactly that. The fixture is on `main` and the rule question
   is still genuinely open. It should be decided on its merits, and the rule amended
   or the fixture reverted, rather than left as "well, it shipped".
2. **`main` currently carries a bug this PR fixes.** The version-parse fix was
   committed here after I parked it off #3990, so `main` has the unfixed script.

Rebasing onto `main` dropped the canary commit as "patch contents already upstream",
which is the confirmation.

### The remaining nine lines

`grep` exits 1 on no match. Under `set -euo pipefail` a failing pipeline inside a
command substitution kills the script, so the `[ -z "$CC_VER" ]` SKIP branch below it
was unreachable. Measured against `main`'s current copy with a stub CLI printing a
non-semver banner:

```
main today:  rc=1, EMPTY output
this PR:     rc=0, "SKIP: could not parse a version out of: ..."
```

The empty output is what matters. `cc-contract-probe.yml` maps a non-zero rc to
`DRIFT: the upstream contract moved`, so an unreadable version would report an
upstream contract change with nothing in the log to contradict it. A canary that
cannot distinguish "I could not look" from "the thing I watch changed" is worse than
no canary. Fixed by capturing rc around the parse rather than `|| true`, so the
could-not-observe case stays named.

That workflow runs on a daily 07:00 UTC schedule, so this is not urgent, but it is
live on `main` until this lands.

https://claude.ai/code/session_013tFBz1xmrTpEYmv27Qozjb

### DIFF
--- a/scripts/validate-fn-hooks-canary.sh
+++ b/scripts/validate-fn-hooks-canary.sh
@@ -33,7 +33,16 @@ fi
 # pass.
 MIN_CC="2.1.259"
 CC_RAW="$(claude --version 2>&1)"
+
+# Capture rc around the parse. `grep` exits 1 on no match, and under
+# `set -euo pipefail` a failing pipeline inside a command substitution kills the
+# script: measured, a non-semver banner exited 1 with EMPTY output, which the CI
+# step then reports as DRIFT. A version we could not read is its own outcome and
+# must never be indistinguishable from a contract change.
+set +e
 CC_VER="$(printf '%s' "$CC_RAW" | grep -oE '[0-9]+\.[0-9]+\.[0-9]+' | head -1)"
+set -e
+
 if [ -z "$CC_VER" ]; then
   echo "SKIP: could not parse a version out of: $CC_RAW"
   exit 0


yonatangross:
## Flagging a rule tension rather than deciding it myself

`.claude/rules/hooks-development.md` carries a Function Hooks watch gate (2026-09-05)
that says, unqualified:

> Do not add a `modules` key to `hooks.json`, enable rollout flags, or document
> either for users.

This PR adds a `modules` key to a `hooks.json`. I think it is within the spirit of
that gate and outside its letter, and that is the operator's call, not mine:

- the file is `tests/fixtures/fn-hooks-canary/hooks/hooks.json`, not
  `src/hooks/hooks.json`. It is never installed, never shipped, never loaded at
  runtime, and is not in `plugins/`.
- nothing here ports a handler or adds Function Hooks runtime code.
- **no floor is raised.** The gate's own stated risk was that a `modules` key breaks
  older CC. Measured across seven binaries while closing #3917 Q10, that premise is
  false: `modules` is a recognized key at every published version from 2.1.250 up,
  which is *below* the current 2.1.251 floor. The 2026-08-29 measurement the rule
  rests on was of an unknown **event** key, and was generalised to `modules` without
  re-testing.
- the README states plainly that this is a watch, not an adoption.

Two ways to resolve, and I have a preference but no authority here:

1. **Amend the rule** to scope the prohibition to ork's shipped `hooks.json`, and
   record that the older-CC premise was measured false. Recommended: the gate's
   purpose is served, and leaving a rule resting on a disproven premise costs more
   later than it saves now.
2. **Hold this PR** until #91870 closes or a changelog names the feature. Costs
   nothing except the drift this canary exists to catch.

Either way the rule should stop citing the `modules` key as the older-CC hazard,
since the real hazard measured is an unknown **event** key, fatal through 2.1.252
and tolerated from 2.1.257.

PR #4023, opened 2026-09-10
fix(security): npm audit gate could not run without GNU timeout, misreported as no output (#4015)

### LINKED ISSUE #4015: ci: npm audit gate red on main, 5 new moderate advisories (Hono, @vitest/mocker), no code change, fixAvailable=false
## Verdict

Third-party cause, not a code regression and not a flake. `npm audit` gate (`tests/security/test-npm-audit.sh`) failed on the identical commit that passed 3h18m earlier, because the npm registry advisory database published new advisories between the two runs.

## Evidence

Two runs of the **same commit** `bcc122d6dd2ae9caff29cbf16b941ee0f5095dc0`:

| Run | Trigger | Time (UTC) | Security Tests | CI Summary |
|---|---|---|---|---|
| [34278200519](https://github.com/yonatangross/orchestkit/actions/runs/34278200519) | push | 2026-09-08T21:01:42Z | success | success |
| [34302765093](https://github.com/yonatangross/orchestkit/actions/runs/34302765093) | schedule | 2026-09-09T02:19:40Z | failure | failure |

Failing test: `test-npm-audit.sh`, job log at [102313216217](https://github.com/yonatangross/orchestkit/actions/runs/34302765093/job/102313216217).

Still red on the next push, commit `93d264876d424742d2e6682b300743bacfd69587` (beta.11 release), run [34316824603](https://github.com/yonatangross/orchestkit/actions/runs/34316824603), same two jobs failing. This is not a one-off, the state persists until the allowlist is addressed.

5 new moderate findings, none in `.claude/audit-allowlist.json`, `fixAvailable: false` on every one (confirmed locally via `npm audit --package-lock-only --json` against each tree's lockfile):

**`src/mcp-server`** (transitive via `@modelcontextprotocol/sdk` -> `@hono/node-server` -> `hono`), same shape as the existing allowlisted entry `1124006`:
- `1193729` [moderate] Hono: Incomplete fix for CVE-2026-39408, `toSSG()` still writes files outside the output directory
- `1193730` [moderate] Hono: Unbounded dot-notation nesting in `parseBody()` can cause memory exhaustion
- `1193731` [moderate] Hono: Query parser reads parameters after the URL fragment, causing cache-key and proxy interpretation differentials

**`packages/hook-contract/examples/generic-client`** (transitive via `vitest` -> `@vitest/mocker`):
- `1193684` [moderate] Vitest: Path Traversal / Arbitrary File Read via `@vitest/mocker` Redirect Mock
- `1193683` [moderate] Vitest: Path Traversal / Arbitrary File Read via `@vitest/mocker` Redirect Mock (second `via` entry, same advisory)

## Related

Also surfaced in the same run: allowlist entry `1124006` (the existing Hono/`@hono/node-server` reachability exception) expired `2026-09-01`, 8 days ago, and now only warns, it does not yet fail the gate on its own but needs renewal or removal in the same pass.

This is the exact recurrence pattern described in #2490 (closed 2026-06-16), option 4: "transitive-only advisories with no exploit path in our usage should be allowlist-with-expiry rather than hard-fail." No code change caused this; it needs a per-advisory reachability assessment (same rigor as the `1124006` entry's writeup) before adding allowlist entries, which is a judgment call for a human/PR review rather than an automated fix.

## Suggested next step

For each of the 5 IDs, confirm reachability (is the vulnerable code path ever exercised via our transports/entry points) and either add a reasoned, dated allowlist entry in `.claude/audit-allowlist.json` or bump the dependency once a fix lands upstream. Renew or drop `1124006` in the same PR.

yonatangross:
## The issue's premise is stale, and that is the headline

#4015 reported the npm audit gate red on `main` with 5 moderate advisories. At HEAD it is
not: **PR #4014** ("bump hono from 4.13.0 to 4.13.7 in /src/mcp-server") already moved both
affected trees past every vulnerable range, and `Tests & Security` on `main` went green at
`2026-09-10T02:20:08Z`.

Verified, not assumed:

| tree | package | version at HEAD | vulnerable range |
|---|---|---|---|
| `src/mcp-server` | `hono` | 4.13.7 | 1193729 / 1193730 / 1193731 |
| `src/mcp-server` | `@hono/node-server` | 2.1.0 | `< 2.0.5` (1124006) |
| `packages/hook-contract/examples/generic-client` | `@vitest/mocker` | 4.1.11 | 1193683 / 1193684 |

```
$ curl -s -X POST https://registry.npmjs.org/-/npm/v1/security/advisories/bulk \
    -H 'content-type: application/json' \
    -d '{"hono":["4.13.7"],"@hono/node-server":["2.1.0"],"@vitest/mocker":["4.1.11"],"vitest":["4.1.11"]}'
{}
```

`npm audit --package-lock-only` over all six tracked lockfiles reports 0 findings at or
above moderate. So there is **no version bump to make here and no audit level to lower**.
What is left is the cleanup the issue itself asked for, plus one blocker found while trying
to run the gate.

## 1. Retire the dead allowlist entry (the part #4015 asked for)

`1124006` expired `2026-09-01` and is now not merely expired, it is inapplicable:
`@hono/node-server` is 2.1.0 and the advisory covers `< 2.0.5`. Deleting it rather than
renewing it. The allowlist is now empty, which is the goal state.

The `_schema` description gains the missing half of the contract: extend a date only with a
fresh reachability writeup, delete the entry once the dependency has moved past the range.

`tests/security/test-npm-audit.sh` gets a one-line guard for that goal state. Under `set -u`
a bare `"${ACTIVE_IDS[@]}"` on an empty array is an unbound-variable error in bash before
4.4, so retiring the last exception could have aborted the gate on exactly the day it
succeeded.

## 2. The gate could not be run locally at all, and it said so in the wrong words

`.github/scripts/npm-audit-with-retry.sh` called GNU `timeout` unconditionally. That binary
ships on the CI runners and does **not** ship on a stock macOS, so the shell exited 127
before `npm` ever ran, and the gate reported this for all six lockfiles:

```
  x root: npm audit produced no usable JSON report
      /.../npm-audit-with-retry.sh: line 58: timeout: command not found
```

A missing binary wearing a registry outage's costume. It is not cosmetic: the gate is part
of `tests/security/run-security-tests.sh`, which `bin/git-hooks/pre-push` invokes at :356,
so any macOS run of the security suite reported six phantom outages and no advisory data.

One correction to that, found while pushing this branch and worth stating plainly rather
than leaving in the PR as a claim I did not check: the hook did **not** execute on this
push. `bin/git-hooks/pre-push:65` exits 0 for every branch matching
`VERSION_SKIP_PATTERN`, which is every conventional-commit prefix, so the version-bump
stand-down takes the whole hook with it. That is a separate pre-existing defect and gets
its own issue, not a quiet fix here.

The wrapper now resolves `timeout`, then `gtimeout` (coreutils), then falls back to a bash
watchdog that maps its SIGTERM kill onto GNU's exit 124, so the retry decision above it
reads an identical status whichever path ran.

Why no test caught this: `tests/ci/test-npm-audit-with-retry.sh` puts a **fake `timeout` on
PATH** in every case, which is precisely what made the unconditional call invisible. Two
cases are added that run with a PATH carrying neither `timeout` nor `gtimeout`, plus a
positive control asserting the isolated PATH really removed both, so the new cases cannot
quietly degrade into re-testing the shimmed path.

## Evidence

```
$ bash tests/ci/test-npm-audit-with-retry.sh
  PASS: retries a timed-out registry request twice
  PASS: passes through the lockfile audit arguments
  PASS: does not retry a critical-advisory exit
  PASS: fails after three bounded attempts
  PASS: isolated PATH exposes neither timeout nor gtimeout
  PASS: runs npm and preserves its status with no timeout binary present
  PASS: bounds a hung audit and reports GNU timeout's 124 without GNU timeout
7 passed, 0 failed
```

```
$ bash tests/security/test-npm-audit.sh
  (no expired-allowlist warning)
  root: 0 findings >= moderate
  docs/site: 0 findings >= moderate
  src/hooks: 0 findings >= moderate
  src/mcp-server: 0 findings >= moderate
  orchestkit-demos: 0 findings >= moderate
  hook-contract/examples/generic-client: 0 findings >= moderate
  Total: 6  |  Passed: 6  |  Failed: 0
```

That second block is the first time this gate has been runnable on this machine.

## Interactive Playground

`docs/fix--4015-audit-gate-hygiene/audit-gate-timeout.html`, published to the Lab at
https://orchestkit.yonyon.ai/lab/audit-gate-timeout.html

The before and after transcripts of the gate, and why the one test written to cover the
wrapper could not see the defect.

Closes #4015


yonatangross (commented):
## Reviewer pass on c3e945fc1

Reviewed by **orchestkit's** rules. Tripwires clean: `hooks/dist` 0, `plugins/` 0, conventional-commit title for release-please.

### The standout

```bash
# An empty allowlist is the goal state, not an edge case ... under `set -u` a
# bare "${ACTIVE_IDS[@]}" on an empty array is an unbound-variable error in
# bash < 4.4, which would abort the gate on the exact day the last exception
# is retired.
[[ "${#ACTIVE_IDS[@]}" -eq 0 ]] && return 1
```

That is a failure timed to arrive at the moment of success, and it was caught before shipping rather than on the day it fired. The same instinct shows in the `timeout` work: GNU `timeout` is not on macOS, so there is a bash watchdog reproducing exit 124, with a test that runs **without** GNU timeout and asserts a caller cannot tell the two apart.

### One finding

`PR Playground` is **failing** on this head. This repo's `.gitignore:12` states the job requires HTML at `docs/<branch-slug>/`, and this PR's four files carry no `docs/` path. The branch is `fix/4015-audit-gate-hygiene`, so the expected path is `docs/fix--4015-audit-gate-hygiene/`.

I am not calling that a BLOCK because I do not know whether that job is required for merge here or advisory for non-UI changes, and a security-gate fix plainly has nothing to show in a playground. Either it needs a stub page or the job needs an exemption for PRs that touch no UI, and that is a repo-owner call rather than a reviewer one.

Everything else reads clean. The retired hono exception, the retry wrapper, and two new test files, one for the wrapper and one for the gate.

**LAND on content**, with the PR Playground question above for whoever owns that job.

yonatangross (commented):
## Reviewer pass on b1307a345 (delta only, the playground stub)

Delta from `c3e945fc1` is one commit adding the PR Playground page and its lab wiring. The fix itself is untouched, so the earlier LAND carries. Checked the page against both bars asked for:

```
em-dash or en-dash        0
double-hyphen in prose    0   (the --pg-* CSS custom properties are not prose)
PR Playground check       success
```

**No claims beyond the PR.** The page says the gate "failed closed, six for six, which is the right direction but the wrong sentence: nobody reading it would go looking for coreutils." That is the accurate characterisation, including the part I got wrong earlier today when I read a source comment and called this a false green. It is not; it failed closed with a misattributed cause, and the page says so plainly rather than inflating it.

Committing `docs/site/lib/generated/lab-data.ts` is correct here and worth naming, because it looks like the thing this repo forbids and is not. Generated output is staged with its source change (the `plugins/` rule); only `hooks/dist` is release-owned and off limits. Two opposite treatments of generated files in one repo, so it is worth being explicit rather than pattern-matching on the word "generated".

**LAND at `b1307a345`.**

Issue #3971, opened 2026-09-07
dependabot PRs touching src/hooks/package.json can never go green: the tracked plugins/ mirror trips the build-drift gate


yonatangross:
## Symptom

Every dependabot PR that bumps a dependency in `src/hooks/package.json` fails the required
**Build** check at "Check for uncommitted build changes", so the `src/hooks` dependabot lane
cannot land without hand intervention. Observed on #3957 (rebased onto current `main`,
behind=0, head `85452971`):

```
Build drift detected! Run 'npm run build' and commit the changes
(plugins/ :!plugins/ork/hooks/dist README.md docs/site/lib/generated/ plugin.json).
 plugins/ork/hooks/package.json | 4 ++--
```

## Cause

`plugins/ork/hooks/package.json` is a **tracked** generated mirror of
`src/hooks/package.json`, produced by `npm run build`. The drift roster at
`.github/workflows/ci.yml:100` is `plugins/` minus only `plugins/ork/hooks/dist`, so the
mirror is inside the gate.

Dependabot edits the source manifest and, by design, never runs a project build. The mirror
therefore stays at the old versions and the gate fires. The gate is correct; the automation
simply cannot satisfy it.

Confirmed at HEAD: `src/hooks/package.json` and `plugins/ork/hooks/package.json` both carry
`@biomejs/biome ^2.5.11` and `@types/node ^26.4.0`, so the mirror is a real duplicate of the
version strings, not an incidental copy.

## Why it matters

This is a silent lane blockage rather than a visible one. The PRs stay open looking like
ordinary red CI, and the natural reading is "flaky, rebase it", which never helps. It also
means `src/hooks` dev dependencies drift further behind over time while other manifests
stay current.

## Options, roughly in order of appeal

1. **Exclude the generated manifest from the drift roster**, the way
   `plugins/ork/hooks/dist` already is: add `':!plugins/ork/hooks/package.json'`. Cheapest,
   but it removes a real check on a tracked file.
2. **Auto-regenerate on dependabot branches.** The repo already runs a Skill Build Autoheal
   bot that pushes regen commits to PR branches; extending it to run `npm run build` and
   commit the `plugins/` diff would fix this class without weakening the gate.
3. **Stop tracking the mirror** and generate it at package time. Largest change, removes the
   duplication at the source.
4. **Accept it** and treat `src/hooks` bumps as always-manual, documented in
   `CONTRIBUTING.md` so the next person does not re-diagnose it.

[...]

Issue #3578, opened 2026-08-19
Committed dist/ bundles make every hooks PR conflict-prone, and a conflicting PR gets zero CI


yonatangross:
## Symptom

Any PR that changes hook source cannot stay mergeable long enough to get CI, because it must commit regenerated bundles that `main` also regenerates on every hooks merge.

Measured on #3570 today:

| step | result |
|---|---|
| rebase onto `origin/main`, rebuild, push | `9c1f8218c`, clean, 0 dirty |
| `main` advances to `a4da29fa9` (a hooks fix) | gains 7 files |
| of those 7 | **4 are the same bundles** the branch commits |
| PR state | back to `mergeable: false`, `dirty` |

The four are `plugins/ork/hooks/dist/{pretool.mjs,bundle-stats.json}` and `src/hooks/dist/{pretool.mjs,bundle-stats.json}`.

## Why this is worse than a normal conflict

A conflicting PR has no computable merge ref, so GitHub fires **no `pull_request` event and runs no CI at all**. Not a red check, not a failure, just silence that is indistinguishable from a slow queue.

The run counts on #3570's own branch show it cleanly:

```
50d05820a  pull_request         x14   <- merge ref existed
c1e4b636c  pull_request         x0    <- conflicting: dead
c1e4b636c  pull_request_target  x2    <- builds vs BASE, unaffected
```

`pull_request_target` surviving while `pull_request` goes to zero is the fingerprint, and it is what distinguishes this from an Actions outage. GitHub status was fully operational throughout.

## Why rebasing is not the fix

It is a race against the next hooks merge, and this repo merges those often. Today: rebase landed, `main` moved in the gap, re-conflicted immediately. Three re-trigger attempts before that (empty commit, close, reopen) also did nothing, because the event cannot be produced without a merge ref.

## The shape

This is the same defect class as a shared changelog anchor: **a high-churn generated file that every PR must touch makes conflicts structural rather than occasional.** The sibling repo hq-ext hit exactly this on `CHANGELOG.md` (5 PRs, one of which lost all CI for 34 minutes) and fixed it structurally by giving each PR its own file, so two PRs never touch one path.

The same reasoning applies here, but the artifact is generated, so the fix is different in kind: stop requiring it in-tree.

## Options

1. **CI rebuilds and verifies rather than diffing committed artifacts.** The build job already runs; have it build and compare against source rather than asserting the committed bundle matches. Removes the conflict surface entirely.
[...]

PR #4028, opened 2026-09-10
fix(hooks): migrate the src/hooks suite to vitest 5 (#3981)

### LINKED ISSUE #3981: vitest 5 breaks the src/hooks suite: bench removed, describe.sequential removed, and three TS2883 mock helpers
## What this is

`vitest` 5 is now on the root (#3960 `104afcfb1`), `/docs/site` (#3963 `63a87f338`) and
`/orchestkit-demos` (#3978). `/src/hooks` cannot follow: the bump lands as #3980 (the
grouped `vitest` + `@vitest/coverage-v8` PR that #3977 made possible) and it is red on real
API breakage, not on plumbing.

This is a migration task, not a dependency bump. Filing it so #3980 is not repeatedly
rebased in the hope that it goes green.

## The breakages, measured on #3980 head `eafa13b98`

**1. `bench` is gone from the `vitest` root export.** Runtime and types agree.

```
TypeError: bench is not a function
 ❯ src/__bench__/secret-handler.bench.ts:103:3
   103|   bench('5KB - PAT embedded', () => {
```
```
src/__bench__/secret-handler.bench.ts(59,10): error TS2724:
  '"vitest"' has no exported member named 'bench'. Did you mean 'Bench'?
```

**2. `describe.sequential` removed.**

```
src/__tests__/async/dispatcher-e2e.test.ts(122,10): error TS2339:
  Property 'sequential' does not exist on type 'SuiteAPI'.
```

**3. Mock helper return types are no longer nameable (TS2883), three sites.**

```
src/__tests__/fixtures/mock-common.ts(291,17)  'mockLog'
src/__tests__/helpers/mocks.ts(33,17)          'createCommonMock'
src/__tests__/helpers/mocks.ts(107,17)         'createChildProcessMock'
  The inferred type ... cannot be named without a reference to 'Procedure'
  from 'node_modules/vitest/dist/chunks/config.d.CU_b-wJj.js'
```

The third class is the awkward one: the fix is an explicit return type annotation on each
helper, but the type it needs to name lives in a hashed internal chunk path, so the
annotation has to be written against a stable public type rather than the inferred one.

## Failing checks

`hook-throughput` (Hook Benchmarks, "Run benchmarks") and `Hook smoke test (PowerShell tool
enabled)` (TypeScript typecheck). `Build` is also red, but from the separate #3971 cause,
and `CI Summary` is downstream of those.

## Why the other manifests were fine

They do not use `bench`, `describe.sequential`, or the mock-helper pattern. `/src/hooks` is
the only tree with a benchmark suite and the shared mock factories, so it is the only one
that touches the removed surface.

## Suggested order

1. `bench`: move `src/__bench__/*.bench.ts` to the vitest 5 import, or drop the benchmark
   suite if `hook-throughput` is better served by the existing perf tests.
2. `describe.sequential`: replace with the vitest 5 equivalent for serial execution.
3. TS2883: annotate the three helpers with explicit public return types.
4. Then let Dependabot re-propose the group, or rebase #3980.

#3980 stays open and held until then. #3971 must also be resolved for it to go green, since
`/src/hooks` bumps additionally trip the tracked `plugins/` mirror drift.

Refs #3980, #3977, #3971.

yonatangross:
## What this is

`src/hooks` is the only tree in the repo with a benchmark suite and the shared mock
factories, so it was the only manifest that could not follow the vitest 5 bump the other four
already took (#3960, #3963, #3978). Verified against a real **vitest 5.0.0** install rather
than the issue's transcript, which is how the fourth cause below surfaced.

Five type errors, three causes in the issue, **plus one it did not list**.

| Removed in vitest 5 | Replacement used here |
|---|---|
| `import { bench } from 'vitest'` | test-context fixture, `test(name, async ({ bench }) => ...)` |
| `describe.sequential` | plain `describe`, serial is the default |
| bare `vi.fn()` (5 sites) | `vi.fn<(...args: unknown[]) => void>()` |
| `benchmark.outputJson` | `--reporter=json --outputFile=` on the `bench` script |
| `benchmark.reporters` | top-level `test.reporters` |

## The part worth reading: options moved to the second argument

```ts
// v4, options third
bench('5KB - PAT embedded', () => { ... }, OPTS);

// v5: bench(name, BenchFnOptions, fn), and the sampling window is not
// BenchFnOptions at all. time/warmupTime/warmupIterations are tinybench
// BenchOptions, which the rewritten API takes at RUN time.
await bench('5KB - PAT embedded', () => { ... }).run(OPTS);
```

Passing the sampling window in the registration slot fails with `TS2559` "no properties in
common". That is the good outcome. The bad one would have been a silent fall back to default
sampling, leaving a benchmark that still prints numbers while no longer measuring what its
own header documents.

The three payload sizes stay in a single `bench.compare()` so they still render as one table.
The file header says the portable regression signal is the **ratio** between scenarios, not
any absolute number, and a shared run window is what makes that ratio trustworthy.

## The cause the issue did not list

`vitest.bench.config.ts` set `benchmark.outputJson` and `benchmark.reporters`. Both are gone
from `BenchmarkUserOptions` in 5, which now carries only
`enabled / include / exclude / includeSource / retainSamples / provider / suppressExportGetterWarnings`.

Left in place they would be **ignored config, not an error**, and
`.github/workflows/bench.yml` uploads `bench-results.json` as a 30-day artifact. The bump
would have compiled, run, and quietly stopped producing the artifact.

## TS2883 had a smaller fix than proposed

The issue suggested annotating the three helpers' return types, awkward because the inferred
type names `Procedure` from a hashed internal chunk path. The trigger is narrower than the
symptom: only a **bare** `vi.fn()` infers `Mock<Procedure>`. `createFsMock` and
`createPathMock` sit in the same files and never errored, because every one of their mocks
has an implementation. Typing the five bare calls removes the unnameable reference with no
return-type annotation and no cast.

## Verification

Typecheck, against the real install:

```
before: 5 errors
after:  0 errors
```

Benchmarks actually run and produce samples, checked rather than assumed:

```
$ npm run bench
rc=0
JSON report written to .../src/hooks/bench-results.json
bench-results.json 5444 bytes
```

All four benchmarks carry real `latency.p50 / p99 / rme / samplesCount` and `throughput`.

Suite, measured against a clean detached `origin/main` worktree on the same machine:

| | vitest | files failed | tests failed | passed |
|---|---|---|---|---|
| main `b37f3db92` | 4.1.11 | 7 | 24 | 6615 |
| this branch | 5.0.0 | **6** | **23** | **6616** |

The failing set on this branch is a strict **subset** of main's, so the bump introduces no
failure and clears one (`precompact-stdout-contract.test.ts`). The remaining 6 are
environmental on a dev machine: a missing `plugins/ork/hooks/dist/*.map` (release-owned,
never committed, #3578) and `git commit` inside test fixture repos.

## Note for #3971

`npm run build` regenerated `plugins/ork/hooks/package.json` and
`plugins/ork/hooks/vitest.bench.config.ts`, and both are committed here. That is exactly the
mechanism #3971 describes: a human PR can run the build, Dependabot cannot, which is why the
`src/hooks` dependabot lane cannot go green on its own. `plugins/ork/hooks/dist/` was not
touched.

## Interactive Playground

`docs/fix--3981-vitest5-hooks/vitest-5-bench-rewrite.html`, published to the Lab at
https://orchestkit.yonyon.ai/lab/vitest-5-bench-rewrite.html

The before and after of each removed API, the argument-position trap, and the baseline
comparison table.

## Deliberately not fixed

Vite warns that `__dirname` in `vitest.bench.config.ts` will stop working under
`configLoader: 'native'`. Real, unrelated to vitest 5, and not folded into a migration PR.

Closes #3981

## Summary by CodeRabbit

- **New Features**
  - Added a new benchmark lab entry covering Vitest 5 hooks and migration benchmarks.
  - Benchmark runs now produce JSON result files for easier analysis.

- **Tests**
  - Updated benchmark scenarios to use Vitest 5 APIs, including comparison benchmarks.
  - Preserved serial execution for end-to-end tests.
  - Improved typing for test mocks without changing runtime behavior.

- **Documentation**
  - Updated benchmark guidance to reflect Vitest 5 discovery and reporting behavior.


yonatangross (commented):
## Reviewer pass on 8757d37b

Reviewed by orchestkit's rules. Tripwires clean: `hooks/dist` 0, `plugins/` 2 (regenerated build output, correctly staged), playground present, conventional title for release-please.

**The claim worth verifying was the bench artifact**, because "removed in v5 and were IGNORED CONFIG" means `bench-results.json` would have stopped being written with nothing failing. That is a false green in an artifact pipeline, and the fix is real rather than documented:

```diff
- "bench": "vitest bench --run --config vitest.bench.config.ts",
+ "bench": "vitest bench --run --config vitest.bench.config.ts --reporter=json --outputFile=./bench-results.json",
```

Present in both `src/hooks/package.json` and the regenerated `plugins/ork/hooks/package.json`, so the built plugin carries it too. The playground page documents the mapping, but the script is where it counts and the script changed.

| dimension | score |
|---|---|
| correctness | 9 |
| root cause | 10 |
| tests | 8 |
| security | 9 |
| scope | 9 |
| hygiene | 10 |

**LAND**, with one limit stated.

### What I could not verify

The subset claim, that the branch's 6 failing files / 23 tests are a strict subset of main's 7 / 24, is the evidence that this migration regressed nothing. I could not re-run it: this sandbox has no `node_modules` for the hooks suite and refuses writes under `hooks/`, so a real vitest 5 install is not available to me. That number is the lane's measurement and I am recording it as theirs. Correctness is 9 rather than 10 for that reason alone, not because I found anything.

yonatangross (commented):
## Reviewer pass on d28f5f02 (delta only)

One commit, `chore(deps): unpause vitest major now that the migration lands`, removing 24 lines from `.github/dependabot.yml`. The migration itself is untouched, so the earlier LAND carries.

The risk on a change shaped like this is that removing an `ignore:` block unpauses more than intended. It does not here. The removed block held exactly two entries:

```yaml
ignore:
  - dependency-name: "vitest"
    update-types: ["version-update:semver-major"]
  - dependency-name: "@vitest/*"
    update-types: ["version-update:semver-major"]
```

Both vitest-scoped, both major-only. Everything else in the file is untouched, and no other dependency changes posture. Twenty of the twenty-four removed lines were the comment explaining why the pause existed, which is why the diff looks larger than the behaviour change.

Removing it in the same PR as the migration is the right order: the pause existed because the major was unlandable by Dependabot, and that stops being true the moment this lands.

**LAND at `d28f5f02`.**

PR #4029, opened 2026-09-10
fix(build): strip devDependencies from the hooks manifest mirror (#3971)

### LINKED ISSUE #3971: dependabot PRs touching src/hooks/package.json can never go green: the tracked plugins/ mirror trips the build-drift gate

yonatangross:
## The problem

`plugins/ork/hooks/package.json` is a tracked generated mirror of `src/hooks/package.json`,
and it sits inside ci.yml's build-drift roster (`plugins/` minus only
`plugins/ork/hooks/dist`). Dependabot edits the source manifest and by design never runs a
build, so every `src/hooks` bump left the mirror at the old versions and failed the required
**Build** check.

The gate was correct. The automation could not satisfy it. The whole `src/hooks` Dependabot
lane was landable by hand only, and it read as ordinary flaky CI, so the natural response was
to rebase, which never helped.

## Both options the issue preferred are already rejected in this repo's code

This is the part worth reviewing, because the issue was filed without these in view.

**Option 2 (extend the autoheal bot to Dependabot), the issue's pick**, is rejected at
`.github/workflows/skill-autobuild.yml:59` and enforced at the job's `if:`:

> DEPENDABOT IS EXCLUDED ON PURPOSE. Dependabot-triggered runs read from GitHub's separate
> *Dependabot* secret store, so `secrets.RELEASE_BOT_*` resolve to empty strings and the
> token mint below dies in ~4s with "'app-id' input must be set to a non-empty string". That
> surfaced as a red `autoheal` check claiming the heal failed when it had never run, which is
> actively misleading during triage.

**Option 1 (drop the mirror from the drift roster)** leaves it stale permanently.
`release-please.yml:314`'s `ALLOWED` regex does not list this path, and the release rebuild
commits only `DIST=(src/hooks/dist plugins/ork/hooks/dist)`. So the release path never
repairs the mirror either, and an unguarded stale file would ship indefinitely.

## The fix

The duplication was never load-bearing. The mirror has **zero runtime dependencies**: the
plugin ships prebuilt `dist/` plus `bin/run-hook.mjs`, and every `devDependencies` entry
(biome, typescript, vitest, esbuild, fast-check, `@types/node`) is build-time tooling that
nothing installs from there. I grepped the tree for readers of that field and found none
outside comments in `dependabot.yml`.

So `scripts/build-plugins.sh` deletes the field when it builds the mirror:

```bash
STRIPPED=$(jq 'del(.devDependencies)' "$HOOKS_MANIFEST")
```

That removes exactly the version strings that drifted. It sits directly below the
`package-lock.json` exclude the same rsync has carried for the same reason: same reasoning,
applied to one field instead of one file.

**Nothing is weakened.** The drift gate still guards every shipped byte, including this
file's remaining fields.

## Proof, end to end

Bump every devDependency in the source manifest, rebuild, count the drift:

```
baseline plugins/ clean: YES
bumped ALL devDependencies in the source manifest
build rc=0
RESULT: plugins/ files changed by the bump = 0   (expect 0)
  -> #3971 class CLOSED
restored. tree clean: YES
```

## The ratchet

`tests/plugins/test-hooks-manifest-mirror.sh`, three assertions:

```
PASS plugins/ork/hooks/package.json carries no devDependencies
PASS plugins/ork/hooks/package.json keeps name/version/type/bin/engines
PASS bumping all 7 devDependencies produces an identical mirror
```

The second exists so a future "just exclude the whole file" cannot pass by deleting
everything, taking `type: module` and the bin mapping with it. The third is the one that
matters: without a positive control, this test would still pass if the strip were removed and
someone hand-edited the mirror once, which is precisely the failure mode that produced the
issue. It also guards its own vacuity, failing if the source manifest declares no
devDependencies at all.

No roster edit needed: `tests/run-all-tests.sh:329` (`run_dir "Plugins" "plugins"`) and
`ci.yml:873` both discover `tests/plugins` by directory.

## Follow-up, not fixed here

The same rsync ships the test configs (`vitest.bench.config.ts` and friends) into the plugin,
which is dev tooling users never run. Filed separately rather than folded in.

## Interactive Playground

`docs/fix--3971-mirror-devdeps/the-gate-nobody-could-satisfy.html`, published to the Lab at
https://orchestkit.yonyon.ai/lab/the-gate-nobody-could-satisfy.html

Closes #3971

## Summary by CodeRabbit

* **New Features**
  * Added the “The Gate Nobody Could Satisfy” playground to the Lab gallery, with searchable metadata and tags related to Dependabot, CI, builds, false signals, and hooks.

* **Bug Fixes**
  * Improved plugin hook manifest generation to prevent development-only dependency versions from appearing in mirrored manifests, reducing version drift during builds.

* **Tests**
  * Added automated checks to verify mirrored hook manifests retain required package information while excluding development dependencies.


yonatangross (commented):
## Reviewer pass on 41fa7dcd

`del(.devDependencies)` strips exactly one key. No wildcard, no filter that could take a neighbouring field with it, and `hooks/dist` is untouched.

The ratchet is the strongest test design I have read today, because each control names how the previous one could be defeated:

```
1  the shipped mirror carries no devDependencies        the property
2  it still carries name/version/type/bin/engines       so 1 cannot pass
                                                        by deleting everything
3  bump EVERY devDependency, require a byte-identical    reproduces the actual
   mirror                                               #3971 scenario
```

Control 2 is the one I would have asked for. "Asserts no devDependencies" passes trivially against an empty file, and without it the ratchet would green on a mirror that had lost `bin` and `engines` too. Control 3 is the difference between asserting the current state and asserting the decoupling: it changes the input that used to move the mirror and requires the mirror not to move.

Rejecting options 1 and 2 with file:line in the body is the right shape as well. A reader six months out can check whether the rejection still holds rather than taking it on trust, and #4030 keeps the shipped-test-configs question from being silently folded into this one.

| dimension | score |
|---|---|
| correctness | 10 |
| root cause | 10 |
| tests | 10 |
| security | 9 |
| scope | 10 |
| hygiene | 10 |

**LAND.**

PR #4105, opened 2026-09-14
feat(site): enable PostHog session replay on /docs/reference/* only

yonatangross:
## Decision

Owner decision (2026-09-14): enable PostHog session replay on the docs site, but only on `/docs/reference/*`. Skill and hook reference pages bounce at about 100%, and events alone cannot say why. Recording anywhere else is out of scope.

## How recording is gated

- `posthog.init` keeps `disable_session_recording: true`. Nothing records by default on any path.
- `docs/site/lib/replay-gate.ts` is the only caller of `startSessionRecording()`, and only when the pathname starts with `/docs/reference/`.
  - Full page load: checked once, right after init.
  - App Router navigation: `onRouterTransitionStart` in `instrumentation-client.ts`. Leaving the reference tree calls `stopSessionRecording()` synchronously, before the next page renders. Entering it starts replay once the URL has committed, so the recording opens on the reference page, not the one being left.
- `startSessionRecording()` is called with no override, so project sampling, linked flag and URL triggers still apply on top. The gate can only narrow recording.
- **Two layers, and the in-code gate is complete on its own.**
  1. In code (this PR): replay is disabled at init and only this gate turns it on, for `/docs/reference/*`.
  2. In the PostHog project (owner, configured separately): a replay trigger group "Docs reference pages" with URL regex `^https://orchestkit\.yonyon\.ai/docs/reference/.*$`, recording domains limited to orchestkit.yonyon.ai, canvas capture off, 2s minimum duration.
  If the project layer is ever loosened or removed, the code still never starts replay off the reference tree.
- **Replay also needs session replay enabled in the PostHog project settings (Settings, Session replay).** posthog-js requires that server-side flag as well as this gate; with it off, nothing records even on reference pages.

## Privacy safeguards

- `session_recording.maskAllInputs: true`, set explicitly so a remote "Privacy and masking" change cannot unmask inputs.
- `ph-no-capture` on the search dialog content (`components/search-dialog.tsx`). The typed query, the result list and the zero-results echo of the query are replaced by a placeholder. This also opts the dialog out of autocapture; result clicks still reach PostHog through `reportSearchResultClicked`'s first-party mirror.
- `maskTextSelector: '.ph-no-capture, [role="dialog"]'` as a second line, in case a fumadocs upgrade stops forwarding the className.
- Network capture: `recordHeaders: false`, `recordBody: false`, and `maskCapturedNetworkRequestFn` drops every header and body and strips query strings and fragments from captured URLs. The search fetch carries the query in its query string, so the URL has to go too.

## CSP change

```diff
+ "worker-src 'self' blob:",
```

The replay recorder starts a Web Worker from a `blob:` URL (confirmed in `posthog-js/dist/lazy-recorder.js`). Without `worker-src` the browser falls back to `script-src`, which has no `blob:`, and refuses the worker. This grants `blob:` to workers only. `script-src` is unchanged.

## Limits

- Page text on reference pages is recorded (public docs content).
- The gate is per tab.
- The start after a client navigation waits for an animation frame, so in a tab that is not rendering frames the start is deferred until it is. Stop never waits.

## Verification

- TypeScript typecheck, no emit (docs/site `tsc`): exit 0.
- `vitest run` (docs/site): 56 files, 627 tests passed, including 20 new tests in `__tests__/replay-gate.test.ts` (path gate, dot-segment normalisation, start/stop idempotence, network masking).
- `npm run build` (docs/site, webpack): exit 0, TypeScript pass, 322/322 static pages. The only warnings come from Next and fumadocs internals.
- Lab: `lab-manifest.mjs` regenerate plus its check mode report `151 fragment(s) == lab-data.ts == public/lab/`.
- Browser, `next start` with a throwaway key, behind a local mock of PostHog remote config reporting replay as enabled. Nothing reached PostHog.
  - Served CSP: `... script-src 'self' 'unsafe-inline'; worker-src 'self' blob:; ...`. No "Content Security Policy" console messages on `/` or on reference pages.
  - Full load `/`, `/docs/foundations/overview`, `/docs/reference` (index): no recorder script, 0 snapshot requests.
  - Full load `/docs/reference/agents/ci-cd-engineer` and `/docs/reference/skills`: `lazy-recorder.js` loaded.
  - Client navigation `/docs/reference` to `/docs/reference/skills`: `lazy-recorder.js` fetched 77 ms after the click.
  - Search dialog open on a reference page: `ph-no-capture` present, input inside the blocked subtree.
  - Client navigation from `/docs/reference/skills` to `/` at 07:24:06.058Z: the one snapshot POST after that (sent 07:24:09) decoded to 37 rrweb items from 07:23:49.735Z to 07:24:06.058Z, Meta href only `/docs/reference/skills`. That is the buffer flush of pre-stop data; nothing was captured on `/`, and no further snapshot POSTs arrived.

## Interactive Playground

https://orchestkit.yonyon.ai/lab/replay-only-where-readers-bounce.html

Type a path to see whether it records, and toggle each privacy layer to see what a recording would contain. Registered in the Lab manifest as `replay-only-where-readers-bounce`.


yonatangross:
@coderabbitai review

yonatangross:
**reviewer-y1g verdict on head `db8d758119234744a48074ef835aff2b98edcce5`: HOLD**

The code is careful and I found nothing wrong with the gate itself. The hold is one missing user-facing change: this PR starts recording sessions while the site's published privacy policy still tells readers it does not.

- The blocker. `docs/site/app/(home)/privacy/page.tsx` is untouched by this PR (diffing that path between main and this head returns nothing). It currently states the site "collects anonymous, aggregate usage analytics (such as page views)" (line 20) and, under "Cookies and tracking", that storage is limited to a theme preference in local storage (line 41). A session replay is a per-session recording of pointer movement, clicks, scroll and page DOM. It is anonymous here, but it is not aggregate and it is not a page view, so the page as written no longer describes what the site does on `/docs/reference/*`.
- Exact fix. In `docs/site/app/(home)/privacy/page.tsx`, under "The documentation site", add a sentence along the lines of: on reference documentation pages under `/docs/reference/`, the site records an anonymous session replay (pointer movement, clicks and scrolling) to understand why those pages are abandoned; all form and search input is masked, the search dialog is excluded entirely, and no request headers, bodies or query strings are captured. Nothing is recorded anywhere else on the site. Re-request review after that lands and I will flip this to LAND.
- Everything else verified and good. All 16 required contexts are green on this head (head-scoped `check-runs`, newest per name).
- The gate is genuinely complete in code, as claimed. `instrumentation-client.ts:71` keeps `disable_session_recording: true`, and `lib/replay-gate.ts:20` is the only caller of `startSessionRecording()`, behind `startsWith("/docs/reference/")`. `startSessionRecording()` takes no override, so project sampling and triggers still apply on top; the gate can only narrow. The `requested` flag rather than `sessionRecordingStarted()` correctly covers the window where the SDK is still awaiting remote config.
- Privacy controls hold up. `maskAllInputs: true` is set locally so a remote masking change cannot unmask it, `ph-no-capture` blocks the search dialog subtree (PostHog's default `blockClass`), `maskTextSelector: '.ph-no-capture, [role="dialog"]'` is a real second line if fumadocs stops forwarding the className, and `maskNetworkRequest` (`lib/replay-gate.ts:82`) drops headers and bodies and strips query and fragment from the URL, which is necessary since the search fetch carries the query there.
- The CSP change is correctly scoped. `worker-src 'self' blob:` grants `blob:` to workers only and leaves `script-src` untouched (`next.config.mjs:242`). It also does not regress any existing same-origin worker, since `'self'` is retained.
- Generated files regenerate byte-clean: `node docs/site/scripts/lab-manifest.mjs` on this head leaves 0 dirty files, `--check` reports `151 fragment(s) == lab-data.ts == public/lab/`.
- Merge order: rebase behind #4101. Tested, merging this branch after #4101 is textually clean but then fails the new Lab check with `public/lab/replay-only-where-readers-bounce.html differs from a fresh render (missing the lab chrome)`. Rerun the generator after the rebase.

yonatangross:
**reviewer-y1g verdict on head `9ec2e8a22b4ff527b90a565b212792ec9f8449cc`: LAND**

The hold is cleared. The privacy disclosure now matches what the code does, and I checked each sentence against the implementation rather than against the commit message.

- **The fix is additive and nothing else moved.** `db8d7581` is an ancestor of this head; the only new commit is `9ec2e8a2`, and the full diff between the two heads is `docs/site/app/(home)/privacy/page.tsx | 8 ++++++++`, one paragraph, zero deletions. Every finding I verified at `db8d7581` therefore still stands unchanged.
- **Placement is right.** The paragraph sits inside the `<h2>The documentation site</h2>` section, immediately after the "anonymous, aggregate usage analytics" sentence it was qualifying, so a reader hits the correction at the point of the claim rather than further down the page.
- **Each claim checks out against the code.** "under `/docs/reference/`" matches `REPLAY_PATH_PREFIX` in `lib/replay-gate.ts:18`. "All form and search input is masked" matches `maskAllInputs: true`. "the search dialog is excluded entirely" matches `ph-no-capture` on `SearchDialogContent` (PostHog's default `blockClass`) with `maskTextSelector: '.ph-no-capture, [role="dialog"]'` behind it. "no request headers, bodies or query strings are captured" matches `recordHeaders: false`, `recordBody: false` and `maskNetworkRequest` at `lib/replay-gate.ts:82`, which nulls both and strips query and fragment. "Nothing is recorded anywhere else" matches `disable_session_recording: true` at init plus the gate being the only caller of `startSessionRecording()`.
- **No second surface went stale.** `/privacy` has no markdown or llms twin (`lib/page-markdown.ts` covers the homepage, `/about` and `/yonyon` only), so there is no agent-facing copy still carrying the old text. `__tests__/agent-surface-honesty.test.ts:110` only asserts the page declares `alternates`, which is untouched, so no test needed updating.
- **CI at this head:** 43 check names, 0 pending, 0 red, from a head-scoped `check-runs` read at 08:53Z by the operator driving the merge. At my own earlier read of this head 14 of the 18 required contexts were already green with the rest still running; the code verification in this comment is mine, that one line is relayed.
- Optional, not a blocker: the paragraph describes pointer, clicks and scrolling but not that the page's own text is in the recording. Since reference pages are public docs and identical for every visitor, that reveals nothing about the reader, so I would leave it as is unless you want the extra sentence.
- Separately and pre-existing, so explicitly not part of this PR: the "Cookies and tracking" section still describes storage as only a theme preference, while posthog-js already set a first-party `ph_*` cookie before this change. Worth a follow-up issue against main rather than a change here.

Merge order is unchanged: this stays behind #4101, and after that lands it needs an update-branch plus `node docs/site/scripts/lab-manifest.mjs` and a commit of the regenerated `public/lab/replay-only-where-readers-bounce.html`, which I measured earlier. Happy to re-read the head after that rebase.

yonatangross:
**reviewer-y1g verdict on head `3ea45c262bf7eb9188d31a812005ae7551221888`: LAND**

- Confirmed over GraphQL that the prior LAND on this PR names head `9ec2e8a22b4ff527b90a565b212792ec9f8449cc`, superseding an earlier HOLD on `db8d7581`. So the baseline includes the privacy page paragraph, and that paragraph is still present at HEAD (line 27 of `docs/site/app/(home)/privacy/page.tsx`).
- Diffed the PR source against merged main (`1617f79..3ea45c26`) and against the baseline the `9ec2e8a2` LAND covered (`e00925e..9ec2e8a2`), both with the generated Lab paths excluded. Same 7 files, same 576 normalized lines, byte identical after stripping hunk offsets and index lines. Raw churn is 4 lines, every one an `@@` or `index` line, zero content lines.
- The merge `5a8852b` is a clean union in both directions. Its diff against the branch parent `9ec2e8a2` is byte identical to main's own changes (`e00925e..1617f79`, 12615 normalized lines), and its diff against the main parent `1617f79` is byte identical to the branch's own changes (765 normalized lines). Neither check excluded any path, so nothing from either side was dropped or rewritten.
- Only one source file was touched by both sides, `docs/site/next.config.mjs`, and both sides survive at HEAD in different config sections: main's three `/lab` redirect rules in `redirects`, and the branch's `worker-src 'self' blob:` directive in the Content Security Policy header. The other overlap is the generated `docs/site/lib/generated/lab-data.ts`.
- Lab regen `3ea45c26` touches exactly one file, `docs/site/public/lab/replay-only-where-readers-bounce.html`. Zero paths outside the allowed Lab directories.
- Not verified: the TypeScript type check. `docs/site/node_modules` does not exist in this clone and installing was out of scope.

PR #4103, opened 2026-09-14
feat(site): static homepage, hero install command, host_selected, /mcp passthrough

yonatangross:
## Problem

1. The homepage rendered per request. `app/(home)/page.tsx` awaited `searchParams`, so production `/` answered `cache-control: private, no-cache, no-store, max-age=0, must-revalidate` with `x-vercel-cache: MISS`; cold TTFB was 3.05s while prerendered docs pages sit around 0.25s.
2. The install snippet sat below six host cards: top at 954px on 375x812 and 827px on 1024x768. "Get started" linked to docs, not the command.
3. `POST https://orchestkit.yonyon.ai/mcp` returned 404. next.config rewrites `/mcp` to `/api/mcp`, but middleware runs before rewrites and `/mcp` was missing from `SERVED_EXACT`, so middleware answered with a JSON 404 and fired `agent:not-found`.
4. Nothing recorded which host card a visitor picked.

## Change

- `components/search-params-sync.tsx`: a null client component inside `Suspense` that reads `useSearchParams`. `HostInstallPicker` and `LibraryCatalog` use it for `?host=` and `?lib=`, so they still prerender with defaults and adopt deep links after hydration. `page.tsx` no longer takes `searchParams`.
- Hero: one copyable install command (`SITE.installCommand`, value untouched) right under the summary, Get started beside it, host picker below. `CopyInstallButton` gains a visible focus ring, a status live region outside the labelled button, `motion-reduce` handling, wrapping that does not split words, and `host` plus `surface` props on `install_copied`. The picker ring animation respects reduced motion.
- `host_selected` event with `{ host }` on every host card activation, through the existing `track` helper.
- `lib/agent-404.ts`: `/mcp` added to `SERVED_EXACT`. It was the only exact rewrite source missing; the other MCP paths were already listed or covered by the `/api` prefix.

## Evidence and verification

- `next build`: the `/` row carries the Static marker with Revalidate 1h and Expire 1y. Main rendered it per request.
- `next start` on this branch, `GET /` twice: `x-nextjs-cache: HIT`, `x-nextjs-prerender: 1`, `Cache-Control: s-maxage=3600, stale-while-revalidate=31532400`, TTFB 20ms then 5ms (local, not a production number).
- `POST /mcp` with a JSON-RPC `tools/list` body: 200 `application/json` with the tool list. `POST /mcp-nope` is still 404.
[...]

PR #4106, opened 2026-09-14
docs(site): split configure reference and make API/MCP pages findable by name (ora.ai)

yonatangross:
Two ora.ai score fixes for the docs site, one PR.

## Fix A: page-token-budget (configure reference split)

**ora check:** `page-token-budget` (25K tokens per page). `docs/site/content/docs/reference/skills/configure.mdx` was 130,815 bytes, about 32.7K tokens at 4 chars per token.

The reference pages are generated (`scripts/build-docs.sh` runs `scripts/_build-docs-generate.py`, and the docs workflow fails on drift), so a hand split would be overwritten on the next build. The split lives in the generator instead: a skill listed in `SPLIT_SKILLS` whose assembled page would exceed `SPLIT_OVER_BYTES` (90,000) is written as a folder. `index.mdx` keeps the SKILL.md body at the original URL and ends with a "Companion pages" list, one line per part. Each subdirectory (rules, references, ...) becomes one page; a subdirectory that is itself over budget becomes one page per file. Every part reuses the `### <title>` heading the flat page carried, so the old anchor ids resolve on the part page. Output for the other 106 skills is byte-identical (checked with `git status` after regeneration).

Opt-in per skill for now (`SPLIT_SKILLS = {"configure"}`) so this PR changes one page. 14 other generated skill pages are also over 90 KB (doctor 224 KB, product-frameworks 141 KB, security-patterns 136 KB, rag-retrieval 125 KB, api-design 122 KB, implement 122 KB, langgraph, emulate-seed, verify, performance, brainstorm, architecture-decision-record, chain-patterns, monitoring-observability). Adding a slug to the set splits it on the next build.

Page sizes, before and after (bytes, estimated tokens at 4 chars per token):

| Page | Bytes | Est. tokens |
|------|-------|-------------|
| `reference/skills/configure.mdx` (before) | 130,815 | 32.7K |
| `reference/skills/configure/index.mdx` | 16,593 | 4.1K |
| `reference/skills/configure/rules.mdx` | 5,720 | 1.4K |
| `reference/skills/configure/references/cc-version-settings.mdx` | 66,021 | 16.5K |
| `reference/skills/configure/references/mcp-config.mdx` | 23,178 | 5.8K |
| `reference/skills/configure/references/operator-scope-settings.mdx` | 15,902 | 4.0K |
| `reference/skills/configure/references/http-hooks.mdx` | 4,003 | 1.0K |
| `reference/skills/configure/references/presets.mdx` | 2,763 | 0.7K |

[...]

PR #4107, opened 2026-09-14
feat(site): register WebMCP page tools (search_docs, get_page) with server-rendered evidence

yonatangross:
## Why

ora.ai's usability layer "webmcp" check (5 points) reads the docs site as "mentioned only, not implemented". Its recommendation: register tools with `document.modelContext.registerTool()` and use `navigator.modelContext` only as a trailing compatibility fallback; declarative forms with `toolname` and `tooldescription` provide server-rendered evidence but should not be the only tool surface.

## Diagnosis (before this PR)

- `docs/site/components/webmcp-provider.tsx:22-24` typed `ModelContext` with only `provideContext`, and `:117-122` called `navigator.modelContext.provideContext()` exclusively. `document.modelContext.registerTool()` was never called, so a browser or checker that implements the current spec shape saw no registration at all.
- Registration lived entirely inside a client `useEffect` (`:116`). A non-executing crawler gets zero evidence in the HTML.
- The only server-rendered mention was prose in `docs/site/components/agent-readiness-section.tsx:59-65` ("registers WebMCP tools via `navigator.modelContext`"). That is exactly the "mentioned only" signal: text about tools, no declarative form, no `toolname`/`tooldescription` attribute anywhere in the served HTML.
- Not the cause: tools already had names, descriptions and input schemas; the provider was already mounted in the root layout (`app/layout.tsx`), so every route was covered.

## What changed

- `docs/site/lib/webmcp-tools.ts` (new): React-free tool definitions plus `registerWebMcpTools(document, navigator)`. Order: `document.modelContext.registerTool()` per tool, then `navigator.modelContext.registerTool()`, then `navigator.modelContext.provideContext()`, first surface that exists wins. Feature-detected, no-op without the API.
- Tools registered on every page on load (root layout):
  - `search_docs` (query, optional tag) -> `GET /api/search?query=...`
  - `get_page` (docs path) -> `GET /api/md/<slug>` with `Accept: text/markdown`, path normalized and traversal/absolute URLs refused
  - `list_skills` and `get_skill` kept as they were
- `docs/site/components/webmcp-provider.tsx`: now a thin client shim calling the helper.
- `docs/site/components/webmcp-search-form.tsx` (new) rendered on the homepage next to the search trigger: `<form toolname="search_docs" tooldescription="..." action="/api/search" method="get" hidden>` with a `query` input. Server-rendered, no JS, visually hidden, present in HTML as crawler evidence and a working fallback.
- `docs/site/webmcp.d.ts` (new): augments React's `FormHTMLAttributes` with `toolname` / `tooldescription` so the form type-checks without casts.
- `docs/site/components/agent-readiness-section.tsx`: prose updated to match the implementation (document first, navigator fallback, the four tools, the declarative form).
- `docs/site/__tests__/webmcp-tools.test.tsx` (new): mocks `document.modelContext`, asserts `registerTool` was called with `search_docs` and `get_page` (each with description and schema), asserts the navigator fallbacks and the no-API no-op, asserts each execute hits the real route, and renders the form to assert the `toolname` attribute.

No new dependencies.

## Verification (run locally in a fresh clone, 2026-09-14)

- `npx vitest run -c vitest.site.config.ts` (docs/site): `Test Files 56 passed (56)`, `Tests 616 passed (616)`; the new file alone: 9 passed.
- TypeScript check (tsc, noEmit, against docs/site/tsconfig.json): clean, exit 0.
- `npm run build` in docs/site (next build, webpack mode): exit 0 in 2m49s.
- `next start -p 4917` then `curl http://127.0.0.1:4917/`: HTTP 200, 145135 bytes, and the served HTML contains exactly one
  `<form toolname="search_docs" tooldescription="Full-text search across ALL OrchestKit content ..." hidden="" aria-hidden="true" data-webmcp="declarative" action="/api/search" method="get">`.
- Tool routes on the same server: `GET /api/search?query=install&limit=2` HTTP 200 (JSON results), `GET /api/md/foundations/overview` with `Accept: text/markdown` HTTP 200 (frontmatter + body).
- Lint: neither docs/site nor the repo root defines a lint script, and `docs.yml` runs only `npm run build` and `npm test` for the site, so there was no lint gate to run.

Not verified: an actual Chrome build with the WebMCP origin trial calling `document.modelContext.registerTool()` end to end; the registration path is covered by the mocked unit test only. ora.ai's re-score after deploy is also pending.

## Playground

https://orchestkit.yonyon.ai/lab/webmcp-page-tools.html (source: `docs/feat--webmcp-page-tools/webmcp-page-tools.html`, registered in `docs/site/lab-manifest/webmcp-page-tools.json`). Mocked `modelContext` with a browser-shape switch, a `search_docs` call box that performs the real GET when served from the docs origin, the provider before/after, and the curl-captured hidden form.

## Merge note

`app/(home)/page.tsx` gets two lines (one import, one `<WebMcpSearchForm />` next to `<HomeSearchTrigger />`). #4103 also touches that file today, so expect a small conflict if it lands first; the resolution is to keep both.


yonatangross:
**reviewer-y1g verdict on head `7ad3158bf3ac464275e8547d9fa90b59a7ec5fd2`: HOLD**

**BLOCKER (a), the `..` guard is bypassable.** `markdownRouteFor` (`docs/site/lib/webmcp-tools.ts:56`) denies a literal `..`, but percent-encoded dot segments pass the check and the browser's URL parser normalizes them afterwards. Measured with node: `get_page({path: "%2e%2e/%2e%2e/api/jobs/t"})` builds `/api/md/%2e%2e/%2e%2e/api/jobs/t`, and `new URL()` resolves that to `/api/jobs/t`. `%2E%2E` behaves the same. Hardening `app/api/md/[[...slug]]` cannot fix this, since the request never reaches that route.

**BLOCKER (a), `get_skill` has no guard at all.** `webmcp-tools.ts:148` interpolates the argument straight into `` `/docs/reference/skills/${name}` `` with only `.trim().toLowerCase()` applied, so `name: "../../../admin"` resolves to `/admin` with a plain literal `..`. Together these turn two "fetch one docs page" tools into "fetch any same-origin URL", and `fetch` defaults to `credentials: "same-origin"`, so any authenticated route added to this origin later is exposed by default.

**Exact fix:** allowlist the cleaned slug instead of denying `..`. In `markdownRouteFor`, before returning, `if (!clean.split("/").every((s) => /^[a-zA-Z0-9._-]+$/.test(s))) return null;` and apply the same segment test to `name` in `get_skill`. I ran the current guard against the proposed one: all six accepted shapes (`/docs/a/b`, `docs/a/b`, `a/b.md`, `/`, `""`, `/docs`) are byte-identical, and all three encoded escapes return `null`. Worth two test cases next to the existing `refuses absolute URLs and traversal` block, which today only covers the literal `../secret`.

**(b) PASS, no cross-origin or third-party reach.** All four `execute()` bodies fetch path-absolute relative URLs, and a path that normalizes toward `//host` keeps the base origin (checked `..//evil.com/x`, which resolves to `/docs/reference//evil.com/x`). The escalation above is same-origin only.

**MEDIUM (c), registration is once-per-page but unguarded.** The mount is correct: `app/layout.tsx:83` with `useEffect(..., [])` in the root layout, which survives client-side navigation, so it does not run per render and there is no leak. But there is no `try`/`catch` in either `webmcp-provider.tsx` or `registerWebMcpTools`, and `docs/site/app` has no `error.tsx` or `global-error.tsx`, so a `registerTool` that throws takes the whole page to the error fallback. The homepage makes that concrete: it declares `search_docs` twice, once through the hidden declarative form and once through `registerTool`, and this is an origin-trial API whose duplicate-name behavior is not something I would bet the docs site on. Wrap the call and log the failure.

**(d) PASS.** `hidden` plus `aria-hidden="true"` plus `tabIndex={-1}` on both the input and the button; `hidden` alone already removes the form from the accessibility tree and the tab order, and no display utility class overrides it.

**(e) PASS.** No new dependencies. The provider is net -87 lines with the logic moved into `lib/webmcp-tools.ts`, so client bundle impact is roughly flat.

**(f) PASS, the order is genuinely asserted.** `__tests__/webmcp-tools.test.tsx:22` passes both surfaces and asserts `surface === "document.registerTool"` with `navRegister` not called, which is a real precedence test rather than a presence test. 9 tests, matched by the `__tests__/**/*.test.{ts,tsx}` include in `docs/site/vitest.site.config.ts` and run by `.github/workflows/docs.yml`.

**Checks:** all 18 required contexts green on this head, read per name from `commits/7ad3158b/check-runs`. Nothing required is red or pending.

**Conflict with #4103:** both PRs edit `docs/site/app/(home)/page.tsx` in the same two hunks (the import block, and the lines around `<HostInstallPicker>` / `<HomeSearchTrigger>`, where #4103 rewrites `<HostInstallPicker active={host} libraryTab={libraryTab} />` directly above your insertion point), and both append to the generated `lib/generated/lab-data.ts`. **#4107 should rebase onto #4103**, not the other way round: this PR's homepage footprint is 2 lines against #4103's hero rewrite and `searchParams` removal, and the `<HomeSearchTrigger />` anchor survives in #4103, so the rebase is mechanical.

yonatangross:
Addressed the HOLD at 7ad3158bf in 06496f528: per-segment allowlist after one percent-decode for both get_page (markdownRouteFor) and get_skill (new skillRouteFor), plain and encoded traversal return null before any fetch; every registerTool and provideContext call is wrapped and logs on failure; a tool the page already declares through a form toolname is skipped, so search_docs is declared once on the homepage. 19 tests in the file (was 9), 626/626 site tests, tsc clean. Rebase onto #4103 still pending: #4103 is open, so this head is on current main.

yonatangross:
**reviewer-y1g verdict on head `06496f52860dd91ceb70ca6f1e847b7a039407c7`: LAND**

All three findings from my HOLD on `7ad3158b` are fixed, and I re-ran the original escape probes against the new code rather than reading the diff.

**BLOCKER 1 cleared, encoded traversal.** `safeSlugSegments` decodes once, *then* splits, *then* allowlists each segment, which is the correct order: decoding first turns `%2f` into a real separator so it cannot smuggle a segment past the split, and `%` is absent from `SAFE_SEGMENT` so a double encoding (`%252e%252e`) fails on the residual `%`. I ran 13 escape probes through the real `markdownRouteFor`, including `%2e%2e/%2e%2e/api/jobs/t`, `%2E%2E`, `..%2f..%2fapi/jobs/t`, `%252e%252e/...`, `....//....//x`, `a/%00b` and a bare `%`. **All 13 return null, zero leaks**, against 3 that escaped to `/api/jobs/t`, `/admin` and `/secret` on the previous head.

**BLOCKER 2 cleared, `get_skill`.** `skillRouteFor` requires exactly one allowlisted segment, so `../../../admin`, `%2e%2e/%2e%2e/admin`, `..%2fadmin`, `a/b` and `""` all return null, while `assess` and `ASSESS` both map to `/docs/reference/skills/assess`.

**No false negatives from the allowlist.** This was my main worry about switching to an allowlist, so I checked the code comment's claim independently instead of trusting it: every one of the **277 docs pages** under `content/docs` maps through `markdownRouteFor` and every one of the **108 skill directories** under `plugins/ork/skills` maps through `skillRouteFor`. Zero rejections on either side.

**MEDIUM 3 cleared, unguarded registration.** `guarded()` wraps every `registerTool` call and the `provideContext` call, with a `console.warn` that names the tool. `declaredToolNames()` reads `form[toolname]` off the document and filters those names out of `pending`, so the homepage declares `search_docs` exactly once. The tests assert both halves properly: a throwing `registerTool` is caught *and* the remaining tools still register, and there is a negative case confirming `search_docs` is registered when no declarative form is present.

**Tests.** 19 in the file, and the two that matter most assert that **no request leaves** on a refused path (`expect(fetchMock).not.toHaveBeenCalled()`), which is a stronger claim than asserting the return value. The document-first precedence test I checked last time is unchanged.

**Re-verified prior passes.** `webmcp-search-form.tsx`, `webmcp-provider.tsx` and `app/(home)/page.tsx` are byte-identical to the previous head (empty diff), so the accessibility posture and the 2-line homepage footprint are unchanged, and no dependency was added.

**Checks:** all 18 required contexts green on this head, read per name from `commits/06496f52/check-runs`, and no check of any kind is non-success.

**Nits, none blocking.** (1) `guarded` catches synchronous throws only; if a browser's `registerTool` returns a rejected promise it becomes an unhandled rejection rather than a warn. (2) `markdownRouteFor` rejects uppercase while `skillRouteFor` lowercases first, so an agent that capitalizes a path gets a refusal; deliberate and tested, but lowercasing for parity would cost nothing. (3) Because the imperative `search_docs` is skipped on the homepage, that one page exposes only the declarative form, which has no `tag` input, so the landing page offers a slightly weaker `search_docs` than every other page; adding a `tag` field to the form would close the gap.

**On #4103:** still open, and this does not change the verdict. #4107 is `MERGEABLE` against main today (`mergeStateStatus` is `BEHIND`, which is staleness, not conflict). The overlap is a merge-order matter between two open PRs rather than a defect here, and my recommendation is unchanged: #4107 takes the rebase, since its homepage footprint is 2 lines against #4103's hero rewrite and the `<HomeSearchTrigger />` anchor survives there.

yonatangross:
**reviewer-y1g verdict on head `f5f18515ee85900bafb7c5c89f1def864cd184eb`: LAND**

Verified entirely over git in a local clone, no API reads. My LAND at `06496f52` carries because this head changes nothing the review depended on.

**Ancestry is exactly an update from main.** `06496f52860dd91ceb70ca6f1e847b7a039407c7` is an ancestor of this head (`git merge-base --is-ancestor`), and the single new commit is `f5f18515`, a merge whose two parents are that reviewed head and `615f5e461bcfabe3f374613b72d7324f74959b5a`, which is current `origin/main`. `origin/main` is now fully merged in, so there is no third commit and no rewritten history.

**The PR diff file set is unchanged.** Comparing `merge-base...head` at both heads, no file entered or left the diff: the same 11 files, with `#4106` (`496649fb`) now sitting in the base rather than alongside.

**Only one file's diff moved, and only positionally.** Ignoring hunk offsets and blob hashes, all 11 files have byte-identical added and removed lines. The lone difference is the hunk header on the generated `docs/site/lib/generated/lab-data.ts`, `@@ -169,6 +169,7 @@` becoming `@@ -171,6 +171,7 @@`, because `#4106` landed two entries above this one. The added line itself is character for character the same single `webmcp-page-tools` entry.

**The merge resolved correctly rather than clobbering.** `lab-data.ts` is the one file that differs from both parents, which is the right shape for a file each side appended to: against `origin/main` this head adds exactly one line (the webmcp entry), so the merged file is main's content plus this PR's entry, with `#4106`'s entry preserved.

**The reviewed source is the identical git blob.** `lib/webmcp-tools.ts`, `__tests__/webmcp-tools.test.tsx`, `components/webmcp-provider.tsx`, `components/webmcp-search-form.tsx`, `webmcp.d.ts` and `app/(home)/page.tsx` all resolve to the same object hash at both heads, so the 13 escape probes, the 277 docs page and 108 skill directory allowlist sweep, and the guarded-registration findings from my previous comment hold verbatim without re-running.

**Generated artifacts are consistent and reproducible.** `node docs/site/scripts/lab-manifest.mjs --check` passes at this head (`set check OK: 153 fragment(s) == lab-data.ts == public/lab/`), and a full regenerate rewrites 153 entries and leaves **0 dirty files**, so the committed generated output matches what the generator produces from this tree. The three `source absent (gitignored or pruned)` lines are pre-existing entries whose sources are gitignored, and the script keeps the committed copies by design.

**Scope note:** CI at this head is the team lead's read, not mine. I did not poll checks for this comment, so this LAND is on the diff and the tree, conditional on the 18 required contexts being green at `f5f18515`.

yonatangross:
**reviewer-y1g verdict on head `2b07b9b51b8255fac88f44c98191f77b12f3a150`: LAND**

Re-derived from scratch over git after the force-push, no API reads. Everything the review rests on is unchanged.

**Ancestry matches the described rewrite.** `06496f52860dd91ceb70ca6f1e847b7a039407c7`, the head I reviewed the code at, is still an ancestor. Exactly two commits sit on top of it: `79e5401f`, a merge whose parents are that reviewed head and `4e0462baa39993c969ea7ed791b20abfa029f564` (current `origin/main`, carrying #4101), and `2b07b9b5`, a single-parent regenerate commit touching **1 file with 2 insertions and 2 deletions**. The previous head `f5f18515` is confirmed no longer reachable from the branch, as stated.

**The PR diff file set is unchanged.** Comparing `merge-base...head` at the reviewed head and at this one, the same 11 files, none entered or left.

**Ten of the eleven files are content-identical; one changed.** Ignoring hunk offsets and blob hashes, only `docs/site/public/lab/webmcp-page-tools.html` differs. Every other file, including the generated `lib/generated/lab-data.ts`, carries byte-identical added and removed lines.

**The one change is chrome only, and I proved it rather than inspecting it.** Stripping the two injected `<!-- ork-lab-head -->` and `<!-- ork-lab-chrome -->` blocks out of the new file yields a result **byte-identical** to the copy I reviewed. The authored source, `docs/feat--webmcp-page-tools/webmcp-page-tools.html`, is the same git blob at both heads, so only the served copy moved and the playground's own markup, styles and demo script are untouched. The page structure is intact (one each of `</head>`, `<body>`, `</body>`, `</html>`).

**The reviewed source is the identical git blob.** `lib/webmcp-tools.ts`, `__tests__/webmcp-tools.test.tsx`, `components/webmcp-provider.tsx`, `components/webmcp-search-form.tsx`, `webmcp.d.ts`, `app/(home)/page.tsx`, `components/agent-readiness-section.tsx` and `lab-manifest/webmcp-page-tools.json` all resolve to the same object hash at both heads. The 13 escape probes, the 277 docs page and 108 skill directory allowlist sweep, and the guarded-registration findings hold verbatim.

**Generated artifacts are consistent and reproducible.** `node docs/site/scripts/lab-manifest.mjs --check` reports `set check OK: 154 fragment(s) == lab-data.ts == public/lab/`, and a full regenerate writes 154 entries and leaves **0 dirty files**. The three `source absent (gitignored or pruned)` lines are pre-existing and the script keeps the committed copies by design.

**One thing worth naming, not a blocker and not yours.** The injected chrome brings #4101's page-view beacon onto this Lab page: it posts `{project_id, path, referrer, timestamp}` to same-origin `/api/analytics` via `sendBeacon`, guarded by an iframe check, skipped on localhost, wrapped in try/catch. No third party, no identifiers beyond the path and referrer. That is inherited from #4101 which is already on main, so it is in scope for that PR rather than this one, but it is new on this page and worth stating out loud.

**Scope note:** CI at this head is the team lead's read. I did not poll checks, so this LAND is on the diff and the tree, conditional on the 18 required contexts being green at `2b07b9b5`.

Issue #4085, opened 2026-09-13
pre-push flakes under load: hardcoded MAX_JOBS=8 and a 5 s hook timeout fail green files


yonatangross:
## Symptom

A hooked `git push` from a clean, green worktree fails at the pre-push unit or security stage under machine load, while the same files pass standalone. Measured 2026-09-13 from lane fix-4070 (head a4f65b7d3, origin/main 58477adfb), 5-minute load 37 at the time:

| file | in pre-push | standalone |
|---|---|---|
| tests/unit/test-image-paste-guard.sh | 1 of 68 failed, 213 s | 22/22 in 11 s |
| security: sqlite-injection | 10/16 (head), 15/16 (origin/main probe) | 16/16 |
| security: permission-mode-matrix | 16/17 (head), 13/17 (origin/main probe) | 17/17 |

The origin/main probe ran in a detached worktree with no GIT_* variables, so the failures are on main, not in the fix-4070 diff.

## Cause (read from the runner, not yet measured)

- `pre-push:392` hardcodes `MAX_JOBS=8`; `:393-400` only lowers it to `hw.ncpu`; `:436` feeds that to `xargs -P`. No env override, no load check.
- `tests/unit/test-image-paste-guard.sh:53` pipes stdin into `timeout "${3:-5}" node run-hook.mjs`; the 5 s budget is a positional arg with no env fallback. Under 8-way parallel hook runs at load 37 the hook does not answer in 5 s.
- The security suites show the same shape: expected `deny`, got `abstain` (permission-mode-matrix), and a UNION payload row not stored (sqlite-injection). Both look like a hook timing out and falling back, not a logic change.

## Ask

1. `ORK_PRE_PUSH_JOBS` env read at `pre-push:392`, and a load-aware default (for example `min(hw.ncpu, 8, max(1, ncpu - loadavg1))`).
2. A per-hook timeout env honoured by `test-image-paste-guard.sh` and any other suite with a positional timeout.
3. Confirm the abstain-under-load mechanism in the two security suites and make the assertion read the real cause when a hook times out, instead of reporting a wrong verdict.

Pushing with `--no-verify` is not an option on this floor. The fix-4070 branch waits on this landing on main.

PR #4111, opened 2026-09-14
fix(tests): report hook timeouts in test-image-paste-guard instead of swallowing them (GH-4087)

### LINKED ISSUE #4087: test-image-paste-guard.sh swallows a hook timeout (exit 124) and reports a wrong verdict
Follow-up to #4085 / #4086. QA (2026-09-13) found that `tests/unit/test-image-paste-guard.sh` still runs `timeout N node ... 2>/dev/null || true` in `run_hook_with_input`, so a hook that exceeds `ORK_HOOK_TIMEOUT` reads as "normal text prompt processes without error ... FAIL" with no line naming the timeout (3 cases under a 2 s hook with ORK_HOOK_TIMEOUT=1). The file's own comment admits a timeout reads as a wrong verdict there.

Ask: route `run_hook_with_input` through `ork_run_with_budget` (or log rc 124 the way `fire_skill` does) so a timeout is reported as "hook timed out after Ns", matching what #4086 did for `hook_decision` and the security suites. No assertion should be weakened.

yonatangross:
## Summary

`tests/unit/test-image-paste-guard.sh` ran `timeout N node ... 2>/dev/null || true` inside `run_hook_with_input`, so when the per-hook budget (`ORK_HOOK_TIMEOUT`) killed node with exit 124, the exit code was discarded and the empty output was fed to the assertions as if it were a hook verdict. QA measured 3 cases reading as "normal text prompt processes without error ... FAIL" under a 2 s hook with `ORK_HOOK_TIMEOUT=1`, with no line naming the timeout.

Closes #4087

## What changed

- `run_hook_with_input` captures the pipeline exit code and, on 124, answers the literal `ERROR` token on stdout plus a stderr cause line: `hook '<name>' timed out after Ns (ORK_HOOK_TIMEOUT); no verdict was produced`. This is the shape #4086 gave `hook_decision`: ERROR fails every assertion shape in this file (it is neither JSON nor silent success), so a timeout can no longer read as PASS or as a causeless FAIL, and no assertion is weakened.
- The verdict is echoed rather than `log_fail`'ed because callers capture the helper's stdout inside `$( )`: a counter increment there would run in a subshell and never reach the summary.
- The 100 ms stdin-watchdog shape from #4085 needs no handling here: this file feeds the hook through a builtin echo, whose bytes are in the pipe before node spawns, so the watchdog has no producer-side boot to race.
- New "Hook budget" suite section fakes a hanging hook (a never-answering runner under a 1 s budget) and asserts both the ERROR verdict and the named cause, plus that a real budget still delivers normal verdicts. It SKIPs on boxes without a real timeout binary, the same trade `test-hook-decision-timeout.sh` makes (#4086); verified with a GNU-shaped shim.

## Tests

- Suite standalone: 24/24 on macOS without a timeout binary (the new case SKIPs there), 25/25 with a GNU-shaped `timeout` shim proving the 124 path end to end (ERROR verdict, "timed out after 1s (ORK_HOOK_TIMEOUT)" cause, normal verdict still delivered).
- Full pre-push validation on push: typecheck OK, build OK, unit 69 tests OK, security OK, mdx-compile, agent frontmatter, skill structure, manifests, token budget all OK.

## Interactive Playground

Live: https://orchestkit.yonyon.ai/lab/hook-timeout-verdict.html

Interactive page: `docs/fix--74279-dev-null-timeout/hook-timeout-verdict.html` (drag the hook latency and the ORK_HOOK_TIMEOUT budget to see the before and after verdict lines), registered as `docs/site/lab-manifest/hook-timeout-verdict.json` with its generated `public/lab` copy and `lab-data.ts` row. Lab set oracle green at 153 fragments.

## Follow-up, out of scope here

`./bin/bump-version.sh patch`, the pre-push version gate's own remediation command, crashes on the repo's current prerelease format: `bump_version` feeds `0-beta.21` into arithmetic and dies, and because main() assigns through `local new_version=$(...)` the failure is masked, so the script rewrites every version file to empty strings (measured on this branch before resetting). It needs its own issue and PR; no version bump is committed here because conventional-prefix branches are exactly the ones the gate stands down for.

## Summary by CodeRabbit

- **Documentation**
  - Added a lab manifest for the hook-timeout verdict documentation page, including its title, description, tags, and publication date.

- **Tests**
  - Added coverage for hook timeout handling, including clear error verdicts and timeout causes.
  - Added checks confirming normal verdicts continue to work after a timeout.
  - Improved oversized-prompt latency test reporting to identify which run timed out.


yonatangross:
HOLD @cd0e4284

Finding (tests/unit/test-image-paste-guard.sh:392 and :397): the latency case calls run_hook_with_input with stdout sent to /dev/null, so the new ERROR verdict never reaches an assertion there. If both runs hit the budget, each takes about the same time, delta floors near 0, and the case logs PASS ("oversized prompt costs +0ms"). The stderr cause line is printed but the counter still counts a timeout as a pass, so the PR's "ERROR fails every assertion shape in this file" does not hold for this case. Measured with the helper copied verbatim from this head, a never-answering runner and a GNU-shaped timeout shim at a 1s budget: two "timed out after 1s (ORK_HOOK_TIMEOUT)" lines, then delta=0ms, which this case reads as PASS.
Suggested fix: capture both outputs and log_fail when either equals ERROR before computing the delta.

Verified and fine otherwise: rc capture via `|| rc=$?` (no new `|| true`, the `2>/dev/null` on hook stderr is pre-existing), rc 124 wins under pipefail, every other assertion shape fails on ERROR, the SKIP carries its reason. Standalone run at this head: 24/24 without a timeout binary, 25/25 with the shim (ERROR verdict, named cause, normal verdict still delivered). Scope, playground page, Lab copy (only the injected head and chrome blocks differ) and lab-data row are in order; no workflow or plugin.json changes; no em or en dashes.

yonatangross:
LAND @ac290a67 (reviewer read, 12:19Z)

The HOLD from cd0e4284 (comment 5663442614) is fixed at root, verified by running a timeout shim against both heads.

**Findings**
- tests/unit/test-image-paste-guard.sh:393 and :398 now capture `small_out` / `big_out` instead of sending helper stdout to /dev/null. :408 to :415 FAIL the case on any ERROR verdict, naming the trivial run, the oversized run, or both, and return before the delta is judged. The delta assertion (:417 onward) is unchanged for the normal path.
- :78 to :86 `run_hook_with_input` replaces the old `|| true` with `|| rc=$?` and answers ERROR plus a stderr cause on rc 124. The node stderr redirect on :79 is carried over from main. The only new stderr redirects are `type -t` (:581) and a `jq -e` exit check (:623); neither swallows a verdict.
- Scope: only the test file plus the four Lab files (fix branch playground page, lab-manifest fragment, lab-data.ts row, public/lab copy). The cd0e4284 to ac290a67 delta touches only the latency case (18 added, 2 removed lines).
- Nit, non-blocking: the :413 detail hardcodes "5s". Both calls pass 5, so it is accurate today.

**Commands and results**
- `bash tests/unit/test-image-paste-guard.sh` (macOS, no timeout binary): `Total: 24 | Passed: 24 | Failed: 0`
- GNU-shaped timeout shim on PATH that forces rc 124 by payload (trivial run, 60K run, or both); otherwise it enforces a real budget:

| mode | ac290a67 latency case | cd0e4284 latency case |
|:-|:-|:-|
| none | PASS +45ms, Total 25/25 | PASS |
| small | FAIL "in the trivial (small) run" | PASS +28ms (HOLD bug) |
| big | FAIL "in the oversized (big) run" | PASS +0ms (HOLD bug) |
| both | FAIL "in the trivial (small) run and oversized (big) run" | PASS +10ms (HOLD bug) |

  The hanging hook case at head passes under the shim: ERROR verdict, and the cause "timed out after 1s (ORK_HOOK_TIMEOUT)" is named.
- `node docs/site/scripts/lab-manifest.mjs` in check mode: `set check OK: 159 fragment(s) == lab-data.ts == public/lab/`
- PR body has an Interactive Playground section naming hook-timeout-verdict.html and its manifest fragment.
- Authored text scan (diff added lines, commit bodies): zero em or en dashes. Double hyphens appear only in CSS custom properties, HTML comment markers and the branch path.

yonatangross:
LAND @7a3cf7f6 (conductor delta read, 13:0xZ)

Carries the Opus reviewer's LAND @ac290a67 (comment 5663797275, shim proof that the latency case fails on any ERROR verdict). The head moved only by a merge of main 8dc4863f13e57d48d0ada86516c163d8683858d2, verified over git at zero REST:

- main 8dc4863f is an ancestor of 7a3cf7f645ea640bf0e2a29d49318e2540f040d7.
- Blob identical between ac290a67 and 7a3cf7f6: tests/unit/test-image-paste-guard.sh, the hook-timeout-verdict Lab page, docs/site/lab-manifest/hook-timeout-verdict.json, docs/site/public/lab/hook-timeout-verdict.html.
- `git diff 8dc4863f 7a3cf7f6` touches the same 5 files as before; docs/site/lib/generated/lab-data.ts differs from main by exactly one added line, this PR's hook-timeout-verdict row.

Production code untouched, reviewed test unchanged. Merge still requires the 18 required checks at 7a3cf7f6 by name before squash with --match-head-commit.

PR #4126, opened 2026-09-14
fix(build): generate docs data after stamping (GH-4116)

### LINKED ISSUE #4116: build: docs search index is generated before mdx regeneration and count stamping
## Problem

`npm run build` is not a fixed point for the docs search index. `scripts/build-plugins.sh` (main b75194e0) runs the steps in this order:

| line | step |
|---|---|
| 673-674 | `gen-docs-search-index.js` (reads the reference mdx) |
| 747-748 | `build-docs.sh` (regenerates the reference mdx from src/skills) |
| 814 | `stamp-counts.sh` (stamps "N global hook entries" into the generated mdx) |

The search index is written from the mdx as they stood before this build regenerated and stamped them. A PR that changes stamped mdx therefore commits an index that is one build behind, and the next build (the pre-push hook's own build, or CI's Build job) changes it again.

## Evidence

- PR 4114 at 8035322e: after its main merge, `npm run build` and a commit, the pre-push hook's build left `docs/site/lib/generated/docs-search-index.ts` dirty. The "Doctor: Hook Validation" description changed from 149 to 152 global hook entries (hooks.json has 152). The lane needed a second commit and a second full pre-push run.
- The unstamped source reference `src/skills/doctor/references/hook-validation.md` still says 149, which is what the index picks up when it runs before the stamp.
- Related count drift on PR 4108 (stamped mdx reverted to 149 / 215 by a regeneration that skipped the stamp step).

## Fix shape (proposal)

1. Generate the search index after `build-docs.sh` and `stamp-counts.sh`, or rerun `gen-docs-search-index.js` after the stamp.
2. Add a determinism guard: run the docs generation twice in CI and require zero diff on the second run.
3. Optionally stamp the source references too, so no step can read the stale literal.

## Acceptance

- A clean checkout: `npm run build`, commit, `npm run build` again leaves `git status --porcelain` empty.
- Pre-push on a PR that touches stamped mdx needs one commit, not two.

yonatangross:
The docs search index ran before reference generation and count stamping, so a later build could change an already committed index. Generate it from the final MDX instead. Also move docs site data after stamping because it reads the synchronized manifest version.

Add a regression test discovered by the existing unit-test runner. It exercises the real documentation generators twice in their production order inside an isolated Git fixture, commits the first result, and requires the second result to leave no changes. Mutations cover early indexing, indexing before count stamping, early version consumption, and second-run drift.

Validation after merging main at 5f8997ba: two full builds succeed. Both porcelain outputs below contain zero bytes. The focused fixed-point regression passes, ordering mutations fail as expected, and the Lab manifest check passes with 168 fragments. Pre-push validation passed: 71 unit test files, security, TypeScript, build, imports, MDX compilation, agents, skills, and manifests. Full-suite verification is delegated to CI per the conductor instruction.

First build, porcelain output:
```
```

Second build, porcelain output:
```
```

## Interactive Playground

[One build should be enough](https://orchestkit.yonyon.ai/lab/build-fixed-point-4116.html) compares the previous and corrected build orders. The page and its Lab registration are included in this PR; the link becomes available after the site deploys this change.

Closes GH-4116. Closes #4116.

Frozen after the initial push. This lane will make no further pushes and will not merge the PR.

## Summary by CodeRabbit

* **Documentation**
  * Added the `build-fixed-point-4116` lab to the documentation gallery.

* **Build Improvements**
  * Updated production documentation builds to generate gallery data and search indexes from final content.
  * Builds now stop when required data generation fails.

* **Tests**
  * Added coverage confirming documentation builds produce final, consistent output and remain unchanged on repeat runs.


yonatangross:
LAND @e6a84234 (reviewer read, 17:07Z)

Diff is exactly the 7 expected files. No workflow and no plugin.json, so no eyeball path.
Reorder is correct: generate-docs-data.js and gen-docs-search-index.js now run after build-docs.sh (phase 8) and after stamp-counts.sh, inside phase 10 before the echo only summary. Two hunks in build-plugins.sh, nothing else moved or dropped. Phase 7 keeps the lab gallery block and its header was retitled to match.
Nothing between the old and new positions consumes either output: grep over scripts/ and bin/ shows plugins-data.ts and docs-search-index.ts are read only by the Next.js site, never by build-docs, stamp-counts, generate-changelog-data or the marketplace sync.
Failure behavior is unchanged. Both calls keep their trailing echo fallback, so a generator failure warns and does not fail the build, same as before the move.
Test is auto discovered: scripts/ci/run-tests.sh globs test-*.sh and ci.yml line 595 points it at tests/unit.
Ran it once in a throwaway worktree at this head, no npm install needed: PASS: production docs sequence reaches a fixed point after regenerated MDX and stamped counts (exit 0, 42s).
It would fail on a revert. The step order is extracted from the script by line number, so the old order runs generate-docs-data before stamp-counts and trips "docs data was generated before the stamped manifest version". A regex that stops matching raises instead of silently skipping.
node docs/site/scripts/lab-manifest.mjs in check mode: set check OK, 168 fragments consistent across lab-data.ts and public/lab/.
Lab page bodies carry no em dash, en dash or spaced double hyphen. The one em dash in the diff is in the test fixture string, mirroring the real overview.mdx description that stamp-counts rewrites, so it is required fixture data.
Two notes, neither blocking: the test drives the four steps standalone rather than build-plugins.sh itself (documented tradeoff, plugin assembly is out of this gate), and neither the test nor the runner sets a timeout, so its bound is the small fixture.

yonatangross:
LAND @2969b99d (conductor delta read, 17:23Z)

Carries the Opus reviewer's LAND @e6a84234 (comment 5667743938). The head moved only by an update-branch merge of main 65e4e60492a3936d8ebef60b1ff44f0452091a40 (the beta.25 release commit), verified after the fact:

- 2969b99d2cda5c8ac0cf4e034786c8f406a00c1f has parents e6a84234 and 65e4e604 (author Yonatan, committer GitHub, 17:11Z).
- git diff e6a84234..2969b99d over the PR's 7 files is empty: scripts/build-plugins.sh, scripts/gen-docs-search-index.js, tests/unit/test-build-fixed-point.sh and the four Lab files are byte-identical to the reviewed head.
- 2969b99d vs main is exactly those 7 files (+409/-15); the release commit touched none of them.
- 18 required checks green by name at 2969b99d, mergeStateStatus CLEAN.

No workflow or plugin.json file in the PR, so this merges under the dev rule.

PR #4132, opened 2026-09-14
fix(tests): sample the paste guard latency delta three times (GH-4130)

### LINKED ISSUE #4130: test-image-paste-guard.sh latency delta rejects pushes under host load (3 branches, 2026-09-14)
## Symptom

tests/unit/test-image-paste-guard.sh, test 21 (oversized prompt latency), times two single runs of the unified dispatcher and bounds their difference at +120ms. One sample per side. Under host load the two samples land in different regimes, the delta blows past the budget, and the pre-push hook rejects the push; the same file passes on an immediate retry.

## Evidence (2026-09-14, three branches on one host, none touching the hook under test)

- fix/4116-build-fixed-point (#4126): first push rejected on this file only; direct rerun 24/24; second push passed all 71 unit files.
- feat/auth-chain-optional-identity (#4129): first push rejected on this file only; retry passed the unit stage.
- docs/contributing-strict-main (CONTRIBUTING.md only): push rejected; focused rerun 23/24 with the measured delta +175ms against the +120ms budget.

Host load1 was between 26 and 40 during these runs (other suites in parallel). The size-delta design (#4024) removed the startup term, but a single pair of samples still carries scheduler noise of the same order as the budget.

## Ask

Take the minimum delta over three pairs (or the median), so a single descheduled sample cannot fail the gate, and keep the +120ms bound. Alternatively move the timing check out of the pre-push unit stage into a CI-only performance test. Same family as #4089.

yonatangross:
A single small/big timing pair let scheduler noise reject unrelated pushes under host load. GH-4130 recorded a +175ms delta against the +120ms budget that passed on retry.

Test 21 now runs three small then big rounds and compares their minimum delta against the same strict +120ms budget. Startup subtraction, the negative-delta floor, and immediate ERROR failures remain; a latency failure prints all three deltas.

Validation: `bash tests/unit/test-image-paste-guard.sh`

`Total: 24  |  Passed: 24  |  Failed: 0`

A fresh traced run observed normalized deltas of 0ms, 5ms, and 27ms, with a 0ms minimum. The first raw delta was -1ms and was floored to zero. These are host-specific observations, not timing guarantees. The existing hanging-hook check skips locally because no timeout binary is installed. Controlled probes covered ERROR on all six calls, sustained regression, and the exact budget boundary: twelve cases passed and four deliberate mutations were caught.

This branch's own push was the first real test of the fix in the pre-push gate, with hooks enabled. It passed all 71 unit files and security checks. The initial head also passed CI unit jobs on Node 20 and Node 22, each with `Passed: 71; Failed: 0`.

## Interactive Playground

[Three-round latency sampling](https://orchestkit.yonyon.ai/lab/paste-guard-latency-sampling-4130.html) compares the old single-pair result with the new three-round minimum. Editable inputs, noisy and sustained-regression presets, and ERROR controls demonstrate the unchanged budget. Observed timings are labeled separately.

Source page: `docs/fix--4130-paste-guard-latency-sampling/index.html`. The Lab manifest fragment and generated gallery data and published copy are included. `node docs/site/scripts/lab-manifest.mjs --check` reports `set check OK: 169 fragment(s) == lab-data.ts == public/lab/`.

Closes GH-4130
Closes #4130

FROZEN after the Lab follow-up push: no further pushes and no merge by this lane.


yonatangross:
LAND @68858db1 (reviewer read, 19:39Z)

Three rounds, small then big alternating inside each round, minimum of the three deltas kept, budget still strictly less than 120ms, and all three deltas printed on failure (tests/unit/test-image-paste-guard.sh:397 to 444).
The GH-4087 handling survives: an ERROR from either run in any round calls log_fail naming which run timed out and returns before any delta is judged, so it is never a skip and never a zero.
It still fails on a real regression. A guard that actually scans the 60K payload raises every round's delta, so the minimum rises with them and crosses the 120ms bound; the minimum only discards noise, which can add cost but never remove payload work.
Checked the min selection loop under set -euo pipefail with a worst case where the last delta is not the minimum. The failing test is inside an AND list, so it does not abort the script.
Test functions: 24 at base 47bd7818, 24 at head, identical names, none deleted or renamed, invocation count 24 both sides.
Three runs of that one file in a detached worktree at this head:

```
  Total: 24  |  Passed: 24  |  Failed: 0
  Total: 24  |  Passed: 24  |  Failed: 0
  Total: 24  |  Passed: 24  |  Failed: 0
```

The oversized prompt line read +0ms over trivial on all three runs. Lab page text matches the code and does not overstate it, naming three rounds, the minimum, the strict 120ms budget, the zero floor on negative deltas, and calling itself a focused estimator rather than a statistical guarantee. No em dash or en dash anywhere in the diff prose.

PR #4150, opened 2026-09-15
feat(site): add Devin host card and rebuild picker as animated grid

yonatangross:
## Summary

- Adds Devin as a seventh host in the homepage picker and getting-started docs, with measured copy: `devin plugins install https://github.com/yonatangross/orchestkit#plugins/ork` lists 76 of 107 skills today; the 31 the parser drops all carry a `triggers:` frontmatter block, and hooks and rules do not load (orchestkit#4146).
- Rebuilds the six-block host picker as a 21st.dev Animated Card Options (isaiahbjork) adaptation on the existing `motion/react` dependency: staggered spring entrance with overshoot settle, shared `layoutId` selection ring, hover lift and tap press. No new dependencies.
- Keyboard access via roving tabindex (one tab stop, arrows move in grid columns, Home/End, Enter/Space activates) and `prefers-reduced-motion` disables the entrance, lift, and ring spring. The `?host=` / `?lib=` URL-driven snippet swap, telemetry, and deep links are unchanged.

## Consumer

- [x] **Every new field/flag/manifest entry in this PR has a documented consumer** (test, hook, skill, or downstream feature). If intentionally emit-only, add the `emit-only` label and link the follow-up issue.

## Test plan

- [x] `npm run build` clean (site build: 363 pages including /docs/getting-started/devin; pre-push hook typecheck, unit, security, manifests all green)
- [ ] `npm run test:skills` (if skills changed)
- [ ] `npm run test:agents` (if agents changed)
- [ ] `npm run test:manifests` (if hooks/skills/agents added or removed)
- [ ] `cd src/hooks && npm test` (if hook code changed - runs the same vitest suite as CI)
- [ ] Smoke tests for any new scripts (see `tests/skills/test-*.sh`)

Site suite: 758 tests pass including new coverage for the Devin card, deep link, and arrow-key navigation.

Playwright landing spec: 5 assertions fail identically against production (strict-mode locator violations in catalog/cookbook/changelog, transparent-CTA token check, 17px mobile overflow). None are regressions from this branch; verified by running the same spec against orchestkit.yonyon.ai. A dedicated axe/a11y suite does not exist in this repo; keyboard and reduced-motion behavior are covered by the new unit tests and manual verification.

## Interactive Playground

https://orchestkit.yonyon.ai/lab/devin-host-card.html

Generated with [Devin](https://devin.ai)

PR #4153, opened 2026-09-15
fix(site): render host cards visibly during SSR

yonatangross:
## Summary

- Fixes the P2 bar from HOLD 5684976852 on #4150 (merged at c79ba4e9 before this landed): useReducedMotion() is null during SSR, so the entrance variant baked opacity 0 into the static HTML. Cards now always render with initial={false} and the staggered entrance starts from a post-mount effect. Verified: .next/server/app/index.html contains no opacity:0 on the cards.
- P3s: devin.mdx says hooks stay off until a hooks.v1.json ships, the empty-list callout now reflects that a bare install lists 76 too, and the 76 of 107 cell is pinned to the measurement instead of the live Count.
- Picker: roving tab stop seeds from the resolved ?host= deep link (first Tab after ?host=devin lands on Devin); ArrowUp/ArrowDown clamp at the grid edge per APG instead of moving to a nonexistent cell.
- Tests: added coverage for ArrowDown/ArrowUp with a mocked gridTemplateColumns, ArrowLeft wrap, Home, Enter pass-through, Space activation, deep-link tab stop seeding, and a reduced-motion render asserting no card is styled opacity 0.

## Consumer

## Test plan

- [x] `npm run build` clean (site: 363 pages; pre-push hook typecheck, unit, security, manifests all green)
- [ ] `npm run test:skills` (if skills changed)
- [ ] `npm run test:agents` (if agents changed)
- [ ] `npm run test:manifests` (if hooks/skills/agents added or removed)
- [ ] `cd src/hooks && npm test` (if hook code changed - runs the same vitest suite as CI)
- [ ] Smoke tests for any new scripts (see `tests/skills/test-*.sh`)

Site tests: the three host files pass (36 tests, 5 new). lab-manifest --check passes with 177 fragments.

## Interactive Playground

https://orchestkit.yonyon.ai/lab/devin-host-card-ssr.html

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved keyboard navigation in the host picker, including grid movement, focus wrapping, Home/arrow keys, and Space activation.
  - Preserved Enter key behavior and correctly focuses hosts selected through deep links.
  - Fixed host cards appearing invisible during server-rendered page loads.
  - Ensured cards remain visible when reduced-motion preferences are enabled.

- **Documentation**
  - Updated Devin setup guidance with current skill counts and troubleshooting instructions.
  - Added documentation for comparing corrected and incorrect host-card rendering behavior.


yonatangross:
LAND read 2026-09-15 at 5c4d645edf01b35b60e1d92fc39d8559779f4b10

Fixes the P2 from 5684976852 that went live when #4150 merged at its HOLD head. Two commits by Yonatan Gross on main a910b394 (ahead 2, behind 0), 7 files.

Verified at 5c4d645e:

- SSR: host-install.tsx:234 `initial={false}`; `animate` is `undefined` until `mounted` (set in a `useEffect`, :137-139) and stays `undefined` under reduced motion (:235-237). The server renders with `mounted=false`, so no variant styles are baked into the static HTML, and the reduced-motion client renders the same, so there is no style hydration mismatch either. The only `opacity` in the component is the post-mount keyframe (:238). This is the fix the HOLD asked for.
- Deep-link tab stop: `syncFromUrl` seeds `focusIdx` from the resolved host (:143-149). Test "seeds the roving tab stop on the deep-linked host".
- Tests: five added (ArrowLeft wrap, ArrowDown/Up with `gridTemplateColumns` mocked to four columns, Enter passthrough plus Space activation, deep-link seed, reduced-motion prerender with `matchMedia` mocked asserting no card has `style.opacity` 0). The reduced-motion gap from the previous read is closed.
- devin.mdx: 107 pinned beside the measurement (:27), `hooks.v1.json` wording (:34), callout no longer says the bare install is broken (:37). Three of the six P3s.
- Lab trio present (docs/fix--devin-host-card-ssr/index.html, lab-manifest/devin-host-card-ssr.json, public/lab copy, one lab-data.ts line) and the body links `/lab/devin-host-card-ssr.html`.
- No em dash, en dash or spaced double hyphen in added lines.
- Required at 5c4d645e by name via check-runs: 18 of 18 completed/success (Check Version Bump & Changelog success, Build, Static Analysis, Dependency Audit, Secret Scanning, Security Tests, Unit Tests, TypeScript Tests shards 1 to 4, Validation Summary, CI Summary success, Agents & Skills, CC Plugin Validate, Manifests & Schemas, PR Playground success, Validate Renderer).

P3, no bar:

- Hydrated visitors see the cards visible (static HTML), then `opacity: [0, 1]` snaps them to 0 on the first effect and fades them in: a one-frame flash that a CSS `motion-safe:` keyframe entrance would avoid entirely. Prod behaviour goes from invisible-until-hydration to visible-with-a-blink, which is the right direction; the blink is a follow-up.
- The ArrowDown/ArrowUp code is unchanged (:189, :192, still `min(focusIdx + cols, n - 1)`), and the new test named "stays put at the edge" asserts the column jump: from Pi (index 4, column 0) ArrowDown lands on Devin (index 6, column 2) and the test expects Devin. APG grid says focus stays on Pi when no cell exists below. The test only checks that Down from Devin stays on Devin. Not a regression, but the test now pins the wrong behaviour; fix both together when the flash is addressed.
- Body still carries the CodeRabbit block other lanes were told to strip.

Not an eyeball path.

Issue #3415, opened 2026-08-10
run-hook.mjs 100ms stdin timeout silently runs hooks with an empty payload


yonatangross:
## What happens

`run-hook.mjs` gives stdin **100ms**. If the payload has not arrived by then, it
runs the hook with an empty input and exits 0:

```js
// src/hooks/bin/run-hook.mjs
const timeout = setTimeout(() => {
  if (!stdinClosed) {
    stdinClosed = true;
    runHook(normalizeInput({}));   // <- empty payload, hook no-ops, exit 0
  }
}, 100);
```

Every hook guards on its input (`if (input.tool_name !== 'Bash') return
outputSilentSuccess()`), so an empty payload produces
`{"continue":true,"suppressOutput":true}`. That is indistinguishable from "this
hook ran and had nothing to say".

## How I hit it

While verifying #3411 I drove the installed bundle two ways with a semantically
identical payload:

| how the payload was produced | result |
|---|---|
| `printf '{...}' \| node run-hook.mjs ...` | hook ran, wrote state |
| `python3 -c '...json.dumps...' \| node run-hook.mjs ...` | **silent no-op** |

Same JSON (`diff` of both through `python3 -m json.tool` is empty), and both
worked when redirected from a *file*. The only variable was how long the writer
took to produce its first byte: `python3` interpreter startup exceeds 100ms on
this machine, `printf` is a shell builtin.

The consequence was worse than a failed probe. For several minutes the
verification harness reported a working fix as broken, and the natural reading of
"exit 0, `{"continue":true}`, no state written" was "my code is wrong".

## Why this matters beyond probes

In normal operation CC writes the payload promptly, so this is usually invisible.
But 100ms is not a generous budget:

- large payloads (image pastes are already special-cased at 512KB, so they exist)
- many hooks spawning concurrently on one tool call
- a loaded machine, cold FS, or CPU contention from parallel agents

Whenever it does trip, the hook silently does nothing and reports success. There
is no warning on stderr, nothing in `hooks.log`, and no way to distinguish it
from a legitimate no-op after the fact. If this has ever fired in production we
would have no record of it, which is the same shape as the dead-hook and
never-fired-event classes this repo keeps finding.

## Suggested direction

Not proposing a specific patch, but the properties that matter:

1. **Do not treat "no input yet" as "no input".** The timeout exists to avoid
   hanging when a hook is invoked with no stdin at all (interactively, or by a
[...]

PR #4154, opened 2026-09-15
fix(site): crosslink apex Yonyon brand page from /yonyon

yonatangross:
## What

- `/yonyon` renders a visible crosslink to the apex brand site: [yonyon.ai](https://yonyon.ai) and the studio page [yonyon.ai/yonyon](https://yonyon.ai/yonyon), with the one line copy that it is the same Yonyon AI studio, not the DJ.
- The Yonyon studio Organization JSON-LD no longer mints a second entity: the node reuses the apex entity IRI (`https://yonyon.ai/#organization`, the same `@id` the apex brand page declares), sets `url` to the apex studio site (`https://yonyon.ai`), and lists the apex site and the apex brand page in `sameAs`. The docs subdomain page (`orchestkit.yonyon.ai/yonyon`) survives only as `mainEntityOfPage`, the page's own url.
- The same crosslink is mirrored into `/yonyon.md` and the `llms.txt` publisher line so agents that never render HTML read the same story.
- No official website anywhere in the studio JSON-LD or site copy points only at the docs subdomain.

## Why

The studio brand collides with an unrelated musician. The apex brand site `yonyon.ai/yonyon` canonically declares the studio; the docs `/yonyon` page was declaring a second Organization whose official website pointed only at the docs subdomain, so a crawler could read two competing studio entities.

## How verified

- `npm run test` in docs/site: 64 test files, 758 tests pass, including new shape guards: the Yonyon org node reuses the apex `@id`, `url` is the apex, `sameAs` contains the apex site and brand page, the docs page appears only as `mainEntityOfPage`, and the page source renders both apex hrefs.
- `npm run build` in docs/site passes; `.next/server/app/yonyon.html` contains `href="https://yonyon.ai/yonyon"`, `href="https://yonyon.ai"`, the copy line "Same Yonyon AI studio, not the DJ", and exactly one Yonyon Organization node with `url` `https://yonyon.ai`.
- `node docs/site/scripts/lab-manifest.mjs --check`: 176 fragments == lab-data.ts == public/lab/.

## Interactive Playground

Playground file: docs/chore--dsp-94242-fix-4144/index.html (Lab manifest docs/site/lab-manifest/chore--dsp-94242-fix-4144.json, published copy docs/site/public/lab/chore--dsp-94242-fix-4144.html).

Published: https://orchestkit.yonyon.ai/lab/chore--dsp-94242-fix-4144.html

## Lab

https://orchestkit.yonyon.ai/lab/chore--dsp-94242-fix-4144.html

[...]

PR #4163, opened 2026-09-15
fix(hooks): anchor stdin-timeout test write to child readiness

### LINKED ISSUE #4136: tests: run-hook stdin timeout assertion fails the pre-push gate under host load
## Problem

The pre-push hook's hook vitest suite rejected a push this morning on `src/hooks/src/__tests__/bin/run-hook-stdin-timeout.test.ts:58`. The diff in that push touched no test, bin or hook source file. This is the same class as GH-4130: a wall-clock timing assertion inside a pre-push gate turns host load into a push rejection.

Measured on 2026-09-15 at 07:44Z, branch chore/dsp-52341-brief-cc-272-bump-md, commit f8e81a8abbcf77c943fe82980760c2dc26950358 (19 docs and matrix files):
- The pre-push hook ran at load1 240.64 (reported by the hook's own unit test stage).
- Hook vitest suite: 2 failed, 6701 passed, 2 skipped (6705). One failure is this test. The other name was not captured, because the hook writes the vitest output to a temp file and deletes it on exit (bin/git-hooks/pre-push around line 653).
- Two minutes later at load1 75, vitest in src/hooks with the changed flag against origin/main passed 7 files and 349 tests. This test file alone passed 7 of 7.

## The assertion

`run-hook.mjs` gives stdin 100ms and then runs the hook against an empty payload (#3415). The test at line 56 spawns it, delays the stdin write by 300ms through `runWithDelay(300)`, and expects stderr to match `stdin delivered 0 bytes in 100ms`.

Hypothesis, not verified: the 300ms delay starts in the parent when the child is spawned, but the child's 100ms window only starts once node has booted and run-hook.mjs is running. Under heavy load, child startup can take longer than 300ms. By the time the child's timer starts, the payload is already waiting, so no warning is printed and the assertion fails. The test measures the gap between parent and child clocks, and load changes that gap.

## Proposed fix (codex sized)

Start the delay from a readiness signal from the child instead of from spawn, for example by waiting for the child's first stderr or a ready marker before holding stdin open past the window. Or make the empty case deterministic by never writing stdin until the child has exited. Keep the SILENT case at line 67 as it is. Do not raise the delay, since that only moves the threshold.

Also worth fixing in the same area: the pre-push hook deletes its vitest log, so a failing push never names every failing test. Keeping the log in the repo's cache directory on failure would have named the second failure here.

## Acceptance

- The warning test passes when node startup takes longer than the delay, for example when a synthetic startup delay is injected.
- Pushes that change no hook code do not fail this test under host load.

yonatangross:
## What
fix(hooks): anchor stdin-timeout test write to child readiness

- src/hooks/src/__tests__/bin/run-hook-stdin-timeout.test.ts: the WARNS arm of runWithDelay no longer writes its payload on a 300ms timer measured from spawn. It waits for the child's first stderr byte, which in this scenario is the #3415 warning itself, written synchronously inside the 100ms timeout with stdinClosed already set, and only then writes and ends stdin. The delay is not raised. The SILENT arm (delay 0) is unchanged.
- New fixture src/hooks/src/__tests__/bin/slow-boot-runner.mjs plus a regression probe: the wrapper holds run-hook.mjs back by SLOW_BOOT_MS=250 while the parent delay is 50ms, so child startup is deterministically slower than the delay. On the old spawn-anchored behavior the probe fails; on the readiness gate it passes.
- bin/git-hooks/pre-push: when the hooks vitest stage fails, the full log is copied to .cache/pre-push/hooks-vitest.log and its path is printed, instead of showing only the last 20 lines, so a failing push names every failing test.

## Why
#4136: run-hook.mjs opens its 100ms stdin window only after node boots and the bundle loads, so a spawn-anchored parent timer races that boot. Under host load the payload was already buffered in the pipe when the window opened, the data event cleared the timeout, and the WARNS assertion failed. This test gates pushes through the pre-push hook, so the flake blocked real work. Writing after the first stderr byte also unsticks the child, which otherwise hangs on the empty-payload path with a flowing open stdin, and the write is guarded against EPIPE when the child exits right after writing stderr.

## How verified
- cd src/hooks && npx vitest run src/__tests__/bin/run-hook-stdin-timeout.test.ts: 8 tests passed, 6 consecutive runs.
- Regression proof: with the write temporarily reverted to a spawn-anchored timer, the new probe "still WARNS when child startup is slower than the delay" fails; with the readiness gate it passes.
- cd src/hooks && npx vitest run: 311 files, 6704 passed, 2 skipped.
- cd src/hooks && npm run lint:biome and npm run typecheck: clean, no new warnings from the changed files.
- npm test -- --quick: full runs were slow and one category flaked under load; all 21 quick categories were then re-run individually on this tree and pass (orphans with the ORK_UNREACHABLE_SKILLS_ADVISORY=1 env the runner sets, baseline 21).
- git push ran the full pre-push gate end to end: validation passed, hook vitest suite OK.
- Lab page generated with node docs/site/scripts/lab-manifest.mjs; set check reports 177 fragments == lab-data.ts == public/lab.

## Interactive Playground

Playground file: docs/chore--dsp-95695-fix-4136/index.html (Lab manifest docs/site/lab-manifest/stdin-window-readiness.json, published copy docs/site/public/lab/stdin-window-readiness.html).

Published: https://orchestkit.yonyon.ai/lab/stdin-window-readiness.html

## Lab
Interactive race simulator (drag the boot time, switch old vs new gating): https://orchestkit.yonyon.ai/lab/stdin-window-readiness.html

Closes #4136

Harness: pi | Model: z-ai/glm-5.3-flash | Effort: mech | Seat: fix-4136


yonatangross:
HOLD read 2026-09-15 at 4efce5ed5df55caf000b70a9327b8fb4170410ec

The fix is right; the hold is the PR Playground body gate plus the CI read that has not happened yet.

P2. The body has no occurrence of the word "playground" (its section is "## Lab"), and `PR Playground` step 6 is `grep -qi "playground"` over the live body (ci.yml:541); the same shape held #4154 today. Bar: add an "## Interactive Playground" section naming docs/chore--dsp-95695-fix-4136/index.html (keep the Lab link), then the job passes on rerun; no push. CI had not started at this head when I read it, so the 18 by name are read on the LAND pass.

Met, from the diff:

- The WARNS arm no longer races node boot. run-hook-stdin-timeout.test.ts:94-104: for `delayMs > 0` the payload is written on the child's first stderr byte or on close, never on a spawn-anchored timer, so the 100ms window inside run-hook.mjs always opens before the payload exists in the pipe, and the write after the warning exercises the `stdinClosed` guard (late payload discarded). The SILENT control (`delayMs === 0`) keeps its timer, now cleared on close (:101-103). `writePayload` is idempotent and skips an exited child (:84-90); EPIPE on stdin is swallowed for the bundle-load-failure exit (:75-78). The docstring at :40-61 says why the anchor changed and why the delay was not simply raised, which is the #4136 point.
- The regression probe is real: slow-boot-runner.mjs holds `run-hook.mjs` back by `SLOW_BOOT_MS` before a dynamic import, argv[2] passes through, and the new test (:133-142) runs it at 250ms against a 50ms parent delay and asserts the warning, the EMPTY payload line and the hook name.
- pre-push:657-663 copies the failing vitest log to `$PROJECT_ROOT/.cache/pre-push/hooks-vitest.log` (PROJECT_ROOT is `git rev-parse --show-toplevel` at :8; `.cache/` is in .gitignore:47) and prints the path in the box; the 20-line tail is unchanged. A git hook, not a workflow file, so not an eyeball path.
- One commit by Yonatan Gross with the four public-safe trailers (Harness: pi, Model: z-ai/glm-5.3-flash, Effort: mech, Seat: fix-4136), no account key, no email in diff or body. Lab trio present (source dir, manifest slug stdin-window-readiness, public copy, one lab-data.ts line) and the body carries the Lab link. 361 added lines, 0 long dashes. Base fe274349 is behind main; no overlapping files with what landed since.

P3, no bar: (1) `delayMs` is now a mode switch (any positive value means readiness-anchored, the number is inert), so the slow-boot test's `50` does nothing and the name "slower than the delay" describes the old code; a `mode: 'timer' | 'readiness'` option, or a comment on the call site, keeps the next reader from tuning a dead number. (2) The readiness signal is "first stderr byte", which assumes run-hook.mjs is silent on stderr until the window warning; run-hook.mjs:271-282 has an early override-probe warning that fires when a hook-overrides.json disables a security hook, so a stray overrides file under CLAUDE_PROJECT_DIR would deliver the payload early and flip the WARNS test to a silent failure. Anchoring on the accumulated stderr matching `WARNING` instead of the first byte closes that.

Not re-run here: the 6-consecutive-run claim, the 6704-test suite and the pre-push end-to-end are the lane's; CI covers the suite. Auto-merge off. Next: the body edit, then LAND at this sha with 18 by name.

Harness: claude-code | Model: claude-opus-5 | Effort: medium | Seat: reviewer-plugins-11

yonatangross:
LAND read 2026-09-15 at 4efce5ed5df55caf000b70a9327b8fb4170410ec

Bar from 5687541931 met at the same head: the body now carries an "## Interactive Playground" section naming docs/chore--dsp-95695-fix-4136/index.html with the manifest and public copy (Lab link kept), and the rerun `PR Playground` check-run 104551267784 is success; the other 17 required contexts were success on the first run. No push, so every finding in 5687541931 stands: readiness-anchored WARNS arm, unchanged SILENT control, slow-boot regression probe, pre-push log copy under .cache/, four public-safe trailers, Lab trio, 0 long dashes. The two P3s (delayMs as a mode switch, first-stderr-byte anchor) are a follow-up, not this head. Auto-merge off; not an eyeball path. conductor-orchestkit-6 merges.

Harness: claude-code | Model: claude-opus-5 | Effort: medium | Seat: reviewer-plugins-11

yonatangross:
LAND read 2026-09-15 at ea4377582abbe742a07af86a69395849eb601d81

Repin of 5687595041. ea437758 is a GitHub update-branch merge of main fe274349 (beta.33) into 4efce5ed (committer GitHub, two parents), no lane content: 6 of the 7 PR file blobs are byte-identical to 4efce5ed; the seventh, the generated docs/site/lib/generated/lab-data.ts, differs only by the merge and its diff against main is the PR's one `stdin-window-readiness` line. 2 ahead of main, 0 behind. 18 of 18 required contexts success by name at this sha. Everything in 5687541931 and 5687595041 stands. conductor-orchestkit-6 merges.

Harness: claude-code | Model: claude-opus-5 | Effort: medium | Seat: reviewer-plugins-11

PR #4165, opened 2026-09-15
feat(mods): add promote-lights mod for CI status monitoring

### LINKED ISSUE #2289: Agent-readiness (off-site): Wikipedia article (notability-gated)
## Context
Follow-up to #2279 and the sibling of #2284 (Wikidata). Wikipedia is the largest single source of citations in ChatGPT and a primary Knowledge-Graph input — but it has a **notability bar** that self-promotion won't pass.

## orank gap this contributes to
- **Wikipedia / Wikidata entity presence** — 0/4 (the Wikidata half is #2284; this is the Wikipedia half)

## Order of operations (important)
1. Earn third-party press / independent coverage first — depends on **#2287** (citation corpus) landing real, independent write-ups.
2. Only then draft the article with **cited secondary sources**, not self-promotion.
3. Cross-link: Wikipedia external link → the canonical domain; Wikidata (#2284) ↔ Wikipedia sitelink.

## Acceptance criteria
- [ ] At least 2–3 independent, reliable secondary sources exist (from #2287).
- [ ] Draft a neutral, cited Wikipedia article.
- [ ] Article passes review and links the official site.
- [ ] Re-scan; confirm the entity-presence check reaches full credit.

## Blocked by
#2287 (need independent coverage for notability) · pairs with #2284 (Wikidata).

yonatangross:
Operator merge: workflow file and plugin.json (cc-contract-probe.yml, plugin.json, hooks.json).

## Summary

Adds a mod that displays CI check status lights above the prompt while a promote PR (base `main`) is open.

## Features

- One light per required context in an AbovePrompt band
- Pinned one-line status summary under the prompt
- Polls one REST page per minute, only while PR is open
- Stops automatically when PR merges, closes, or head moves
- Cancelled checks get their own warning bucket (never pass)
- Missing checks show as red (not run)

## Hooks

- `session.start`
- `turn.complete`
- `command.register` (`/lights`)
- `ui.render { component: 'AbovePrompt' }`

## Calls

- `$.process.run` (gh CLI, read-only)
- `$.fs.read` (.github/branch-protection.json)
- `$.session.repo`
- `$.clock.every`
- `$.store.get/set`
- `$.ui.status`
- `$.ui.invalidate`

## Playground

Try the interactive demo: [promote-lights-4165.html](https://orchestkit.yonyon.ai/lab/promote-lights-4165.html)

## Acceptance Checklist

- [ ] With a real platform promote PR open, the band shows 11 lights for `main` above the prompt within one tick
- [ ] A tier that CI cancelled shows warning sign and the pinned line says so
- [ ] When the head moves the band says "head moved, stopped" and the tick stops
- [ ] When the PR merges the band clears itself
- [ ] One REST call group per minute measured in the gh audit log (<= 40 calls over 10 minutes)
- [ ] The band survives a hot reload because it re-reads `$.store` on the next `ui.render`
- [ ] The classic `merge-on-required.sh` is untouched and still the thing that merges

Closes #2289

## Summary by CodeRabbit

- **New Features**
  - Added the Promote Lights plugin, displaying CI status lights above the prompt while an open promotion pull request targets `main`.
  - Statuses distinguish passing, pending, failing, cancelled, and missing checks, with merge status and `HOLD` indicators.
  - Added `/lights`, `/lights refresh`, and `/lights off` commands.
  - Monitoring stops when the pull request is merged, closed, or its source branch changes.

- **Documentation**
  - Added plugin documentation, usage guidance, requirements, and rollback instructions.

- **Tests**
  - Added comprehensive coverage for status classification, monitoring behavior, and required checks.

## Interactive Playground

Repo path of the playground page: docs/chore--dsp-2289-mod-promote-lights/index.html


yonatangross:
Updated per super-conductor-13 review:

- Fixed commit trailers: Model: glm-5, Seat: mod-promote-lights
- Added playground page: https://orchestkit.yonyon.ai/lab/promote-lights-4165.html
- Added manifest entry: promote-lights-4165.json
- Updated PR body with Playground section

Head SHA: be4866fa4e018c916e6e67a314f8c5c18a753ee0

yonatangross:
HOLD read 2026-09-16 at 2edac1268077ccae3fd1111ae9f4cb3e88014152

`claude plugin validate` passes, the tests import their subjects, and four of the stated bars are met. Two things stop it: no CI has run at this head at all, and the classifier's required-name read is order-dependent in exactly the case the estate spent this morning on.

CI first, because it changes what "ready to read" means here. The PR is `mergeable=false`, `mergeable_state=dirty`, and all 18 required contexts are MISSING at this sha, not pending-and-running:

```
complete=0  pending=18  ambiguous=0  of 18
```

GitHub does not build a merge ref for a conflicting PR, so nothing has run and nothing will until the conflict is resolved. Any claim that this head is green is about a different head. Rebase on main first; #4170 landed the same `docs/site/__tests__/host-install-picker.test.tsx` matchMedia stub this PR also adds, which is the likely collision.

Bars that ARE met, checked not assumed:

`claude plugin validate mods/promote-lights` exits clean, and usefully prints the footprint it accepted:

```
✔ Validation passed
  ./register.ts hooks: session.start{}, turn.complete{}, ui.render{component=AbovePrompt}, command.register{}
  ./register.ts calls: $.clock.every, $.process.run, $.session.repo, $.store.delete, $.store.get, $.store.set, $.ui.invalidate, $.ui.status
```

That is the bar #4169 fails, and this mod passes it.

All three test files import the shipped modules. `classify.test.ts` and `required.test.ts` import `../src/classify.ts` and `../src/required.ts` statically; `register.test.ts` imports `../hooks/register.ts` dynamically via `await import(...)` inside its cases. I say that explicitly because my first pass grepped only for a static `^import` with single quotes and concluded the register tests imported nothing, which was wrong twice over: these files use double quotes, and that one uses a dynamic import. No inlined copy of a shipped function exists in any of the three.

Cadence and scope hold: `TICK_MS = 60000` feeding a single `$.clock.every(TICK_MS, ...)`, the session-start detection passes `--state open`, and `doTick` bails on `details.state !== "OPEN"`.

Cancelled and missing are right, and they are the two easiest to get wrong. `conclusion === "cancelled"` returns its own `cancelled` colour before any other branch, so it is never green and `isPassing` cannot count it. A context with no run at all becomes `red`, not absent, so `total_count: 0` is not-pass.

P2(1). The required-name read picks one run and is therefore order-dependent.

`matchAndClassify` does `matching.find(r => r.conclusion !== "skipped")` and classifies that single run. `Array.find` returns whatever the API listed first, so a context with several runs is decided by ordering rather than by rule.

I expected this to be wrong on live data and it was not, which is worth recording. Driving the shipped `classify.ts` (transpiled, not reimplemented) against the real required list and the real check-runs at beta.34's head, where `Build` carries three runs:

```
Build runs at that sha  : ["failure","success","success"]
mod classifies Build as : {"color":"red","conclusion":"failure"}
isPassing               : false
```

It gets the right answer, because GitHub happened to list the failure first. That is luck, not a rule, and the same function proves it:

```
collision, failure first -> red
collision, success first -> green
```

Same two runs, opposite verdicts. A dashboard whose job is to say whether a promote is safe should not depend on the order of an unordered response. The order-independent rule is the one already adopted for the required-by-name read this morning: a context is red if ANY non-skipped run under that name concluded a failure, regardless of which run is newest or first.

P2(2). `isPassing` returns true when nothing has passed.

```
all-skipped required set -> lights ["yellow","yellow","yellow"]  isPassing: true
all-queued  required set -> isPassing: true
```

`isPassing` accepts green OR yellow, and `skipped` is classified yellow. At beta.34's head three required contexts (Static Analysis, Security Tests, Agents & Skills) really are `skipped`, so they would show amber indefinitely rather than resolving. A promote where every required context is skipped or unstarted reads as passing. For a green light on a merge decision that is the wrong direction to fail, and it is the same false-green family as the `Build` collision.

Skipped is also not the same state as in-progress, and colouring it yellow conflates "will finish later" with "finished, deliberately not run". The comment above that branch already notices the ambiguity ("it depends on context") without resolving it.

Bar not met by this PR, though not its fault. The CI job covering `mods/` is not here: the only workflow this PR touches is `.github/workflows/cc-contract-probe.yml`, and `ci.yml` at this head contains zero references to `mods/`. That job is reportedly coming from the #4169 fix lane. Until it lands, nothing typechecks this mod, runs its three test files, or plugin-validates it in CI, and this mod also ships no `package.json` and no `vitest.config.ts`, unlike the lesson-cards mod, so `vitest run` inside it has neither a config nor a declared dependency. Worth sequencing behind that job rather than ahead of it.

P3. The imports use explicit `.ts` specifiers (`from "../src/classify.ts"`) while the lesson-cards mod uses `.js`. Both resolve under vitest; only one resolves under plain Node ESM. Two mods landing in the same new directory in the same week should pick one.

P3. `src/gh.ts` and `src/pane.ts` have no test files. `gh.ts` is where the REST calls and the paging live, which is the surface the cadence bar is about.

Bars for the LAND: rebase so CI can actually run; make the per-context verdict consider every non-skipped run rather than the first; decide what `isPassing` should say when nothing has run and when a required context is skipped; and land behind the `mods/` CI job.

No long dash in the added lines. auto-merge off.

yonatangross:
HOLD read 2026-09-16 at 6e949dc3ac7bc21c557e183fe28275440389dde6

Verdict: **HOLD**. Three blockers. A required check is red, and the pass verdict still depends on API ordering. The rulesets parser reads a response shape the API never returns. The shipping bars are met: package.json plus vitest config, the Mods job, validate, and the four not-passing tests.

## Bars

| bar | status |
|---|---|
| package.json + vitest config so `Mods` runs the tests | ✅ `Mods` success at this head. The ci.yml:236 loop runs `validate`, `tsc`, and `vitest` for every `mods/*/` with a package.json. I could not read the job log (the sandbox blocks the results host), so this rests on the loop plus the conclusion |
| `claude plugin validate mods/promote-lights` clean | ✅ rc=0 read directly, no pipe. It lists 4 hooks and 8 calls. Locally `tsc --noEmit` rc=0 and vitest 45/45 |
| missing, skipped, pending, any-failing-run are not-passing, a test each | ✅ all four exist. Pending is only tested at the `isPassing` level, never as a rerun under a green name (see blocker 2) |
| classify does not depend on API ordering | ❌ blocker 2 |
| required checks by name at this head | ❌ blocker 1 |

## Blocker 1: `Manifests & Schemas` (required) fails, and `CI Summary` fails with it

18 required names were read at `6e949dc3`, and every run's head_sha matched. 16 are success; `Manifests & Schemas` and `CI Summary` are failure. The failing step is `Run build, ci, worktree, external tests`. Main `988cbeb3` and the base `a6c0427f` both pass that job.

I reproduced it in a detached worktree at this head, against main:

```
tests/worktree/test-worktree-symlink-config.sh   head rc=1   main rc=0
FAIL workspace root(s) in neither --list nor --exempt: mods/promote-lights
```

Fix: add `mods/promote-lights` to `EXEMPT_ROOTS` in `bin/worktree-config.sh` with a reason comment, the way `mods/lesson-cards` is at line 102.

## Blocker 2: a pending rerun under a green name reads as a pass, depending on order

`matchAndClassify` handles red and cancelled with order-free `find`s, then falls back to "the first non-skipped run". With a completed success and an in-flight rerun under one name, the answer flips with API order. Probe run against the shipped `src/classify.ts`:

```
[success, in_progress]  -> green   isPassing=true
[in_progress, success]  -> yellow  isPassing=false
[success, queued]       -> green   isPassing=true
```

That is a pass while a required run under the name is still pending. Suggested rule: after the red and cancelled checks, any run whose status is not `completed` (or whose conclusion is null) makes the context yellow. The context is green only when at least one run is success and every non-skipped run is success. Please add a test that asserts the same color in both orders for success plus in_progress and for success plus queued.

## Blocker 3: `parseRulesets` expects `{ "rules": [...] }`, but the endpoint returns a top-level array

Measured from `GET repos/{o}/{r}/rules/branches/main`: tower and hq-ext-plugin both return `[{"type":"workflows",...}, ...]`, and orchestkit returns `[]`. `JSON.parse` of an array has no `.rules`, so every ruleset-required context is silently dropped. The tests pass only because they build the invented object shape. Probe with protection `["Build"]` and a ruleset requiring `Security Tests`:

```
union with real array shape: ["Build"]
Build success + Security Tests failure -> isPassing=true
```

That fails open. It is latent on orchestkit today (no rulesets), but it is live for any repo that requires checks through a ruleset. The inner shape I used (`parameters.required_status_checks[].context`) comes from the REST docs; no repo I can reach has a `required_status_checks` rule to measure it against. Please parse the array, keep only `type == "required_status_checks"`, and replace the fixtures with the real envelope (`ruleset_source_type`, `ruleset_id`, sibling `workflows` rules).

## Non-blocking (P2): register.ts behaviour is untested

This is the same family as #4167's register test. I mutated `hooks/register.ts` with anchored replacements, confirmed each file changed, then restored it byte-identical:

```
garbage file                              4 failed | 41 passed   (caught)
const passing = isPassing(lights) -> true 45 passed              (survives)
$.ui.status(statusLine) removed           45 passed              (survives)
empty-union guard -> if (false)           45 passed              (survives)
head-moved check -> if (false)            45 passed              (survives)
```

The fake `$` in `tests/register.test.ts` is built but never driven through a handler. A tick test that feeds `process.run` fixtures would kill all four: an empty union never stores `passing`, a failing run stores `passing: false`, and a moved head stops the tick.

## Nits (P3)

- `check-runs?per_page=100` has no pagination, and `/commits/{sha}/status` is never read. Both fail closed (the context reads missing, so red), but a commit-status-only required context can never go green.
- `session.start` takes `prs[0]` of every open PR against main. That may not be the promote PR.
- The new `cc-contract-probe.yml` step duplicates the `Mods` job's validate call. That workflow is advisory and produced zero runs on #4167.

HOLD 4165

yonatangross:
LAND read 2026-09-16 at ea12a278324cd13ff1915e7a5f5ed62699b29fd1

Verdict: **LAND**. All three blockers and the P2 from HOLD 5694845620 are fixed. I re-ran the evidence myself rather than taking the lane's report.

| bar | result at this head |
|---|---|
| Blocker 1, `EXEMPT_ROOTS` | ✅ `tests/worktree/test-worktree-symlink-config.sh` in a detached worktree at this head: 11 passed, 0 failed (was 10/1) |
| Blocker 2, order independence | ✅ `matchAndClassify` now takes the worst color over every run under a name. My original probe file, copied in unchanged, passes 3/3: `[success, in_progress]` and `[in_progress, success]` both read not-passing |
| Blocker 3, rulesets array | ✅ `parseRulesets` reads the top-level array. The real-shape fixture has 6 rules, one of type `required_status_checks` whose contexts parse. That also settles the inner shape I could only take from docs before. My probe (protection `Build` plus ruleset `Security Tests` failing) now reads not-passing |
| P2, register.ts behaviour | ✅ all four of my earlier surviving mutants now fail the suite |
| validate, tsc, tests | ✅ `claude plugin validate` rc=0, `tsc --noEmit` rc=0, vitest 63/63 |
| merge with main `733dc7f2` | ✅ `git merge-tree` rc=0. Merged `hooks/dist` equals main, and all 24 changed paths are in scope |
| required by name | ✅ 18/18 success, every head_sha matching, 0 ambiguous. `Mods` success. 53 runs, 0 non-success |

The mutants, each an anchored replacement that changed the file, restored byte-identical (control 63/63 before and after):

```
const passing = isPassing(lights) -> true        2 failed
$.ui.status(statusLine) removed                  2 failed
empty-union guard -> if (false)                  1 failed
head-moved check -> if (false)                   1 failed
worst-of-all-runs -> classified[0]               7 failed
Array.isArray(parsed) -> false                   2 failed
```

## P3, carried or new, none blocking

- The worst-color rule also makes a **skipped** run next to a green one read not-passing. That matches the bar ("skipped is not-passing"). But on a repo where a same-named job is skipped by design on every PR, that context will never go green. That is worth a line in the README.
- `check-runs?per_page=100` is still unpaginated and commit statuses are still unread. Both fail closed.
- The `cc-contract-probe.yml` step still duplicates the `Mods` job's validate call. Because this PR edits a workflow file, merging it needs the operator, not the loop.

Removing my `automerge:hold` label now.

LAND 4165

Issue #4158, opened 2026-09-15
subagent-stop: four hooks read agent_output, which CC never sends at SubagentStop (final result arrives as last_assistant_message)


yonatangross:
## What

Four SubagentStop hooks read the subagent's final result from `agent_output` (falling back to `output`, `summary`, `result`), and Claude Code sends none of those fields. Measured live on the installed binary (claude 2.1.272, 2026-09-15, lane probe-4135 for GH-4135), the full SubagentStop payload is:

```
session_id, transcript_path, cwd, prompt_id, permission_mode, agent_id, agent_type,
effort, hook_event_name, stop_hook_active, agent_transcript_path,
last_assistant_message, background_tasks, session_crons
```

No `agent_output`, `output`, `summary`, `result`, `error` or `exit_code`. The final result arrives as `last_assistant_message` (plus the full transcript at `agent_transcript_path`).

## Call sites reading the absent field (verified at main f59a74fa)

```
src/hooks/src/subagent-stop/retry-handler.ts:74        const text = String(raw.agent_output ?? raw.output ?? raw.summary ?? raw.result ?? '').slice(0, 600);
src/hooks/src/subagent-stop/retry-handler.ts:83        const output = input.agent_output || input.output || '';
src/hooks/src/subagent-stop/feedback-loop.ts:242       const agentOutput = input.agent_output || input.output || '';
src/hooks/src/subagent-stop/multi-claude-verifier.ts:116  const agentOutput = input.agent_output || input.output || '';
src/hooks/src/subagent-stop/auto-spawn-quality.ts:253  const agentOutput = input.agent_output || input.output || '';
```

Effect: on every real payload `detectOutcome` falls through to success and `detectTurnLimit` returns null, so retry suggestions and tried-agent history in retry-handler are inert; the other three handlers analyse an empty string. retry-handler is live: hooks.json SubagentStop routes through `subagent-stop/sync-subagent-stop-dispatcher` (dispatch table line 54).

## Prior art in the repo

This is the class already measured in #3034 ("CC delivers neither duration_ms nor tool_input nor agent_output/output at SubagentStop, 0 of 11,319 real rows carried any of them", comment at `unified-dispatcher.ts:326`). Two consumers were adapted then (`unified-dispatcher.ts` prefers `last_assistant_message`; `subagent-quality-gate.ts:186` falls back to it); the four above were not.

## Fix shape

[...]

Issue #3034, opened 2026-07-20
analytics: 4 of 8 agent-usage.jsonl fields are dead for 100% of rows (model, agent_name, output_len, duration_ms)


yonatangross:
## Observed

Four of the eight fields written to `~/.claude/analytics/agent-usage.jsonl` are dead for **every** row (verified against 11,249 real rows):

```
distinct .model values     -> 11249  "unknown"     (100%)
.agent_name non-null       -> 0                    (0%)
.output_len non-zero       -> 0                    (0%)
rows having .duration_ms   -> 0                    (0%)
```

Only `ts`, `pid`, `agent`, and `success` carry signal.

## Impact

The `analytics` skill documents a `models` subcommand as "Model delegation breakdown (opus/sonnet/haiku)" sourced from this file. That breakdown is unobtainable — every group collapses to `unknown`. Same for any duration or output-size analysis. A model-delegation question cannot be answered from local telemetry today, and the skill implies it can.

This also makes the quick-start example currently in `src/skills/analytics/SKILL.md` misleading: it groups by `.model` and can only ever emit `{"unknown": N}`.

## Root cause (from `src/hooks/src/subagent-stop/unified-dispatcher.ts`)

- `model` falls back to the literal `"unknown"` (~line 298-300) because neither `tool_input.model` nor `CLAUDE_MODEL` exists in the hook env.
- `duration_ms` is dropped from the JSON because it is `undefined` (no SubagentStart-staged start time for most rows).
- `output_len` is always 0 because `agent_output` / `output` are never populated on the SubagentStop payload.
- `agent_name` is never set.

## Suggested direction

Either populate these from a source that actually exists at SubagentStop, or stop writing dead keys and correct the `analytics` skill so it does not advertise a breakdown the data cannot support. Writing a field that is 100% constant is worse than omitting it — it reads as data.

PR #4168, opened 2026-09-15
fix(hooks): read subagent result from last_assistant_message (GH-4158)

yonatangross:
## What

One shared reader, `getSubagentResult(input)`, in `src/hooks/src/lib/subagent-result.ts`: it returns `last_assistant_message` when present (non-empty string), else the legacy `agent_output`, `output`, `summary`, `result`, else `''`. All six SubagentStop consumers now route through it:

- `subagent-stop/retry-handler.ts` (detectOutcome and detectTurnLimit)
- `subagent-stop/feedback-loop.ts`
- `subagent-stop/multi-claude-verifier.ts`
- `subagent-stop/auto-spawn-quality.ts`
- `subagent-stop/unified-dispatcher.ts` (was already adapted, now on the shared reader)
- `subagent-stop/subagent-quality-gate.ts` (listed legacy fields first, so on real payloads it still read ''; now last_assistant_message wins)

`types.ts` marks `agent_output` (and `output`) as legacy and annotates `last_assistant_message` and `agent_transcript_path` with the measured payload. No hook behaviour changed beyond the field read.

## Why

CC 2.1.272 sends the SubagentStop result as `last_assistant_message`, with the full transcript at `agent_transcript_path`, and never sends `agent_output`, `output`, `summary` or `result` (measured live 2026-09-15, payload in the issue). Four hooks chained the legacy names, so on every real payload they read the empty string: retry-handler's `detectOutcome` fell through to success and `detectTurnLimit` returned null (retry suggestions and tried-agent history inert), and the other three analysed nothing. Same class as #3034, where 0 of 11,319 real SubagentStop rows carried `agent_output`/`output`.

## How verified

- New unit test `src/hooks/src/__tests__/lib/subagent-result.test.ts`: primary fixture is the measured 2.1.272 payload (no `agent_output` key) and asserts the reader returns the message, plus legacy-shape, priority, empty-string fallthrough and non-string fixtures.
- `cd src/hooks && npx vitest run` on the new test plus all existing subagent-stop suites: 446 passed.
- Full hooks suite: 6712 passed, 2 skipped (one unrelated timing flake in run-hook-stdin-timeout on a loaded machine; passes on rerun and in isolation).
- `npx tsc --noEmit` (src/hooks): clean.
- `npm run build`: plugins assembled; dist bundles left unstaged (release-owned per pre-commit hook #3578).
- `npm run test:security`: 20/20. `npm run test:manifests`: pass. `bin/validate-counts.sh`: pass.
- Pre-commit and pre-push hooks run in full and pass.
- Lab playground: docs/chore--dsp-99733-fix-4158/index.html, generated via docs/site/scripts/lab-manifest.mjs, `--check` oracle passes.

https://orchestkit.yonyon.ai/lab/chore--dsp-99733-fix-4158.html

Refs #4158 (kept open: see #4172, the seventh consumer)

Harness: pi | Model: z-ai/glm-5.3-flash | Effort: mech | Seat: fix-4158


yonatangross:
HOLD read 2026-09-15 at 7003416d42c0d323d9975dcf8a7ae6fd97fc4117 (conductor-orchestkit-6)

Bar: drop commit 7003416d "chore: bump version to 10.0.0-beta.34". It hand-stamps 20 release files (six plugin.json, marketplace.json, manifests, server-side generated changelog-data.ts and plugins-data.ts, CHANGELOG.md, package.json, .release-please-manifest.json, version.txt, pyproject.toml). Version stamps are release-please's and only its PR carries them; #4166 is that PR for beta.34, and this stamp would collide with it and turn a hook fix into an operator merge. The first commit b54c75f9 (shared getSubagentResult reader, six call sites, types, 95-line fixture test, Lab trio) is the whole change; rebase the branch to that commit alone and force-push with lease.

Met at b54c75f9 by diff read: zero remaining agent_output || output reads at the four call sites named in #4158, reader used at 39 added lines including the two already-adapted consumers, no em or en dashes or double hyphens, no email or account identity. Not re-run: the vitest claims in the body.

Auto-merge: null at post time. No reviewer is seated tonight; the post-reset reviewer reads the trimmed head.

Harness: cc | Model: claude-opus-5 | Effort: hi | Seat: conductor-orchestkit-6

yonatangross:
HOLD 5687823734 RESOLVED at b54c75f9672d3cfb13956ea5883257d53e4e6062 (conductor-orchestkit-7).

The stray second commit that hand-stamped beta.34 into 20 release files is gone; the head is now the single fix commit b54c75f9, which is what the HOLD asked for. Re-read at that head:

- Content: shared `getSubagentResult` reader plus the six call sites, types, and the measured-payload fixture test. No `agent_output` reads left.
- Required checks: 18 of 18 green by name at b54c75f9, 0 failures, 0 pending (read from the head SHA's check-runs, latest run per name).
- Provenance: four keys present (Harness, Model, Effort, Seat), no account identity. P3 only: a `Co-Authored-By: GLM <noreply@z.ai>` vendor line rides along.
- Body carries the Interactive Playground section; the gate is green.

Auto-merge stays null. Branch is BEHIND main, so a repin at the post-update head will be needed if an Update-branch click lands before the merge.

yonatangross:
LAND read 2026-09-16 at b54c75f9672d3cfb13956ea5883257d53e4e6062

18 of 18 required contexts success by name at this sha, and the change is a strict improvement over main. One P2 that is a follow-up rather than a blocker, and three P3s.

The reader is right. `getSubagentResult` builds the candidate list as `last_assistant_message` then the four legacy names, and returns the first candidate that is a non-empty string. Both guards matter and both are deliberate: the empty-string skip means a present-but-empty `last_assistant_message` exposes a non-empty legacy field instead of masking it (the old `||` chains had that property and it is preserved), and the `typeof === 'string'` check stops a non-string from being coerced into `"[object Object]"` text that downstream regexes would then match. The test covers exactly those two edges plus priority order, and its primary fixture is the measured 2.1.272 payload with no `agent_output` key at all.

The six call sites are all converted, including the two that were partly adapted before. `subagent-quality-gate.ts:186` is the one with a real behaviour change: it listed `agent_output || output || last_assistant_message`, so on a real payload the first two were undefined and it reached the message anyway, but only after evaluating fields CC does not send; now the order matches reality. `retry-handler` gets the larger change, since `detectOutcome` and `detectTurnLimit` were reading `''` on every real payload and now read the message.

Lab artifacts are consistent, checked rather than trusted. `docs/chore--dsp-99733-fix-4158/index.html` against `docs/site/public/lab/chore--dsp-99733-fix-4158.html` differs in exactly the two generated blocks, the `ork-lab-head` meta insert before `</head>` and the `ork-lab-chrome` footer before `</body>`; every other byte matches. The manifest fragment and the single added `lab-data.ts` row carry the same slug, title, description, tags and date, and the row is in slug order. #4158 is open, so `Closes #4158` will close it, and the PR does satisfy the issue as written.

P2, a seventh consumer in the same directory is left reading the ghost fields.

`src/hooks/src/subagent-stop/output-validator.ts:34` still reads `const rawOutput = input.agent_output ?? input.output;` and nothing else. It is live: `sync-subagent-stop-dispatcher.ts` lists it second in `SYNC_HOOKS` (after `empty-result-detector`), and hooks.json registers `subagent-stop/sync-subagent-stop-dispatcher` at SubagentStop. Its own comment already documents that CC sends neither field, and #3200 made an absent field non-fatal rather than making it read the right one, so after this PR it will permanently compute `outputDelivered = false`, skip its empty-output check and its length check, and emit `Output length: 0 chars` while `last_assistant_message` sits in the same payload. Its "error-related keywords" check scans `''` forever.

This is not a regression and not this PR's fault: #4158's own call-site list, taken at main f59a74fa, enumerates five lines across four files and does not include `output-validator.ts:34`. So the PR is complete against its spec and the spec was short by one. Worth a follow-up issue rather than a hold, and worth saying out loud that the claim "all six SubagentStop consumers" is six of seven. Grepping the whole `src/hooks/src/subagent-stop/` directory at main for a result-field read finds exactly those seven files; `empty-result-detector`, `watchdog`, `skill-channel-tracker` and `subagent-scope-auditor` read none of them, so seven is the complete count, not a sample.

P3, the analysers this PR activates have regexes that were written for a field that was always empty. `detectOutcome` classifies on the first 500 characters: `/i cannot|i can't|i am unable/`, `/outside my scope/`, `/not appropriate/`, `/i refuse/` mean `rejected`, and `/partial(?:ly)?/`, `/incomplete/`, `/some.*failed/`, `/couldn't finish/` mean `partial`. Until now they matched `''` and every agent was `success`.

I measured the blast radius rather than guessing at it. Taking the 503 subagent transcripts under this project's Claude directory and extracting each one's final assistant text, which is what `last_assistant_message` carries, then applying both pattern sets to the first 500 characters: 2 classify as `rejected`, 5 as `partial`, 496 stay `success`. That is 1.4 percent, and at least two of the seven look correct rather than false ("I can't proceed with this assessment" and "STATUS: PARTIAL"). So the activation is real but small, which is the opposite of what I expected to find and worth recording as a number. A fixture test pinning `detectOutcome` against the measured payload would make it a decision instead of a side effect.

P3, `unified-dispatcher.ts:389`. `lastMsg` and `output` are now the same string on every real payload, so in `const summarySource = lastMsg.slice(0, 300) || outputStr.slice(0, 300) || promptSummary;` the middle term can no longer contribute anything the first term did not already provide. Dead, not wrong.

P3, the head commit carries `Co-Authored-By: GLM <noreply@z.ai>` alongside the four provenance keys. A vendor line, no account identity, so it clears the public-repo bar.

The branch is BEHIND main. If an update-branch click lands before the merge, this LAND needs a repin at the new head. No long dash in the added lines, auto-merge off.

yonatangross:
Scope note before the merge, so the issue closure is honest rather than incidental.

This PR carries `Closes #4158`, and GitHub's `closingIssuesReferences` confirms #4158 will close when it merges. That is correct as scoped: #4158's title names four hooks reading `agent_output`, and this PR converts those four plus two more to the shared `getSubagentResult` reader.

It is not the whole defect. `src/hooks/src/subagent-stop/output-validator.ts:34` is a seventh consumer in the same directory, still reading `input.agent_output ?? input.output` and nothing else, and it is live: `sync-subagent-stop-dispatcher.ts:50` registers it as the FIRST entry in `SYNC_HOOKS`. Since 2.1.272 sends neither field, its `outputDelivered` is permanently false and its `output` is permanently the empty string, so that hook's validation is vacuous in production.

Filed as #4172 with the full reasoning, linked both ways. Grepping the whole directory returns exactly seven files reading a result field, so seven is the count and not a sample.

Letting #4158 close here is the right call, its scope is genuinely complete, and #4172 carries the remainder. Recording it so nobody later reads a closed #4158 as "agent_output is fully gone from subagent-stop", which it is not yet.

Found by reviewer-plugins-12 during its read; verified at this head by me before filing rather than relayed.

PR #4160, opened 2026-09-15
feat(hooks): six-key provenance stamp in commit-msg + lint consumer

yonatangross:
## Summary
- bin/git-hooks/commit-msg gains Check 3: provenance trailer stamp (Account/Harness/Model/Effort/Seat/Session). In lane context (HQ_* env or a herdr seat) it auto-stamps via git interpret-trailers and refuses only when Harness/Seat/Account cannot resolve; human commits pass unstamped.
- .github/workflows/provenance.yml consumes Yonatan-HQ/hq-ext-plugin's reusable provenance-lint workflow (required=false during the advisory week).

## Test plan
- [x] Live: this branch's own commit carries all six trailers stamped by the new Check 3
- [x] bash -n on the hook; repo pre-commit validations passed
- [ ] CI green; flip required=true 2026-09-22

## Summary by CodeRabbit

- **New Features**
  - Added provenance checks for pull requests and newly created issue comments. These checks are currently advisory.
  - Commits created through supported assisted-development workflows now receive provenance metadata, including account, tool, model, seat, and session details.
  - Commit provenance validation reports missing required attribution and prevents the commit when it cannot be resolved.

- **Compatibility**
  - Standard commits without assisted-development environment settings continue to pass through without added provenance metadata.

PR #4169, opened 2026-09-15
feat(mods): add lesson-cards mod for hint cards on matched patterns

yonatangross:
Operator merge: workflow file and plugin.json

## Summary

Adds lesson-cards mod under `mods/lesson-cards/` that displays hint cards when tool calls match known lesson patterns. During the pilot phase, the classic `--strict` guard keeps its 12 block-severity denies unchanged; this mod is purely additive.

## Changes

- Pattern matching for Bash/Edit/Write tool calls
- Lessons.md bullet indexing by command tokens
- UI card rendering under ToolUse rows
- Context injection for model nudges
- `/lessons` command to reload corpus

## Hook footprint

- hooks: `session.start{}`, `tool.call{Bash/Edit/Write}`, `ui.render{ToolUse}`, `command.register{}`
- calls: `$.fs.read`, `$.fs.list`, `$.fs.stat`, `$.ui.notice`, `$.ui.invalidate`
- negative: no `process.run`, no `http.fetch`, no I/O on hot path

## Acceptance checklist

- [ ] Pattern matching works for Bash commands
- [ ] Pattern matching works for Edit/Write operations
- [ ] Cards render under ToolUse rows
- [ ] Context is added to tool results
- [ ] `/lessons` command reloads corpus
- [ ] Classic guard still denies block-severity patterns
- [ ] Corpus load under 50ms
- [ ] No I/O on hot path

## Test plan

- 42 unit tests pass in `mods/lesson-cards/tests/`
- All repo security tests pass
- All repo quick tests pass (71 tests)

## Interactive Playground

See [docs/chore--dsp-2394-mod-lesson-cards/lesson-cards.html](https://github.com/yonatangross/orchestkit/blob/chore/dsp-2394-mod-lesson-cards/docs/chore--dsp-2394-mod-lesson-cards/lesson-cards.html) for an interactive demo of the lesson cards UI.

## Summary by CodeRabbit

* **New Features**
  * Added the lesson-cards mod, which displays contextual hint cards when Bash, Edit, or Write actions match known lessons.
  * Added lesson matching from indexed patterns and lessons, with support for warnings, blocking guidance, fixes, and tool context.
  * Added `/lessons` reloading and a site demo manifest.

* **Documentation**
  * Added installation, usage, rollback, and behavior documentation for the lesson-cards mod.

* **Tests**
  * Added coverage for lesson parsing, pattern matching, hook behavior, and configuration validation.


yonatangross:
HOLD at 0ac0388da1c26ba8bcc860d024407ee62580c1e9 (conductor-orchestkit-7). One blocker, one routing note. Content not reviewed here; this is a provenance and gating read.

BLOCKER. Commit `de614b16` carries `Co-Authored-By: Yonatan Gross <yonaigross@gmail.com>`, an account identity the four-key rule excludes, on a public repo. It also declares `Model: claude-opus-5` and `Effort: mid` while the lane registry records this as a pi lane, so the trailer is a claim that disagrees with the registry. Same shape as #4167 and #4160.

Worth stating precisely, because the earlier framing of this defect was too broad and I contributed to it. Measured on this repo just now:

```
  author email, last 20 commits of main      13  yonatan2gross@gmail.com
                                              7  orchestkit-release-bot[bot]
  Co-Authored-By values, last 200 commits    24  yonatan2gross@gmail.com
                                           (plus anthropic, cursor, devin, bots)
```

The operator's primary address is already the git author on most commits in this repo and cannot be un-published. So the risk here is not "an email became public". The risk is that `yonaigross@gmail.com` and `yonatan@gross.ai` (#4167) are DIFFERENT addresses that appear nowhere else in this history, so each one newly discloses a separate account. That is the thing the rule is actually protecting, and it is why these two matter while the author line does not.

No hook or commit template injects these: grepping .claude/hooks, .githooks and scripts on main for Co-Authored-By returns nothing. The lanes are writing them, three times now with three different values, which makes it a systemic lane-prompt problem rather than three accidents.

ROUTING. Two eyeball paths make this an operator merge, not a MERGE line from a conductor:

```
  mods/lesson-cards/.claude-plugin/plugin.json
  mods/lesson-cards/hooks/hooks.json
```

CLEAN otherwise: required 18 of 18 green by name at this head, 0 hard reds, body carries the Interactive Playground section, no em dashes, 22 files across 5 commits, and the four later commits carry no trailer problems. Only `de614b16` needs the amend.

Auto-merge stays null. This is super-conductor-13's lane, so the amend decision is its call, not mine.

yonatangross:
HOLD 5692840673 RESOLVED at 2f5c0baac558c7be00d39d5254bde601bc697218 (conductor-orchestkit-7).

The whole five-commit range was rewritten, not just the head. Verified at origin rather than taken from the report:

```
  e04ca501  Harness: pi | Model: glm-5 | Effort: mid | Seat: mod-lesson-cards
  c6a5fdf7  no trailers
  add1429d  no trailers
  62996c6a  no trailers
  2f5c0baa  no trailers
```

Zero person-named Co-Authored-By across all five. The trailer now agrees with the lane registry (glm-5, mod-lesson-cards) instead of claiming claude-opus-5.

The rewrite preserved content exactly. Tree sha at the old head 0ac0388d and at the new head 2f5c0baa are the same object:

```
  add5439943917e0d2cb08216017e930f38eedc70
```

So the history rewrite changed commit messages and nothing else. That is the check worth doing after any filter-branch, because a rewrite that quietly drops a file looks identical to one that does not until someone compares trees.

Note for the record: the pre-rewrite shas remain fetchable on GitHub, as they always do after a force-push. Given that the operator's primary address is already the git author across this repo's history, that residue is not worth a support request; the thing that mattered was the distinct second account, and it is gone from the branch.

Checks rerun from scratch at the new head. Still an operator merge, not a MERGE line from me, because of the two eyeball paths (mods/lesson-cards/.claude-plugin/plugin.json and mods/lesson-cards/hooks/hooks.json). Content read still owed by reviewer-plugins-12; I have read provenance and gates only.

yonatangross:
HOLD read 2026-09-16 at 2f5c0baac558c7be00d39d5254bde601bc697218

The mod does not load. `claude plugin validate` refuses it, and one of its four source files does not parse at all. CI is green over that because nothing in CI looks at `mods/`.

The message-only rewrite does not affect this read: the root tree sha is `add5439943917e0d2cb08216017e930f38eedc70` at both 0ac0388d and 2f5c0baa, so every finding below was measured on identical content.

P1(1). `mods/lesson-cards/src/corpus.ts` is not valid TypeScript, and CC's own validator says so.

```
$ claude plugin validate mods/lesson-cards
Validating plugin manifest: .../.claude-plugin/plugin.json
Validating hooks: .../hooks/hooks.json

✘ Found 1 error:
  ❯ modules../register.ts: lesson-cards: .../src/corpus.ts does not parse:
    Expected ")" but found "three" (line 6, column 46) ...
✘ Validation failed
```

esbuild agrees independently, same file, same column:

```
src/corpus.ts          FAILS   Expected ")" but found "three"   corpus.ts:6:45
src/card.ts            PARSES
src/match.ts           PARSES
src/types.ts           PARSES
hooks/register.ts      PARSES
```

The cause is one character pair inside a doc comment. Line 6 is

```
 * - ~/.claude/hq/floor-*/lessons.md (newest three files)
```

and `floor-*/` contains `*/`, which closes the `/**` block opened on line 1. Everything after it on that line is then parsed as source: `lessons.md` reads as a member expression and `(newest three files)` as a call, which is where the error lands. A glob path in a JSDoc block is all it takes.

The consequence is not cosmetic. `hooks/hooks.json` declares `{"modules": ["./register.ts"]}`, `register.ts` imports `../src/corpus.js`, and the module graph therefore fails to load. This mod would ship as a plugin Claude Code cannot run.

P1(2). The tests that would have caught it do not import the code they are named after.

`tests/corpus.test.ts` imports nothing but `vitest`. It carries its own copy of `parseLessonsMd`, with a comment that says so in as many words:

```
// For now, let's inline the parser logic for testing
function parseLessonsMd(content: string): ... {
```

`tests/register.test.ts` imports only `type { $, MatchedLesson } from '../src/types.js'`, never `register.ts`. Of the three test files only `tests/match.test.ts` imports a shipped module (`../src/match.js`).

So "All 42 unit tests pass" is true and means less than it reads: 201 lines of `corpus.ts` and 144 of `register.ts` are covered by tests that never load them, which is exactly why a file that does not parse could sit behind a green suite. A test file that reimplements its subject also cannot detect drift between the copy and the original, which is the slower version of the same problem.

P1(3). `mods/` has no CI coverage at all, and that is what 18 green contexts mean here.

`.github/workflows/ci.yml` at this head contains no reference to `mods/`. Nothing typechecks it, nothing runs its vitest suite, nothing plugin-validates it. The only repo wiring this PR adds is one `start_audit` line in `tests/security/test-npm-audit.sh` and one `EXEMPT_ROOTS` entry in `bin/worktree-config.sh`. Both of those are correct and worth keeping; neither would notice a syntax error.

That is the structural finding, and it matters more than the typo: 3,196 lines entered the repo through a gate that does not look at them. The next mod lands the same way. `claude plugin validate` is free, offline and would have failed this PR in one second.

Bars for the LAND:

1. Fix line 6 so the comment does not self-terminate, and make `claude plugin validate mods/lesson-cards` exit clean. That command is the oracle, not the 18 contexts.
2. Have `tests/corpus.test.ts` import `../src/corpus.js` and delete the inlined parser; have `tests/register.test.ts` import `../hooks/register.js`. A test whose failure mode is invisible to the module under test is not coverage.
3. Wire `mods/` into CI, at minimum a plugin-validate step plus the mod's own vitest run, before a second mod arrives.

What is sound, and it is most of the PR:

The mod API is real, which was my first doubt and is worth recording since `mods/` does not exist on main and no loader for it lives in this repo. The CC 2.1.273 binary carries `"ui.render"`, `"command.register"`, `"session.start"`, `"tool.call"` and `"ui.invalidate"` as quoted strings, so `register.ts` is written against a surface that exists rather than an invented one.

`card.ts`, `match.ts`, `types.ts` and `register.ts` all parse. The `start_audit` addition means the new package is actually audited rather than silently exempt, the `EXEMPT_ROOTS` entry matches the pattern of the five siblings already there, and the root `typescript: ^7.0.2` devDependency is the same version `src/hooks` and `packages/cli` already declare, so it introduces no new major.

Lab artifacts are consistent: `docs/chore--dsp-2394-mod-lesson-cards/lesson-cards.html` against `docs/site/public/lab/lesson-cards-mod.html` differs in exactly the two generated blocks, and the manifest fragment and the added `lab-data.ts` row agree on the slug.

P3s:

The acceptance checklist in the body is eight unchecked boxes, including `Classic guard still denies block-severity patterns`, `Corpus load under 50ms` and `No I/O on hot path`. Those are the author's own criteria and none is claimed met.

The body still carries `Co-Authored-By: Yonatan Gross <yonaigross@gmail.com>`, which the commit rewrite removed from the commits but not from the PR body. The address is also a typo of the operator's real one, so it attributes a commit to an account that does not exist.

Four of the five commits carry no provenance trailers, the head commit `2f5c0baa` among them. Only `e04ca501` has the four keys.

auto-merge off. No long dash in the added lines, and no account identity in any commit trailer after the rewrite.

yonatangross:
LAND read 2026-09-16 at 973483246763f963d458694bb5dc18f7420e60ae, clearing the HOLD at 2f5c0baa

All three bars from that HOLD are met, and I re-ran the exact oracle that founded it. 18 of 18 required contexts success by name at this sha, 0 ambiguous under the failed-sibling-run rule.

Bar 1 MET. `claude plugin validate mods/lesson-cards` now exits clean, where at 2f5c0baa it refused with `src/corpus.ts does not parse`:

```
✔ Validation passed
  ./register.ts hooks: session.start, tool.call, ui.render, command.register
  ./register.ts calls: $.fs.list, $.fs.read, $.fs.stat, $.ui.invalidate, $.ui.notice
```

The parse failure is gone at its source. Line 6 of `src/corpus.ts` now reads `~/.claude/hq/floor-N/lessons.md`; the `floor-*/` glob whose `*/` closed the `/**` block opened on line 1 is no longer there. Fixing the comment rather than the parser is the right call.

Bar 2 MET. All three test files import the shipped modules:

```
tests/corpus.test.ts     ../src/corpus.js
tests/match.test.ts      ../src/match.js, ../src/corpus.js
tests/register.test.ts   ../hooks/register.js, ../src/types.js
```

The inlined `parseLessonsMd` copy is gone. The functions still defined inside the test files are genuine helpers, a fake `$` capability object, a hook capturer, a type coercion and a quoted-command wrapper, not reimplementations of the subject. That is the difference that matters: a helper that builds an input is fine, a copy that replaces the module under test is not.

Bar 3 MET, and the job is better than the bar asked for. `ci.yml` gains a `mods-validation` job that loops every directory under `mods/` and runs three gates on each: `claude plugin validate`, `npx tsc --noEmit`, and `npx vitest run`. It accumulates into a `status` variable rather than exiting on the first failure, so one broken mod does not hide the next one. The CLI is pinned to 2.1.273 with a stated reason (hooks registered on newer events only validate on newer CLIs, and users' installs auto-update, so the gate should match production rather than the 2.1.251 floor), installed with `--ignore-scripts`. The mod also now ships `package.json`, `tsconfig.json` and `vitest.config.ts`, so those two gates have something to run under.

The job ran at this head and concluded success (check run 104706343863).

P2. The new job is advisory, not required. `Mods` is not in the branch protection contexts list, so it cannot block a merge; today it passes, but a future mod that fails validation would land anyway with 18 of 18 still green. Given the whole reason this PR exists is that 18 green contexts sat over a file that would not parse, the job is worth promoting to required once it has a run or two of history.

P2. The loop fails open for a mod without a `package.json`:

```bash
if [ -f "$dir/package.json" ]; then
  ... tsc --noEmit ... vitest run ...
else
  echo "no package.json, skipping typecheck and tests"
fi
```

That branch is live right now, not hypothetical: #4165's `mods/promote-lights` ships no `package.json` and no vitest config, so under this job it would get `claude plugin validate` only, and its three test files would never run while the log says "skipping" and the job stays green. Either require a `package.json` per mod (a one-line refusal in the else branch) or make the absence itself a failure.

P3. The head commit still carries no provenance trailers, and the four non-feat commits in the series carry none either. Unchanged from my last read and still not blocking.

What I checked and did not need to re-check: the lab artifacts, the `start_audit` line and the `EXEMPT_ROOTS` entry were all sound at 2f5c0baa and are untouched here.

No long dash in the added lines, no account identity in any commit trailer. `automerge:hold` is still on this PR; on this LAND it can come off. auto-merge off.

PR #4170, opened 2026-09-15
fix(site): report reduced motion in host picker tests

### LINKED ISSUE #4155: docs-site tests: 7 unhandled AbortError rejections from the host-card entrance animation under happy-dom (main red at dbc05e28)
## What

The Docs Site workflow on main is red at dbc05e28 (run 35006486848, the merge of #4153): `Run docs-site tests` exits 1 with 64 files and 763 tests passed but **7 unhandled rejections**:

```
AbortError: The animation was canceled.
  Animation.cancel node_modules/happy-dom/lib/animation/Animation.js:161
  NativeAnimationExtended.cancel node_modules/motion-dom/dist/es/animation/NativeAnimation.mjs:73
  NativeAnimationExtended.stop node_modules/motion-dom/dist/es/animation/NativeAnimation.mjs:92
  AsyncMotionValueAnimation.stop ... MotionValue.stop ... frameloop batcher
```

Vitest attributes them to `__tests__/host-install-picker.test.tsx` (last test before the error: "copies two Codex lines as one clipboard payload"). The same job passed on the PR head 5c4d645e, so the failure is nondeterministic (a rerun of the failed job was requested to measure that).

## Cause, hypothesis

#4153 moved the host-card entrance from a prerendered `initial` variant to a post-mount `animate` (host-install.tsx, `mounted` state set in `useEffect`). Under happy-dom, motion-dom now starts real Web Animations after mount; when the test unmounts the tree mid-animation, `Animation.cancel()` rejects the animation's finished promise and nothing awaits it, so vitest reports an unhandled rejection. Whether the 7 appear depends on unmount timing versus the animation frame loop, which explains green on the PR head and red on main.

## Fix shape

Pick one, prove it with three consecutive green runs of `npx vitest run -c vitest.site.config.ts __tests__/host-install-picker.test.tsx` under load:

- In the picker tests, mock `matchMedia` to reduced motion (the component then passes `animate={undefined}` and starts no animation), or wrap renders in a `MotionConfig reducedMotion="always"` provider, keeping the one reduced-motion assertion that exists.
- Or in the component, start the entrance only when `!reduceMotion && mounted` AND the document is visible, and add an `onAnimationComplete`/cleanup that stops motion values before unmount.
- Or add a vitest `onUnhandledRejection` filter for `AbortError: The animation was canceled` from motion-dom (least preferred: it hides the class).

## Evidence

- Run 35006486848 (main, dbc05e28): Docs Site Build failure at step "Run docs-site tests"
- PR #4153 head 5c4d645e: Docs Site Build success
- Prod is unaffected: the site deploys through Vercel's git integration, not this workflow; https://orchestkit.yonyon.ai/ static HTML has 0 cards at opacity 0 as of 18:2xZ.

Related: #4153, #4150, #4136 (another timing-dependent test in a gate).

yonatangross:
## What

Test-only fix in `docs/site/__tests__/host-install-picker.test.tsx`: a `beforeEach` stubs `window.matchMedia` to report `prefers-reduced-motion` as reduce for every test in the file, so motion-dom never starts the card entrance that #4153 added. The stub is a new `reducedMotionMatchMedia` helper passed to `mockImplementation`; the one test that exists to assert reduced-motion rendering keeps its own explicit stub. `host-install.tsx` is untouched. No vitest `afterEach` for cleanup was added because `__tests__/setup.ts` already registers `afterEach(cleanup)` for every test.

## Why

GH-4155: the Docs Site workflow went red on main at dbc05e28 although all 763 tests passed, because vitest exited nonzero on seven unhandled rejections of `AbortError: The animation was canceled` attributed to `host-install-picker.test.tsx`. Under happy-dom, motion-dom starts real Web Animations for the post-mount entrance (per card: opacity, scale, y keyframes, 500 ms, 50 ms stagger); when the test ends, @testing-library unmounts the tree, motion cancels the animations mid-flight, and each cancelled `Animation.finished` promise rejects with nobody awaiting it. The job was green on the PR head, so the rejections were timing dependent.

Placement in `beforeEach` is load bearing, not cosmetic: motion initializes a module-global ref from `matchMedia` on the first `useReducedMotion()` call in the module graph and never re-reads it, so the stub must be installed before the first render of the file. `beforeEach` runs before the first test, which covers the whole file.

## How verified

`cd docs/site && npx vitest run -c vitest.site.config.ts __tests__/host-install-picker.test.tsx`, three consecutive runs:

```
 Test Files  1 passed (1)
      Tests  16 passed (16)
   Duration  590ms

 Test Files  1 passed (1)
      Tests  16 passed (16)
   Duration  735ms

 Test Files  1 passed (1)
      Tests  16 passed (16)
   Duration  644ms
```

All three runs show no `Errors` line; before the fix the same command printed `Errors 7 errors` under identical pass counts.

Full docs/site suite, one run:

```
 Test Files  64 passed (64)
      Tests  763 passed (763)
   Duration  10.52s
```

Repo gates: `npm run test:security` 20 of 20 passed, the docs-site drift test passed, and `node docs/site/scripts/lab-manifest.mjs --check` passed with 178 fragments matching lab-data.ts and public/lab copies. A full `npm test` run on this workstation passed 23 of 25 categories: the E2E failure was an environment precondition (missing `src/hooks` node_modules, passes after install) and the Performance category failed only on hook wall-clock latency (failure-handler measured 521 ms against a 500 ms target) on a loaded machine, with no hook, plugin or src path touched by this PR.

## Lab

https://orchestkit.yonyon.ai/lab/picker-reduced-motion.html

An interactive replay of the mechanism: mount the seven host cards, cancel the WAAPI entrance mid-flight, and see exactly when the AbortError rejections appear and why the reduced-motion stub prevents them.

Closes #4155

Harness: pi | Model: z-ai/glm-5.3-flash | Effort: mech | Seat: fix-4155

## Interactive Playground

docs/chore--dsp-97579-fix-4155/index.html

Shows the entrance animation with and without the reduced-motion stub, so the
rejection the test suite was hitting is visible rather than described. The
published copy is docs/site/public/lab/picker-reduced-motion.html, generated
from docs/site/lab-manifest/picker-reduced-motion.json.

## Summary by CodeRabbit

- **Tests**
  - Improved test stability by consistently honoring reduced-motion preferences, preventing animation-related failures and unhandled errors.
  - Preserved dedicated coverage for reduced-motion behavior.

- **Documentation**
  - Added a lab manifest entry documenting a test-run issue involving canceled animations, false failure signals, and the reduced-motion solution.
  - Included details on the affected test environment and validation results.


yonatangross:
LAND read 2026-09-16 at 4723ca813d0361796005e9e5d76e2bcdfd4a1599

18 of 18 required contexts success by name at this sha, test-only change, no source file touched. One P2 that is a coverage consequence worth a follow-up, and two P3s. I read motion's own source rather than take the mechanism on the body's word, and the body is right.

The mechanism claim checks out, exactly as written. `framer-motion/dist/es/utils/reduced-motion/use-reduced-motion.mjs:36` is

```
!hasReducedMotionListener.current && initPrefersReducedMotion()
const [shouldReduceMotion] = useState(prefersReducedMotion.current)
```

and both refs are module-level objects in motion-dom (`const prefersReducedMotion = { current: null }`, `const hasReducedMotionListener = { current: false }`, motion-dom.dev.js:6574). So the first `useReducedMotion()` call in the file's module graph latches the value for the whole file and no later `matchMedia` change can move it. `beforeEach` is therefore the correct and only placement, and the comment explaining it is accurate.

The stub matches the query motion actually issues, which is worth stating because it is not the obvious one. `initPrefersReducedMotion` calls `window.matchMedia("(prefers-reduced-motion)")`, without `: reduce`. The stub returns `matches: query.includes("reduce")`, and it is satisfied because "prefers-reduced-motion" contains "reduce" as a substring of "reduced". It works; it works by a hair.

Everything else in the fix is sound. `setup.ts:26` does register `afterEach(cleanup)`, so the body's reason for adding no `afterEach` is true. The diagnosis of the original failure matches the code: the entrance starts real WAAPI animations under happy-dom, unmount cancels them, and each cancelled `Animation.finished` rejects unobserved, which is why the job could pass 763 tests and still exit nonzero.

Lab artifacts are consistent, checked not trusted. `docs/chore--dsp-97579-fix-4155/index.html` against `docs/site/public/lab/picker-reduced-motion.html` differs in exactly the two generated blocks and nothing else. The fragment's slug is `picker-reduced-motion` while its `source` points at the `chore--dsp-97579-fix-4155` directory, which is a slug that does not mirror its directory, unlike the other Lab entries landing today; the manifest fragment, the `lab-data.ts` row and the published page all agree on it and the gate passed, so it is consistent, just unconventional. #4155 is open, so `Closes #4155` will close it.

P2, the file no longer exercises the motion-enabled path at all, and because of the latch it cannot.

Before this change 15 of the 16 tests ran with happy-dom's real `matchMedia`, that is with motion ON, and one test stubbed reduced motion explicitly. After it all 16 run reduced. The only assertion in the file about the entrance is `never prerenders cards at opacity 0 under reduced motion`; there is no counterpart asserting the cards end visible when motion is on. That is the failure #4153's entrance could actually ship to a user who has not asked for reduced motion, and it is now untested here.

This is not fixable inside this file. The latch means one module graph gets one answer, so a motion-on test has to live in a separate test file that never imports the reduced path first, or the rejection has to be handled rather than avoided. Follow-up, not a blocker: the pre-existing coverage was incidental rather than deliberate (no test asserted anything about the animation running), and trading an incidental path for a green suite is the right call today.

P3, the explicit stub inside the reduced-motion test is now dead code, and the comment says it is kept on purpose. By the time test 263 runs, `beforeEach` has already installed the same stub and an earlier test has already latched `prefersReducedMotion.current`. Its `vi.spyOn(...)` changes nothing, and its trailing `vi.restoreAllMocks()` cannot unlatch the ref either. The test now passes because of the `beforeEach`, not because of its own setup, which also means it is no longer independent of test order. Worth deleting the inner stub or keeping it with a comment that says it is belt-and-braces, rather than one that implies it is load bearing.

P3, `reducedMotionMatchMedia` returns `matches: true` for any query containing "reduce", including one that asks the opposite question. No such query exists in this code path today, so it costs nothing now; pinning the exact string motion queries would cost nothing either.

The branch is BEHIND main. If an update-branch click lands before the merge, this LAND needs a repin at the new head. No long dash in the added lines, auto-merge off, four provenance keys with no account identity.

yonatangross:
LAND read 2026-09-16 at 596061e853e7010ec26bfa6bfb47d9b7769b4f04, repin of the LAND at 4723ca81

18 of 18 required contexts success by name at this sha. The update-branch brought no content of its own, and the one file that did move is the generated Lab data, which I checked as a set rather than as a blob because a blob compare would have been misleading here.

Four of the five files are byte-identical at 4723ca81 and 596061e8:

```
docs/chore--dsp-97579-fix-4155/index.html                4ceefa9eeb  SAME
docs/site/__tests__/host-install-picker.test.tsx         fff0397129  SAME
docs/site/lab-manifest/picker-reduced-motion.json        0475a43e12  SAME
docs/site/public/lab/picker-reduced-motion.html          c05dd8d938  SAME
docs/site/lib/generated/lab-data.ts                      95b5996243 -> 5be181a166  MOVED
```

`lab-data.ts` moved because both sides added a row, which is the known conflict shape for that generated file, so the right check is the slug set and not the bytes. Comparing slug sets across the merge base (1de3f6a2), main and this head:

```
base   180 rows
main   181 rows   main minus base:  chore--dsp-99733-fix-4158   (#4168's landing)
head   181 rows   head minus base:  picker-reduced-motion       (this PR's row)
                  base minus head:  (empty)
```

The two additions are disjoint and nothing was dropped, so the merge resolved the generated file correctly and a merge into main lands 182 rows. The PR's own patch on that file is still a single added line, and the file list is still the same five files, +562/-0.

The head commit is the merge itself (parents 4723ca81 and 1de3f6a2, committed by GitHub), mergeable_state is now clean.

The findings from the base read stand unchanged at this sha, since the test file and both Lab pages are byte-identical: the `beforeEach` placement is correct against motion's own module-level latch, the stub matches the `(prefers-reduced-motion)` query motion actually issues, and the P2 remains that all 16 tests now run reduced so the motion-enabled path has no coverage in that file and cannot get any there.

No long dash in the added lines, no account identity in the commit trailer or body, auto-merge off.

PR #4171, opened 2026-09-15
fix: stop the pre-commit runner from chmod-mutating tracked scripts

### LINKED ISSUE #4084: tests/run-all-tests.sh chmod +x on every .sh during pre-commit dirties tracked file modes
Found by lane fix-4070 (2026-09-13): tests/run-all-tests.sh line 130 runs chmod +x on every .sh file during the pre-commit run and changed the mode of test-ascii-density-ratchet.sh, leaving a dirty tracked file the lane had to restore by hand. Ask: only chmod files the test run needs, or check modes and fail with a message instead of mutating the tree; add a test that the pre-commit run leaves git status clean.

yonatangross:
## What

`tests/run-all-tests.sh` opened every run, including the pre-commit `--lint` path, with two `find -exec chmod +x {} \;` passes over `tests/` and `src/hooks`. On a tracked script whose git mode was 100644, the chmod flipped the on-disk bit and `git status` picked up a mode change nobody asked for: the validator mutated the worktree it was validating. That is how `tests/test-ascii-density-ratchet.sh` got dirtied for lane fix-4070 (#4084).

- Both chmod passes are replaced with a read-only exec-bit preflight: one `git ls-files -s -- '*.sh'` call; a file is compliant when the index records 100755 AND the exec bit is set on disk; otherwise the run fails, naming each file and its repair command (`chmod +x <file> && git add <file>`). The runner never chmods. Outside a git work tree (tarball checkout) the check degrades to a skip, like the old courtesy pass.
- The seven tracked `.sh` files that sat at 100644 are fixed to 100755 in this PR, so the check is green on main: `5-deps-pr.sh`, `.claude/coordination/lib/coordination.sh`, `.claude/scripts/feedback-lib.sh`, `.claude/scripts/memory-lib.sh`, `.github/scripts/npm-audit-with-retry.sh`, `scripts/ci/check-skill-structure.sh`, `scripts/ci/version-skip-pattern.sh`.
- New `tests/ci/test-run-all-tests-no-mutation.sh` drives the pre-commit `--lint` invocation (the one `bin/git-hooks/pre-commit` step 7 makes, with the same GIT_* env scrub) in a throwaway git repo whose only tracked script is a copy of the runner: a clean fixture runs green with `git status --porcelain` empty afterwards; a stripped disk bit or an index mode of 100644 fails the run, names the file and the fix, and repairs nothing.

## Why

#4084. A validator that writes cannot be trusted: the mutation surfaced as a mode change riding in an unrelated lane's commit. The pre-commit path is the worst place for it, because it runs seconds before history is written and the author's `git add` picks up whatever the hook left behind. Failing loudly with the repair command beats silently repairing, because the repair is then an explicit, staged decision.

## How verified

- `bash tests/run-all-tests.sh --quick`: 20 categories passed, 0 failed, advisory warnings 263 under the 267 baseline
- `git status --porcelain` after the run: exactly the PR's own staged diff, nothing added by the suite
- `bash tests/ci/test-run-all-tests-no-mutation.sh`: 3 scenarios, PASS
- `node docs/site/scripts/lab-manifest.mjs --check`: 176 fragments == lab-data.ts == public/lab/
- `npx tsc --noEmit -p src/hooks/tsconfig.json`: clean (the pre-push gate)

## Lab

https://orchestkit.yonyon.ai/lab/chore--dsp-94218-fix-4084.html

Closes #4084

Harness: pi | Model: z-ai/glm-5.3-flash | Effort: mech | Seat: fix-4084

## Summary by CodeRabbit

- **Bug Fixes**
  - The test runner no longer changes file permissions during execution.
  - Added a read-only check to verify tracked shell scripts have the correct executable status.
  - Invalid permissions now stop the run with the affected file and a clear repair command.
  - Git status remains clean after successful test runs.

- **Tests**
  - Added coverage for clean repositories, missing executable permissions, staged mode changes, and protected Git environments.

- **Documentation**
  - Added a lab manifest documenting the updated test-runner behavior.

## Interactive Playground

docs/chore--dsp-94218-fix-4084/index.html

Shows the old mutating `find ... -exec chmod +x` passes beside the new
read-only preflight, and walks the seven tracked scripts that were at 100644
through both, so the difference between "validator fixed the tree" and
"validator reported the tree" is visible rather than described. The published
copy is docs/site/public/lab/chore--dsp-94218-fix-4084.html, generated from
docs/site/lab-manifest/chore--dsp-94218-fix-4084.json.


yonatangross:
HOLD at c556f67b60341175760719efa5fefe07f05acd41 (conductor-orchestkit-7). Two blockers, one of them is the PR's own bug committed into the PR.

P1, the fix did not hold on its own branch. Seven tracked scripts carry a bare mode flip 100644 to 100755 with zero content change:

```
.claude/coordination/lib/coordination.sh
.claude/scripts/feedback-lib.sh
.claude/scripts/memory-lib.sh
.github/scripts/npm-audit-with-retry.sh
5-deps-pr.sh
scripts/ci/check-skill-structure.sh
scripts/ci/version-skip-pattern.sh
```

Counted from the patch: 7 `old mode 100644` / 7 `new mode 100755` lines against 6 files with content hunks. This PR exists to stop the pre-commit runner from chmod-mutating tracked scripts, and it ships exactly that mutation. The body claims `git status --porcelain` was empty after `bash tests/run-all-tests.sh --quick`, which is consistent only if the mutation was already staged before the fix landed. Either way the evidence offered does not cover the defect that is visible in the diff. Drop the seven mode changes unless one of them is a deliberate, separately justified correction, and say which.

P2, the branch is DIRTY. `git merge-tree --write-tree origin/main HEAD` gives exactly one conflict, `docs/site/lib/generated/lab-data.ts`, which is a generated file and should be resolved by regeneration rather than by hand. Because the branch will not merge, only 1 check run exists at this head, so nothing about CI has been established here.

P3, the body has no `## Interactive Playground` section. The page ships at `docs/chore--dsp-94218-fix-4084/index.html`, so this is a body edit, not new work. Same shape as #4170.

Provenance is clean: four keys, no account identity.

Auto-merge stays null. A lane is being dispatched for the rebase plus the mode-change removal.

yonatangross:
Correction to my HOLD (5692621829): the P1 was wrong. Retracting it, and replacing it with a smaller but real finding.

RETRACTED P1. I called the seven 100644 to 100755 mode flips "the bug appearing inside its own fix". They are not. Reading the actual hunk in tests/run-all-tests.sh, this PR replaces two mutating `find ... -exec chmod +x` passes with a read-only, fail-closed preflight that aborts the run when any tracked `*.sh` is not mode 100755 with the exec bit set. Counted on origin/main over all tracked .sh files:

```
 495  100755
   7  100644
```

The seven at 100644 are exactly the seven this PR flips. So the flips are the one-time migration the new gate requires, not collateral from the old chmod. I judged the diff without reading the fix it belonged to, which is the same mistake my own HOLD accused the PR of.

STANDS, P2, the branch is DIRTY. `git merge-tree --write-tree origin/main HEAD` gives exactly one conflict, `docs/site/lib/generated/lab-data.ts`, a generated file that should be resolved by regeneration.

NEW P2. `5-deps-pr.sh` at the repo root is 0 bytes on main. The PR makes an empty file executable so it satisfies the new gate. That is the gate documenting a stray rather than the stray being cleaned up. Delete the file instead of chmod-ing it, unless it is a placeholder something depends on, in which case say what.

NEW P3, design, not a blocker. The gate's rule is "every tracked .sh is 100755", so it marks sourced libraries executable too. `.claude/coordination/lib/coordination.sh` is sourced from at least six call sites (`coord-decisions`, `coord-lock`, `coord-status`, and three examples) and is never invoked directly; the same applies to `feedback-lib.sh` and `memory-lib.sh`. Uniformity has its own value and 495 of 502 files already comply, so this is a note, not an objection. Worth one line in the preflight comment saying the rule is deliberately uniform rather than semantic.

STANDS, P3, the body has no `## Interactive Playground` section. The page ships at `docs/chore--dsp-94218-fix-4084/index.html`, so this is a body edit.

Provenance remains clean: four keys, no account identity.

yonatangross:
LAND read 2026-09-16 at 8ffae3dea2a1d739ab61e78d5cd595b90102995b

18 of 18 required contexts success by name at this sha, 0 ambiguous under the failed-sibling-run rule. This is the first content read this PR has had at any head, so I read it from scratch rather than against a prior verdict. One P3 I proved in a fixture, and one note about which oracle to trust on the deleted file.

The fix is real and the replacement is the right shape. Two `find ... -exec chmod +x {} \;` passes become one read-only, git-scoped preflight. I ran the shipped loop verbatim against this repo's tree, which still carries the 7 offenders, and it names exactly them and nothing else:

```
detected 7:
  .claude/coordination/lib/coordination.sh
  .claude/scripts/feedback-lib.sh
  .claude/scripts/memory-lib.sh
  .github/scripts/npm-audit-with-retry.sh
  5-deps-pr.sh
  scripts/ci/check-skill-structure.sh
  scripts/ci/version-skip-pattern.sh
```

The pathspec reaches every depth: 502 tracked `.sh` matched, not just the top level.

The migration is complete, and this is the check that mattered most, because the preflight is fail-closed on every invocation including the pre-commit `--lint` path. One straggler at 100644 would abort every contributor's commit. At this head, of 502 tracked `.sh` files, all 502 are mode 100755 and none is anything else. On main exactly 7 are not, and they are exactly the 7 this PR resolves, 6 by mode flip and one by deletion. So the gate goes green the moment this lands and stays green.

On the seven mode flips: they are the migration the new gate requires, not the bug it fixes. I confirmed that independently rather than taking the retraction on trust, by checking the mode distribution at both ends. The uniform 100755 rule is applied to sourced libraries too, which the comment calls out as deliberate rather than semantic. That is the right way to state a convention you are choosing.

Deleting `5-deps-pr.sh` instead of chmod-ing it is correct: it was 0 bytes on main and is absent at this head. Marking an empty file executable to satisfy a gate would have been the wrong kind of compliance.

Which oracle to trust here, because the obvious one misleads. The cumulative `gh pr diff` still renders `old mode 100644 / new mode 100755` for `5-deps-pr.sh`, since the first commit flips the mode and the second removes the file. The per-commit file lists show that plainly (`modified` then `removed`), the cumulative API file list already says `removed`, and the contents API returns 404 at the head. Three oracles agree; only the rendered patch reads as though a mode change ships.

The new test asserts the property, not the output. `tests/ci/test-run-all-tests-no-mutation.sh` drives the same `--lint` invocation the pre-commit hook makes, in a throwaway repo, and its three scenarios check the things that would actually regress: a clean fixture exits 0 with `git status --porcelain` empty, a stripped disk bit exits non-zero and the file is STILL non-executable afterwards, and an index mode of 100644 exits non-zero with the index mode UNCHANGED. Those last two are the assertions that matter, because they prove the runner reported rather than repaired.

It also scrubs the environment, which I was glad to see: `env -u GIT_INDEX_FILE -u GIT_DIR -u GIT_WORK_TREE -u GIT_PREFIX -u GIT_COMMON_DIR -u GIT_OBJECT_DIRECTORY`. A fixture repo driven from inside a pre-commit hook without that scrub aims its `git add` at the real index, which is a failure this estate has already paid for once.

P3, and I proved it rather than asserting it. The preflight is blind to a non-ASCII path, and it fails open there. `core.quotePath` defaults to true, so `git ls-files -s` prints such a path C-quoted and octal-escaped; `[ -f "$path" ]` then fails on the quoted spelling and the loop `continue`s. In a fixture with two non-compliant scripts, one ASCII and one Hebrew-named:

```
default quotePath      detected 1 of 2:  src/plain.sh
core.quotePath=false   detected 2 of 2:  src/plain.sh, src/שלום.sh
```

The miss is silent and lands on the `[ -f "$path" ] || continue` line whose comment reads "staged deletion: no mode to fix", so a skipped file is indistinguishable from an intended skip. Not live today: of 7617 tracked paths in this repo, zero are non-ASCII, and zero tracked `.sh` are. But this is a durable guard in an estate that writes Hebrew, and the fix is one flag, `git -c core.quotePath=false ls-files -s`, or `-z` with a NUL-delimited read.

A method note on that P3, since it nearly went out wrong. My first pass used `awk '/[^\x00-\x7F]/'`, which reported all 7617 paths and all 502 scripts as non-ASCII. awk does not expand `\x` escapes inside a bracket expression, so the class matched nearly everything. The numbers above come from a byte-level check in python instead.

The degradation path is honest in both directions: outside a git work tree the check skips entirely, matching the old courtesy pass, and a git failure yields an empty list and the run proceeds unguarded, which the inline comment states rather than hides.

No long dash in the added lines, no account identity in the commit trailers or body, auto-merge off. `Closes #4084`, and #4084 is open.

Issue #4173, opened 2026-09-16
Release PRs are permanently blocked: rebuild guard refuses on .devin-plugin, warns instead of failing


yonatangross:
## Summary

Every release PR is permanently blocked. `release-please.yml`'s hook-bundle rebuild refuses to run, reports success anyway, and `ci.yml` then fails the release PR with an error message that prescribes the one remedy that cannot work.

Hit on #4166 (10.0.0-beta.34). It is not specific to that release: the refusing condition is structural and will recur on every future one.

## The chain

**1. `.devin-plugin/plugin.json` is a version-stamped extra-file.**
`.release-please-config.json` lines 69 to 71 register it with `jsonpath: $.version`, so release-please writes the new version into it on the release branch. Measured: main holds `10.0.0-beta.33`, the release branch holds `10.0.0-beta.34`. That difference is correct and intended.

**2. The rebuild guard refuses because of exactly that file.**
`release-please.yml`'s "Rebuild hook bundles on the release PR" step will not build a release branch that differs from main outside an allowlist. From run 35062658242 at 06:13:04Z:

```
##[warning]Release branch release-please--branches--main differs from main
outside release-owned paths; refusing to execute its build:
.devin-plugin/plugin.json
```

**3. The allowlist does not contain it.**
The `ALLOWED` regex at `release-please.yml:318` covers CHANGELOG, the manifest, version.txt, package files, the marketplace and several `plugin.json` variants, generated docs data, and both `dist/` dirs. It has zero occurrences of `devin`. So the guard refuses on every run, not occasionally.

**4. The refusal is a warning, so the job reports success.**
`release-please.yml:330` and `:348` both emit `::warning` rather than failing. Every release-please run today reads `completed/success` in `gh run list` while the bundles were never rebuilt. The failure is invisible at the only place anyone looks.

**5. `ci.yml` then blocks the release PR.**
"Check hook bundles are release-owned (#3578)" compares a fresh build against the committed bundles on the release branch. Since #4168 changed hook source and the bundles were never rebuilt, they differ:

```
 plugins/ork/hooks/dist/bundle-stats.json   8 ++--
 plugins/ork/hooks/dist/subagent.mjs       82 ++++++-----
 src/hooks/dist/bundle-stats.json           8 ++--
 src/hooks/dist/subagent.mjs               82 ++++++-----
 4 files changed, 90 insertions(+), 90 deletions(-)
```

CI Summary aggregates it and the release PR cannot go green.

[...]

PR #4164, opened 2026-09-15
ci(playground): treat generated Lab artifacts as inert (refs #4161, #3811, #3845, #3993)

yonatangross:
Operator merge: workflow file.

## What

Adds three paths to the PR Playground gate INERT list in .github/workflows/ci.yml: the generated Lab page copy (docs/site/public/lab/), the manifest fragment (docs/site/lab-manifest/) and the generated data module (docs/site/lib/generated/lab-data.ts). A comment block documents the case in the same voice as the three widenings before it. No other file is touched.

## Why

Fourth instance of the page about a page circularity this gate keeps hitting. The gate already treats a diff confined to the Lab itself as maintenance (#3811 docs/playgrounds, #3845 docs/audits, #3993 .claude/rules), but the outputs of docs/site/scripts/lab-manifest.mjs live outside that pattern. PR #4161, a fix to one Lab page plus its generated copy under docs/site/public/lab/, failed the gate with "Missing PR playground directory" even though its source page was already inert. A diff confined to generated Lab artifacts plus their source page is Lab maintenance, and demanding a playground for it is a page about a page.

## Verified

actionlint: two SC2001 style findings, both pre-existing on HEAD (verified by running actionlint on the HEAD version of the file; line numbers shifted only). No new findings from this change.

Local regex proof, INERT extracted from the committed workflow, gate semantics mirrored:

```
$ printf '%s\n' 'docs/site/public/lab/sample-page.html' \
    'docs/site/lab-manifest/sample-page.json' \
    'docs/site/lib/generated/lab-data.ts' \
    'docs/site/app/x.tsx' | grep -vE "$INERT"
docs/site/app/x.tsx

$ for f in docs/site/public/lab/sample-page.html docs/site/lab-manifest/sample-page.json docs/site/lib/generated/lab-data.ts docs/site/app/x.tsx; do
>   grep -qE "$INERT" <<< "$f" && echo "inert:      $f" || echo "NON-INERT:  $f"; done
inert:      docs/site/public/lab/sample-page.html
inert:      docs/site/lab-manifest/sample-page.json
inert:      docs/site/lib/generated/lab-data.ts
NON-INERT:  docs/site/app/x.tsx
```

Only the last path is non-inert. The shipped gate test (tests/ci/test-playground-gate-inertness.sh) passes: 23 passed, 0 failed.

After merge: rerun the failed PR Playground job on #4161.

Harness: pi | Model: z-ai/glm-5.3-flash | Effort: mech | Seat: hqd-0aad42dc

PR #4161, opened 2026-09-15
fix(lab): exact URL matches in yonyon crosslink checker (refs #4144)

yonatangross:
## What
The yonyon crosslink checker in docs/chore--dsp-94242-fix-4144/index.html now uses exact equality over an array instead of indexOf on URL strings:

* var list = Array.isArray(node.sameAs) ? node.sameAs : [];
* list.some(function (u) { return u === apex; }) for apex, apexPage and docsPage
* No indexOf or includes on URL strings remains anywhere in the page

The built copy docs/site/public/lab/chore--dsp-94242-fix-4144.html was regenerated from the source page with the lab manifest script. The commit title was shortened to satisfy the repo 72 character hook limit; the full prescribed wording sits in the commit body.

## Why
CodeQL reported js/incomplete-url-substring-sanitization (high) on the built copy at lines 158 and 159. Substring matching on URLs can accept lookalike strings that merely contain the apex as a substring, while the rule is meant to be an exact identity comparison.

## How verified
* node docs/site/scripts/lab-manifest.mjs (regenerated 178 lab entries)
* node docs/site/scripts/lab-manifest.mjs --check (set OK, 178 fragments == lab-data.ts == public/lab)
* node tests/skills/structure/test-docs-site-drift.mjs (SUCCESS)
* Behavioral check of the new rule logic in node: the before node fails url and sameas, the after node passes both, and substring lookalike urls fail the sameas rule instead of passing
* Pre push validation: TypeScript check, build, imports, unit tests (71 passed), security tests (20 passed), mdx compile, agent frontmatter, skill structure, manifests, token budget all OK
* npm test --quick was started locally but exceeded the session time budget before finishing; CI is the gate for the full suite

## Lab
https://orchestkit.yonyon.ai/lab/chore--dsp-94242-fix-4144.html

Refs #4144

Harness: pi | Model: z-ai/glm-5.3-flash | Effort: mech | Seat: fix-4144-b

PR #4178, opened 2026-09-16
fix(ci): allow .devin-plugin/plugin.json in release rebuild guard

yonatangross:
Operator merge: workflow file

## Why

`.devin-plugin/plugin.json` is a release-please extra-files entry
(`.release-please-config.json`, jsonpath `$.version`), so release-please
stamps the new version into it on the release branch. main is at
10.0.0-beta.33 and the release branch at 10.0.0-beta.34, a legitimate
difference. The rebuild guard in the "Rebuild hook bundles on the
release PR" step refuses to build a release branch that differs from
main outside ALLOWED, and that path was not in ALLOWED, so it refused
every time. From release-please run 35062658242 at 06:13:04Z:

```
##[warning]Release branch release-please--branches--main differs from main
outside release-owned paths; refusing to execute its build:
.devin-plugin/plugin.json
```

Because the refusal is a `::warning`, the job still reports success, so
the hook bundles are never rebuilt and ci.yml's "Check hook bundles are
release-owned" gate then blocks every release PR. #4166 (beta.34) is
blocked on it now.

## The change

One line in `.github/workflows/release-please.yml`: the ALLOWED
alternation gains `\.devin-plugin/plugin\.json`, placed next to the
other plugin.json entries, with the same escaped dots as its
neighbours. Nothing else in the file changed.

## Verification

1. The ALLOWED regex was extracted from the edited file and run in
   python (`re.fullmatch`) against the required paths:

```
PASS  match     want=match     .devin-plugin/plugin.json
PASS  match     want=match     CHANGELOG.md
PASS  match     want=match     src/hooks/dist/subagent.mjs
PASS  no-match  want=no-match  src/hooks/src/subagent-stop/foo.ts
PASS  no-match  want=no-match  README-notes.md
all 5 assertions passed
```

   The release stamp matches after the change; CHANGELOG.md and the
   hook bundles keep matching; a hook source and a lookalike README
   name still fail, so the guard keeps catching genuine drift.

2. `git diff --stat` shows `.github/workflows/release-please.yml` with
   1 insertion and 1 deletion.

3. Lab manifest: generated with the repo script, check mode passes
   (set check OK, 183 fragments == lab-data.ts == public/lab).

## Interactive Playground

docs/chore-dsp-39782-4160-allowed-devin/index.html

The page shows the ALLOWED regex with and without the entry, runs the
five verification paths through both versions live, and has a text box
for any path so the guard verdict before versus after is inspectable
rather than described.

Refs #4173

## Summary by CodeRabbit

- **Bug Fixes**
  - Release branch builds now permit expected changes to the `.devin-plugin/plugin.json` manifest, preventing valid release updates from being flagged as unexpected drift.

- **Documentation**
  - Added an interactive lab page explaining and demonstrating release rebuild validation.
  - Added metadata for the new documentation page, including release and CI topics.

## Interactive Playground

docs/chore-dsp-39782-4160-allowed-devin/index.html

Runs candidate paths through the ALLOWED regex with and without the new entry,
so you can see the guard refuse `.devin-plugin/plugin.json` before the change
and accept it after, while `src/hooks/src/subagent-stop/foo.ts` stays refused.
The published copy is
docs/site/public/lab/chore-dsp-39782-4160-allowed-devin.html, generated from
docs/site/lab-manifest/chore-dsp-39782-4160-allowed-devin.json.

Independently verified by conductor-orchestkit-7 at head 8adf46ba, running the
extracted regex through `grep -E` (the same engine the step uses) rather than
reading it:

```
  .devin-plugin/plugin.json             -> MATCH      (want MATCH)
  CHANGELOG.md                          -> MATCH      (want MATCH)
  src/hooks/dist/subagent.mjs           -> MATCH      (want MATCH)
  plugins/ork/hooks/dist/subagent.mjs   -> MATCH      (want MATCH)
  src/hooks/src/subagent-stop/foo.ts    -> no match   (want no match)
  README-notes.md                       -> no match   (want no match)
  Xdevin-plugin/plugin.json             -> no match   (want no match)
```

The last case is the one that matters for the escaping: it confirms the dot in
`\.devin-plugin` is literal and the alternation stays anchored, so the guard
still catches genuine drift.


yonatangross:
HOLD read 2026-09-16 at 8adf46ba202f49bb9ca2424b0d452ea3dfe43aae

The one-line fix is correct and complete, and I verified it against the live release branch rather than against the body's claim. The hold is a single missing hyphen in the playground directory name, which is both the only reason the gate fires and the reason it cannot be satisfied. A body edit will not clear it.

The fix itself is right, checked against real data.

Extracting the ALLOWED regex from the edited file and running it over the 22 paths that actually differ between `main` and `release-please--branches--main` right now:

```
main (today)   outside ALLOWED: ['.devin-plugin/plugin.json']
PR head        outside ALLOWED: NONE
```

So it is not a partial fix. That single path was the whole refusal, and after this change nothing on the live release branch sits outside the allowlist. The diagnosis in the body matches run 35062658242's warning at 06:13:04Z, and the `::warning` plus success exit is exactly why nobody noticed: the rebuild step reported green while refusing to build, and ci.yml's release-owned gate then blocked #4166.

The widening is one literal path and nothing more. Negative controls against the new regex:

```
REFUSE  src/hooks/src/subagent-stop/foo.ts
REFUSE  README-notes.md
REFUSE  .devin-plugin/other.json
REFUSE  docs/chore-dsp-39782-4160-allowed-devin/index.html
ALLOW   src/hooks/dist/subagent.mjs
```

Real source drift is still refused, a sibling file in the same directory is still refused, and even this PR's own new Lab page would be refused on the release branch. The entry is placed next to the other plugin.json entries with the same escaped dots, and the file is otherwise byte-identical.

P1, and it is a rename. `PR Playground` is `completed/failure` at this sha (check run 104695020614, step 4 `Check playground exists`):

```
Missing PR playground directory: docs/chore--dsp-39782-4160-allowed-devin/
```

Two hyphens after `chore`. The branch is `chore/dsp-39782-4160-allowed-devin` and the gate derives the slug by replacing `/` with `--`, so it wants `chore--dsp-...`. This PR shipped `docs/chore-dsp-39782-4160-allowed-devin/` with one hyphen, and the manifest fragment and published copy carry the same single-hyphen slug.

The same hyphen is why the gate fires at all, which is worth seeing because it means the rename fixes both halves at once. Running main's INERT alternation over this PR's five paths:

```
REQUIRES  docs/chore-dsp-39782-4160-allowed-devin/index.html
inert     docs/chore--dsp-39782-4160-allowed-devin/index.html
inert     .github/workflows/release-please.yml
inert     docs/site/lab-manifest/chore-dsp-39782-4160-allowed-devin.json
inert     docs/site/public/lab/chore-dsp-39782-4160-allowed-devin.html
inert     docs/site/lib/generated/lab-data.ts
```

`docs/[^/]*--[^/]*/.*` needs the double hyphen, so the single-hyphen directory is the only non-inert path in the entire PR. Every other file, the workflow and all three Lab artifacts, is already inert thanks to #4164. Name the directory correctly and the gate decides no playground is required and skips steps 4 through 6 outright, the same inert path #4161 took this morning.

Bar: rename `docs/chore-dsp-39782-4160-allowed-devin/` to `docs/chore--dsp-39782-4160-allowed-devin/`, rename the fragment and the published copy to match, and regenerate `lab-data.ts` with `node docs/site/scripts/lab-manifest.mjs`.

One correction to the routing note that came with this PR: adding a Playground section to the body does not fix it. Step 6, `Check playground link in PR body`, is `skipped` here because step 4 already failed, and the body already contains the word `playground` once. The failure is the directory name, not the body.

The other 17 required contexts are success by name at this sha, with no second failed run under any of them.

No long dash in the added lines, no account identity in the commit trailer or body, auto-merge off. This is a workflow file, so it stays an operator-gated merge even once the gate is green.

yonatangross:
HOLD read 2026-09-16 at 818c942a40ea11e466366418fc2574bb6d5c1397

The rename is done correctly and the fix is unchanged, so the content bar from my last read is cleared. What blocks it now is different and mechanical: the branch conflicts with main, so no CI has run at this head and nothing can merge.

The rename is right, all three artifacts moved together:

```
docs/chore--dsp-39782-4160-allowed-devin/index.html          added
docs/site/lab-manifest/chore--dsp-39782-4160-allowed-devin.json  added
docs/site/public/lab/chore--dsp-39782-4160-allowed-devin.html    added
docs/site/lib/generated/lab-data.ts                          modified +1/-0
```

Double hyphen everywhere, and the single-hyphen directory is gone. That means the only non-inert path in the PR is gone with it, so I still expect `PR Playground` to SKIP rather than pass once it runs, which is the outcome to check for rather than a green tick.

The fix itself is byte-identical to the head I already verified. `.github/workflows/release-please.yml` is blob `5a4239c3aa68504aae0242ba29fd32a889818d06` at both 8adf46ba and this head, against `16ba6cd2` on main. So the one-line ALLOWED widening, and the completeness proof I ran against the 22 live release-branch paths, both carry over without re-reading. No hook bundles are in the PR, so the feature-PR arm of the release-owned check has nothing to catch.

The blocker. `mergeable=false`, `mergeable_state=dirty`, held across six polls over about a hundred seconds, so this is a real conflict and not GitHub still computing. All 18 required contexts are MISSING at this sha (complete=0, pending=18) because GitHub builds no merge ref for a conflicting PR. The only check-run present is `Enable auto-merge for Dependabot`, skipped.

The cause is the generated Lab data. `compare main...818c942a` reads `diverged, ahead 1, behind 1`: main moved to 8494ef0a when #4171 landed a few minutes ago, and #4171 also adds a row to `docs/site/lib/generated/lab-data.ts`. Two branches appending a row to the same generated file is the known collision for that path. Take either side and regenerate with `node docs/site/scripts/lab-manifest.mjs`; the `--check` in CI is the oracle for the resulting set.

One correction to the note that came with this head, because it matters for what to do next. This is not "the rename plus a merge from main": commit 818c942a has a single parent, 9d659214, and carries the original `fix(ci): allow .devin-plugin/plugin.json in release rebuild guard` subject. There is no merge commit on the branch, and 9d659214 is now two commits behind main. So an update-branch or rebase is still owed, not already done.

Bar: resolve against main, regenerate `lab-data.ts`, let CI run. I will repin on a blob compare of the workflow file plus the 18 by name with the failed-sibling-run rule, and I will treat a SKIPPED PR Playground as the correct outcome rather than a missing one.

No long dash in the added lines, auto-merge off. Workflow file, so operator-gated even once green.

yonatangross:
LAND read 2026-09-16 at e9a345fb9c3e4a65769909eac590ee3e8cd164fb, operator merge (workflow file)

18 of 18 required contexts success by name at this sha, 0 ambiguous under the failed-sibling-run rule. The rebase resolved cleanly, the fix is the one I already verified, and the gate behaved the way the rename predicted.

The prediction held, which is the point of the rename rather than a side effect. `PR Playground` did not pass, it skipped:

```
3  success  Decide whether a playground is required
4  skipped  Check playground exists
5  skipped  Check playground is published to the Lab
6  skipped  Check playground link in PR body
```

With the directory correctly named `docs/chore--dsp-39782-4160-allowed-devin`, every path in the PR is inert and the gate declines to run at all. That is the second PR to take the inert path today, after #4161, and it is the correct outcome to see rather than a green tick on step 4.

The fix is unchanged from the head I verified earlier. `.github/workflows/release-please.yml` is blob `5a4239c3aa68504aae0242ba29fd32a889818d06` at 8adf46ba, at 818c942a and at this head, against `16ba6cd2` on main. So the one-line ALLOWED widening and the completeness proof I ran against the 22 paths actually differing between main and the release branch both carry over without re-deriving: the old regex left exactly `.devin-plugin/plugin.json` outside, the new one leaves nothing outside, and real source drift, a sibling `.devin-plugin/other.json` and this PR's own Lab page are all still refused.

The generated-file merge is the part worth checking rather than assuming, because a hand-resolved aggregate drops a row silently and nothing goes red. `docs/site/lib/generated/lab-data.ts` goes from 183 rows on main to 184 at this head, exactly one more, and all three recent slugs survive:

```
chore--dsp-94218-fix-4084             present   (#4171's, from main)
picker-reduced-motion                 present   (#4170's, from main)
chore--dsp-39782-4160-allowed-devin   present   (this PR's)
```

A count that is plus one with nothing missing is what regeneration looks like; a hand merge that took one side would have shown 183 or dropped a slug.

The head is a real merge commit this time (parents 818c942a and 8494ef0a), `mergeable` is true, and the file list is still exactly the five expected paths. No `hooks/dist` path is in the PR, so the feature-PR arm of the release-owned check has nothing to catch, which matters here because tripping it would have made this PR red on the very gate family it exists to unblock.

One process note on the sha. I pinned this verdict to the head I read from the API myself in the same call that produced the check state, not to a relayed string. A full sha for this PR that was circulated earlier (`e9a345fb...72ea1b4`) does not resolve; the real one is above. Worth saying out loud because under automerge a LAND line carries a sha straight into a merge command, so a fabricated one fails closed at best.

No long dash in the added lines, no account identity in the commit trailer or body, auto-merge off. This is a workflow file, so it stays an operator-gated merge.

Issue #4159, opened 2026-09-15
ork:glyph renders squeezed and unreadable in non-terminal hosts (Devin desktop): no width cap, double invocation, repeated thoughts


yonatangross:
Observed 2026-09-15 22:0x local in the Devin desktop app (Devin Local Session tab, core repo), operator screenshot at ~/.claude/image-cache/51b75868-4a51-4f93-86ec-acddb050b83c/9.png. Operator verdict: "horrible glyph in devin desktop, why so squeezed".

What the screenshot shows

1. The glyph block is a monospace card of about 100 to 110 columns rendered in a chat pane roughly 700 px wide, so every row is dense to the edge and the two-column key/value layout (deal, pending, feat #5, stack, core) reads as one squeezed wall. The DECIDE block below it has the same shape with ( ) and [ ] rows plus a right-aligned "<- recommended" that only works at terminal width.
2. Double invocation: the transcript shows "Invoked skill glyph" followed by "Invoked skill ork:glyph". The top-level glyph front door delegated to ork:glyph, which then rendered again. One render is expected.
3. The Thoughts section repeats the same two paragraphs three times ("I'm trying to fit the corrected Shai picture into a tight 12-line visual..."), which reads as the skill re-entering its own planning step rather than a host display bug; worth checking whether the re-invocation caused it.

Bar

- Width: glyph output for a non-TTY host (Devin, VS Code chat, web) caps at 72 columns and prefers one narrow table or a vertical list over side-by-side columns; the 12-line budget stays. Detect the host from the environment the way the skill already infers audience and surface (the glyph front door's own description says it picks chat, ask or page).
- Single render: the glyph wrapper must hand off to ork:glyph OR render itself, never both.
- Thoughts repeated three times: reproduce with the same prompt shape in a Devin session and fix the re-entry.
- Add a fixture test that renders a 5-row key/value glyph at 72 columns and asserts no line exceeds it.

Not verified by the filer: whether the repeated Thoughts is Devin's transcript rendering or the skill; only the screenshot was read.

Harness: cc | Model: claude-fable-5-1 | Effort: md | Seat: super-conductor-13

PR #4179, opened 2026-09-16
fix(glyph): cap non-TTY renders at 72 columns

yonatangross:
## Summary

Defect 1 of #4159 (width): the glyph skill taught side-by-side key/value columns with no host awareness, so a 5-row status board renders at 113 columns and wraps into a wall in every non-terminal host: CI logs, chat widgets, VS Code chat, web transcripts, agent desktops.

**Why the shape of the fix differs from the issue text.** The issue says "cap output at 72 columns", which reads like a literal cap on something in the skill file. It is not. The skill file's own inline blocks are all at or under 71 columns (measured across every fenced block in src/skills/glyph/SKILL.md); the width comes from what the model renders when it follows the side-by-side guidance ("comparison table or side-by-side boxes", the key/value board pattern), which lands at 100 to 113 columns for a realistic 5-row status fixture. So the fix is guidance plus an enforceable check on rendered output, not an edit to a wide string:

- src/skills/glyph/SKILL.md gains rule 8: read the surface from the environment the same way the skill already infers audience and surface. With a terminal (TTY), render as before. Without one, cap every diagram line at 72 columns and render key/value or comparison data as one narrow table or a vertical list, never side-by-side columns. The 12-line budget is unchanged and TTY behaviour is unchanged.
- scripts/render-ascii.sh gains a key-value pattern that reads the same surface from stdout: a TTY gets the side-by-side board as before; a pipe or file gets one narrow vertical list capped at --width (default 72) with continuation lines for long values. --mode wide|narrow|auto forces a layout.

plugins/ork/skills/glyph/SKILL.md, the generated Cursor command, and the docs-site skill data are the npm run build output for the src change.

## Fixture test: red before, green after

tests/unit/test-glyph-non-tty-width.sh renders tests/fixtures/ascii/kv-glyph-5row.txt (5 key|value rows) through the render path in non-TTY mode at a 72-column cap and asserts no output line exceeds 72 display columns. Cells are counted by tests/fixtures/ascii/display-width-oracle.py (east_asian_width W/F plus the closed status-glyph vocabulary, so emoji count as 2 columns).

Red at the test commit (ea5120ef3), the render path laid out exactly as the skill teaches:

```
rendered lines: 6, max display width: 113 (cap 72)
✗ line is 113 columns, over the 72-column cap: ┌───────────────────────────────────────────────────────────────┬───────────────────────────────────────────────────────────────┐
✗ line is 113 columns, over the 72-column cap: │  ingest          ✅ healthy, 0 retries                │  reindex         ⏸ waiting, scheduled tonight        │
✗ line is 113 columns, over the 72-column cap: ├───────────────────────────────────────────────────────────────┼───────────────────────────────────────────────────────────────┤
✗ line is 113 columns, over the 72-column cap: │  thumbnailer     ✅ healthy, 3 retries                │  api             🟡 degraded, p95 2.1s                │
✗ line is 113 columns, over the 72-column cap: │  webhook-fanout  🔴 failing, 41 retries, DLQ filling  │                                                       │
✗ line is 113 columns, over the 72-column cap: └───────────────────────────────────────────────────────────────┴───────────────────────────────────────────────────────────────┘
✗ key/value glyph in non-TTY mode exceeds the 72-column cap (GH-4159 defect 1)
exit=1
```

Green at HEAD (b7d3864d0):

```
rendered lines: 5, max display width: 51 (cap 72)
✓ non-TTY key/value glyph stays within the 72-column cap
exit=0
```

## Defects 2 and 3 of #4159

Both are NOT REPRODUCED in this repository. Evidence, gathered before any fix attempt:

- Double invocation ("Invoked skill glyph" then "Invoked skill ork:glyph"): there is no delegation site on main. src/skills/glyph/SKILL.md line 12 declares `allowed-tools: [Read, Grep, Glob]`, which excludes the Skill tool, so the skill body cannot invoke anything, including itself. The body contains no Skill, Task, or Agent call site (step 1 explicitly forbids spawning agents). The only invoke-shaped references to glyph elsewhere are one-way dependency listings (`skills:` frontmatter in visualize-plan and explore) and a subagent prompt line (src/skills/visualize-plan/SKILL.md:210) that tells a subagent to use glyph; none of them is a front door that also renders.
- Repeated Thoughts: with defect 2 unexplained in-repo, there is no planning step the skill could re-enter (step 1: render immediately, do not ask). The transcript's repeated paragraphs and its paired invocation labels are consistent with a host rendering one invocation under both its plain name (glyph, the frontmatter name) and its plugin-namespaced id (ork:glyph). That happens outside this repo's source and could not be observed here.

## What this PR does NOT fix

Defects 2 and 3 of #4159. Both are NOT REPRODUCED: no delegation site exists in this repo for the front door to double-fire from, and no planning step exists for the skill to re-enter, so there is nothing here to change without inventing a fix for an unobserved symptom. The issue stays open carrying them; if the filer can share the host and transcript, the labels ("glyph" vs "ork:glyph") point at host-side skill registration rather than at this repo.

## Interactive Playground

docs/chore--dsp-37524-fix-4159/index.html

The page shows the same 5-row key/value glyph three ways so the squeeze is visible rather than described: the 113-column terminal board, the same bytes wrapped at 72 columns the way a non-TTY host wraps them, and the fixed narrow vertical list at 51 columns max. Both renders are the actual output of scripts/render-ascii.sh key-value on the fixture, with a reproduce block per command.

## Test plan

- [x] bash tests/unit/test-glyph-non-tty-width.sh: red at 113 before the fix, green at 51 after (runs pasted above)
- [x] node docs/site/scripts/lab-manifest.mjs --check: set check OK, 181 fragments == lab-data.ts == public/lab/
- [x] bash tests/ci/test-no-bare-mktemp.sh, tests/ci/test-fault-arms.sh (12 of 12 probes), tests/unit/test-ascii-snapshots.sh (20 pass), test-ascii-i18n.sh, test-ascii-tokens.sh
- [x] bash tests/skills/structure/test-skill-md.sh, tests/skills/test-skill-length.sh, bin/validate-counts.sh
- [x] npm run build: plugins/ and docs-site skill data regenerated and staged with the src change

Refs #4159

## Summary by CodeRabbit

- **New Features**
  - Added a key/value glyph rendering pattern with wide and narrow layouts.
  - Automatically adapts rendering to the host surface: side-by-side output in TTY terminals and a narrow vertical list in non-TTY surfaces.
  - Non-TTY diagram lines are capped at 72 columns while preserving the existing 12-line budget.

- **Documentation**
  - Updated glyph rendering guidance and examples to describe TTY and non-TTY layouts.

- **Bug Fixes**
  - Prevented key/value and comparison diagrams from wrapping poorly in non-TTY environments.


yonatangross:
HOLD read 2026-09-16 at 1207558b5d90d61320235984e472cc32c96d7d11

All three bars are met, and the fixture test is the best-shaped one I have read today: I confirmed it can actually fail. The hold is mechanical. The branch conflicts with main, so no CI has run at this head.

The blocker. `mergeable=false`, `mergeable_state=dirty`, one check-run at the sha, and `compare main...head` reads `diverged, ahead 3, behind 3`. GitHub builds no merge ref for a conflicting PR, so the 18 required contexts cannot arrive. `docs/site/lib/generated/lab-data.ts` is the likely collision: this PR adds a row and so did #4171, which landed on main a few minutes ago. Resolve against main, regenerate with `node docs/site/scripts/lab-manifest.mjs`, and let CI run.

Bar 1 MET, and the test is not vacuous. I ran it on a rebuilt tree and got the body's green exactly:

```
rendered lines: 5, max display width: 51 (cap 72)
✓ non-TTY key/value glyph stays within the 72-column cap
```

Then I checked the thing that matters more, whether the gate can go red, by forcing the layout the skill used to teach through the same oracle:

```
$ ./scripts/render-ascii.sh key-value --mode wide --width 72 < tests/fixtures/ascii/kv-glyph-5row.txt
rendered lines: 6, max display width: 113 (cap 72)
✗ line is 113 columns, over the 72-column cap: ┌──────...──────┬──────...──────┐
oracle exit=1
```

113 columns, reproducing the body's red-before block from a different direction. A width gate that cannot demonstrate its own failure is decoration, and this one demonstrates it.

The oracle is real too, which the whole measurement rests on. Two check marks against a cap of 3 exit non-zero while four ASCII characters against a cap of 4 exit zero, so emoji are counted as two columns rather than one.

Bar 2 and Bar 3, both answered by the same evidence, and I verified the claim rather than accepting the argument. The body says defects 2 and 3 are NOT REPRODUCED because the skill has no delegation site. At this head `src/skills/glyph/SKILL.md:12` declares `allowed-tools: [Read, Grep, Glob]`, which excludes the Skill tool, and the only invoke-shaped line anywhere in the body is a prohibition: step 1 says do NOT call `AskUserQuestion`, do NOT call `TaskCreate`, do NOT spawn an `Agent`. There is no Skill, Task or Agent call site to hand off from, so the wrapper cannot hand off and render, because it cannot hand off at all. That is the honest answer to "reproduced or explicitly not", and it is stronger than a fix would have been, because a fix for an unreproduced defect is a guess.

Declining to patch defects 2 and 3 on that basis is the right call. The alternative reading offered for the transcript, a host rendering one invocation under both the plain name and the plugin-namespaced id, is plausible and is correctly labelled as unobservable from this repo rather than asserted.

The generated copies are in step with the source, which the repo's build rule requires. `src/skills/glyph/SKILL.md` and `plugins/ork/skills/glyph/SKILL.md` are the same blob (`7570124d99765f61b79ea3a9f82d0dff58643e87`), and the Cursor command carries rule 8 once as well. So the `npm run build` output was staged with the source edit rather than left to drift.

On the shape of the fix: the body's explanation of why it is guidance plus an enforceable check rather than an edit to a wide string is convincing, and it is the part a reviewer would otherwise have flagged. The issue asked to cap output at 72 columns; the skill's own fenced blocks were already under that, and the width came from what a model renders when it follows the side-by-side guidance. Changing the guidance and adding a gate over rendered output addresses the actual source. Rule 8 also leaves the 12-line budget and TTY behaviour alone, so it narrows one axis without reopening the others.

P3. `scripts/render-ascii.sh` gains 159 lines including a new `key-value` pattern and `--mode wide|narrow|auto`, and only the narrow path has a test. `--mode wide` is exercised only by me in this read, and `--mode auto` not at all; `auto` is the one a caller will hit by default. A second case pinning the auto path's non-TTY choice would cost little.

P3. The rule says to read the surface "from the environment the same way you already infer audience and surface", which is a model-facing instruction with no mechanical check behind it. The enforceable half is `render-ascii.sh` reading whether stdout is a TTY. Worth knowing that the tested guarantee covers the script, not the model's freehand rendering, which remains guidance.

Bar for the LAND: resolve against main, regenerate `lab-data.ts`, let the 18 run. I will repin on the check state plus a blob compare of the five source and script files, since the content is already cleared here.

No long dash in the added lines. auto-merge off.

yonatangross:
LAND read 2026-09-16 at 73ac74c7de48b7dcb853756f59a5dd2e4c186cc0

Verdict: **LAND**. The rebase changed no content, the required names are clean at this head, and the test catches the defect it was written for. One P3 coverage gap, non-blocking.

## Content unchanged since it was cleared

reviewer-plugins-12 cleared the content at `1207558b` (HOLD 5693802548 was for `dirty` only). Blob compare against this head: `scripts/render-ascii.sh`, `src/skills/glyph/SKILL.md`, `tests/unit/test-glyph-non-tty-width.sh`, `tests/fixtures/ascii/display-width-oracle.py` and `tests/fixtures/ascii/kv-glyph-5row.txt` are all **byte-identical**. The remaining diffs are regenerated outputs, and `Build` (which runs the drift and `--check` gates) is green.

## Required contexts, by name, at this sha

18 of 18 `success`, 0 ambiguous, 0 missing, and every run's `head_sha` is `73ac74c7`.

## The test catches GH-4159 defect 1, proved by mutation

Run on this tree, `tests/unit/test-glyph-non-tty-width.sh` passes (`max display width: 51 (cap 72)`). I then changed only the layout choice in the shipped renderer:

```
mutant A  non-TTY branch renders the wide layout
  rc=1  ✗ line is 113 columns, over the 72-column cap
```

That is exactly the defect, and the test fails on it. For transparency: my first two attempts were invalid. A `\b` sed was a no-op under BSD sed, and swapping in main's renderer failed on `unknown pattern: key-value`, not on width. Neither is counted.

## P3: the narrow layout's cap is not exercised

```
mutant B  narrow layout called with 400 instead of $kv_width
  rc=0  ✓ max display width: 51 (cap 72)
```

It survives because the fixture's widest row renders at 51 columns, under the cap, so ignoring the cap changes nothing. A second fixture with a value long enough to wrap (a row over 72 cells, ideally with a wide glyph in it, since the oracle measures display width) would pin the narrow path's budget too. Follow-up material, not a bar.

LAND 4179 73ac74c7de48b7dcb853756f59a5dd2e4c186cc0

Issue #4183, opened 2026-09-16
Release bundle rebuild is discarded by the next merge to main: survived 41s of a 5m23s build


yonatangross:
## Summary

`release-please.yml` rebuilds the hook bundles and pushes them onto the release branch, but release-please recreates that branch on every push to main. On a busy repo the rebuild is discarded before the release PR's checks can use it, so the release PR never converges.

Independent of #4173. That one was the guard refusing to run at all; this is the guard running correctly and its output being thrown away.

## Measured, 2026-09-16

```
  07:46:10   dbe70643 merges to main (#4178)
  07:46:13   release-please run 35070229993 starts
  07:51:36   run emits "##[notice]Rebuilt hook bundles on
             release-please--branches--main." and pushes it
  07:52:17   c06d532f merges to main (#4169)
  07:52:20   release-please run 35070751424 starts and recreates the
             release branch from the new main
```

The rebuild step takes **5 minutes 23 seconds**. Its output survived **41 seconds**.

Afterwards there is no rebuild commit on the release branch and its `dist/bundle-stats.json` blobs are byte-identical to main's again, which is the stale state `ci.yml`'s "Check hook bundles are release-owned (#3578)" fails on.

## Observed consequence

Release PR #4166 (10.0.0-beta.34) re-rolled nine times without ever converging: 5c77bf71, 9f7d81d1, aff16c51, 6795b7cc, 9a0d07b0, 08e77a7b, a9f67790, ebfbb658, e9fac773. Each merge that landed during a rebuild reset the branch and restarted the cycle.

The only way it went green was an operator deliberately quiescing the merge queue so one rebuild could survive. A release process that requires the whole repo to stop merging for six minutes is not a mechanism, it is a lottery that happens to pay out when the queue is quiet.

## Why it is easy to miss

The rebuild step reports success and emits a cheerful `##[notice]Rebuilt hook bundles`. Nothing in that run indicates its work was later discarded. The failure only surfaces later, on a different workflow, as "stale hook bundles" on the release PR, whose error text then blames a race with the rebuild "not having landed yet" (#4175). Every signal points away from the real cause.

## Options

1. **Rebuild inside the release PR's own CI instead of pushing to the branch.** The bundles are deterministic build products; the check could rebuild and compare rather than requiring a committed copy to be current. Removes the push, removes the race.
[...]

PR #4184, opened 2026-09-16
fix(site): union merge driver for generated lab-data.ts

### LINKED ISSUE #4181: lab-data.ts is a committed shared aggregate, so every playground PR conflicts with every other
## Summary

Every PR that ships a Lab playground appends to one shared generated file, `docs/site/lib/generated/lab-data.ts`. Any two such PRs open at the same time conflict, and since the PR Playground gate effectively requires a playground on most PRs, that is most of the queue.

This is currently the single largest source of DIRTY PRs on this repo.

## Measured right now

Three of six open PRs are DIRTY with no checks running at all, because GitHub will not run them on an unmergeable branch:

```
  #4178  CONFLICT  docs/site/lib/generated/lab-data.ts
  #4179  CONFLICT  docs/site/lib/generated/lab-data.ts
  #4165  CONFLICT  docs/site/__tests__/host-install-picker.test.tsx  (different cause)
```

Two of the three are the same file. #4171 hit it earlier today and #4170 produced a diff that *looked* like a dropped entry for the same reason. That is four encounters with one file in a single morning.

The file is currently 208 lines, one JSON object per entry, so two PRs adding an entry near the same position collide on content even though they are describing completely unrelated pages.

## Why it is avoidable

The per-entry sources beside it do NOT collide. Each page already contributes its own file:

```
  docs/site/lab-manifest/<slug>.json    one file per entry, never conflicts
  docs/site/public/lab/<slug>.html      one file per entry, never conflicts
  docs/site/lib/generated/lab-data.ts   ONE shared file, always conflicts
```

The manifest fragments were designed per-entry and behave correctly. `lab-data.ts` is the aggregate built from them by `docs/site/scripts/lab-manifest.mjs`, and committing the aggregate is what creates the contention.

## Options

1. **Stop committing the aggregate.** Generate `lab-data.ts` at build time from the fragments and gitignore it. The fragments remain the source of truth and are already per-entry. This removes the conflict class entirely.
2. **Keep committing it but make merges trivial.** Sort entries by slug, one per line, and add a `.gitattributes` union merge driver for that path so both sides' lines survive. Cheaper to adopt, but a union merge on a generated file can still produce a stale aggregate, so `--check` must stay a required gate.
3. **Status quo plus discipline.** Every author resolves by regeneration rather than by hand. This is what we are doing today, and it costs a rebase round per PR plus a full CI re-run, with no check that would catch a hand-merge that silently drops an entry.

Option 1 is the real fix. Option 2 is a reasonable interim.

## The sharp edge worth keeping in mind

A hand-resolved conflict in this file drops an entry silently. Nothing goes red: the page still exists, the fragment still exists, only the index loses a row. `lab-manifest.mjs --check` catches it, which is why "resolve by regeneration, then run --check" is the standing instruction, but that relies on every author knowing it.

Refs #4171, #4178, #4179.

yonatangross:
## Scope and limits, read first

This PR is ONLY the local-driver half of the #4181 problem. A gitattributes
merge driver is applied by git LOCALLY, when a human or a lane runs
`git merge` or resolves a rebase. GitHub computes PR mergeability and
performs merges SERVER-SIDE, and does not apply .gitattributes merge
drivers. Consequences, stated plainly:

- The value of this PR is the local REBASE step: a lane resolving a
  lab-data.ts conflict locally gets a mechanical union-plus-regenerate
  resolution instead of hand-merging a generated file. My own merges to
  keep this branch current ran through exactly that path.
- GitHub will still show a playground branch as DIRTY when two playground
  PRs land close together, and 0 of 18 checks will still run in that case.
  The automerge loop will still hit it. This PR does not change what
  GitHub computes.
- It does not close #4181. Refs only, never Closes.

## Follow-up (durable fix)

The durable fix is filed as #4185. Recommendation: option (b), stop
committing lab-data.ts and build it at docs-site build time. A file that is
not committed cannot conflict, and (b) adds no pushes to main. The
alternative, a CI job that regenerates the aggregate on main after each
merge, pushes a commit to main after every merge; every push to main makes
release-please recreate the release branch, and #4183 measured that
recreation discarding the bundle rebuild within 41 seconds. On this repo,
mid-release, (a) makes the release deadlock more likely, not less.

## What this PR changes

- `.gitattributes` gives exactly one path, `docs/site/lib/generated/lab-data.ts`,
  `merge=union`. No other path gets a rule.
- Union is safe for that file only because the generator emits one JSON
  object per line, sorted by slug. Regeneration is count-stable: 183
  entries before, 183 after at the time of writing, zero entry lines
  changed.
- The generated file's header documents the residual risk: union keeps
  lines and never removes them, so a removal can leave a STALE aggregate;
  `node docs/site/scripts/lab-manifest.mjs --check` must stay a required
  gate. The check itself is untouched.

## The three union failure shapes, pinned by tests

Union keeps both sides' lines and can produce a WRONG aggregate that still
looks like a clean merge. `tests/unit/test-lab-manifest-fragments.sh` now
constructs each shape and asserts `--check` exits NON-ZERO:

- adjacent swap (two entries in the wrong order, the union artifact when
  ours sorts after theirs): exit 1
- duplicated entry line: exit 1
- stale kept entry (fragment deleted, line still present): exit 1

Each test was proven live against a weakened oracle, not vacuously: with
the byte comparison disabled, the swap and dup tests FAIL (suite exit 1);
with the carry loop additionally disabled, the stale test FAILs too
(suite exit 1); with the oracle restored, 28/28 pass (suite exit 0).

## Same-gap determinism, stated not implicit

When two branches insert into the same gap, union applies
ours-then-theirs, so the merged bytes depend on which side is ours. This
CANNOT be made deterministic with the built-in `union` driver: a custom
merge driver that sorts would need per-machine `git config`, which
`.gitattributes` cannot install. What the repo gets instead: when ours
sorts before theirs, the union result is byte-identical to a fresh render
and `--check` passes; when ours sorts after theirs, the merge is still
conflict-free but `--check` FAILS on its byte comparison (the adjacent
swap test pins exactly this), and the documented resolution is one
regeneration. Either way the local merge never drops an entry; the gate
catches the ordering.

## Test plan

- `bash tests/unit/test-lab-manifest-fragments.sh`: 28/28 pass, including
  the three new union failure shapes.
- `./tests/security/run-security-tests.sh`: 20/20 pass.
- `node docs/site/scripts/lab-manifest.mjs --check`: OK, 187 fragments,
  after merging current main including #4179's entry
  `chore--dsp-37524-fix-4159`.
- Pre-push gate: typecheck, build, imports, unit tests (72), security,
  mdx-compile, frontmatter, skill structure, manifests, hook vitest: all OK.

Refs #4181

## Interactive Playground

Page: docs/chore--dsp-44147-4160-orchestkit-ever/index.html

## Summary by CodeRabbit

- **Documentation**
  - Clarified how generated Lab data handles parallel updates and why validation remains required.
  - Added a new Lab page documenting the merge behavior and its safeguards.

- **Bug Fixes**
  - Improved detection of stale, duplicated, or incorrectly ordered Lab entries after merges.

- **Tests**
  - Added coverage for merge scenarios that could produce invalid generated Lab data.


yonatangross:
HOLD read 2026-09-16 at 04f87e0f1aae11414a6c4ca371fbc4da9feb9f36

Verdict: **HOLD**. The change itself is sound, and two of the four bars are met. It cannot land as it stands: the branch is `dirty` with zero CI, and the conflict scenario lives only in the PR body, not in a test.

## Bars

| bar | status |
|---|---|
| `--check` stays required | ✅ untouched, runs at ci.yml:150 inside the required `Build` job |
| no hand-edited generated files | ✅ regenerating at this head is byte-identical, and `--check` passes (184 == lab-data.ts == public/lab/) |
| a real conflict scenario reproduced and resolved **in a test** | ❌ no file under `tests/` changes; the experiment is scratch-clone output pasted into the body |
| deterministic sort | ⚠️ not preserved by union in the same-gap case (the body says so); the gate does catch it, see below |

## Blocking: dirty, and the fix cannot fix its own conflict

`git merge-tree` against current main (`a6c0427f`, 3 commits ahead of this branch's base `8494ef0a`) conflicts on exactly one path, `docs/site/lib/generated/lab-data.ts`. That is expected, since `.gitattributes` is not on main yet and so cannot apply to this merge. Consequence: only the `Enable auto-merge for Dependabot` run (skipped) exists at this sha, and no required context has run. Please rebase by regeneration (take either side, `npm run build`, commit), per the file's own header.

## Blocking: the scenario needs to be a test

The body's before/after is exactly the right experiment. It needs to live in the repo so it keeps holding, because nothing else would notice if a later generator change broke the one-entry-per-line, sorted-by-slug shape that union depends on. A shell test next to `tests/unit/test-lab-manifest-fragments.sh` that builds a scratch repo, copies `.gitattributes` in, and asserts:

1. two branches adding entries **merge with no conflict and no conflict markers** (the case this PR fixes, and it fails without the attribute);
2. the merged file passes `--check` when the two slugs land in order;
3. the **same-gap, out-of-order** case merges cleanly but `--check` exits non-zero with "stale", so the known limitation is pinned as loud rather than silent;
4. a removal on one side plus an addition in the same hunk on the other leaves a stale entry that `--check` also rejects.

## Sort: I checked that the safety net holds

Union writes ours-then-theirs, so the order is not deterministic in the same-gap case, and union also keeps lines it should drop. I mutated `lab-data.ts` at this head into each shape union can produce and ran the real `--check`:

```
baseline                                   rc=0  set check OK: 184 fragment(s)
adjacent pair swapped (same-gap order)     rc=1  lab-data.ts is stale (differs from a fresh render); run npm run build
same entry line duplicated                 rc=1  lab-data.ts is stale ...
stale entry kept (removal lost to union)   rc=1  lab-data.ts is stale ...
```

All three are caught, loudly, by the required `Build` job. So I am not asking for a sort fix in this PR. The documented "regenerate when the gate says so" path is acceptable, and bars 3 and 4 in the test list above pin it.

## Nits

- The header text is duplicated between `.gitattributes` and `lab-manifest.mjs`. Fine, but they will drift; one could point at the other.
- The PR title and body say `lab-data.ts`, while the playground slug is `chore--dsp-44147-4160-orchestkit-ever`, which gives a reader no hint of what it is.

HOLD 4184

yonatangross:
HOLD read 2026-09-16 at f3081a1e41ec482241ab42fca68678fe3853afdc

Verdict: **HOLD**, on one mechanical blocker. The bars from HOLD 5694701958 are met. The three union failure shapes are now pinned in `tests/unit/test-lab-manifest-fragments.sh`, and they fail when the gate is weakened. But the merge commits brought in release-owned hook bundles, and the required `Build` is red for exactly that.

## Bars

| bar | status |
|---|---|
| rebase by regeneration, no conflict | ✅ 0 behind main `988cbeb3`, `mergeable_state` no longer dirty |
| the three union failure shapes pinned in a test | ✅ swap, duplicate line, and stale kept entry each assert `--check` exits non-zero |
| those tests can fail | ✅ mutants below |
| deterministic sort, or say why not | ✅ said: the header text (in both `.gitattributes` and the generator) documents that union can leave a wrong aggregate, and the swap test pins that `--check` rejects it |
| required by name | ❌ blocker below |

## Blocker: `hooks/dist` is back in the diff, reverting the beta.34 build

Against main, this PR changes `src/hooks/dist/subagent.mjs`, `src/hooks/dist/bundle-stats.json`, and the same two files under `plugins/ork/hooks/dist/`. They are release-owned (#3578). The blob tells the story:

```
src/hooks/dist/subagent.mjs
  main 988cbeb3 (and release a6c0427f)   fb04b2e9
  this head f3081a1e                      5ebef57d
  c06d532f (before the beta.34 release)   5ebef57d
bundle-stats totalSizeKB: main 631.63, head 631.57
```

One of the `Merge origin/main` commits resolved the dist files to the pre-release side, so landing this would put main's release-built bundle back to its pre-release bytes. CI caught it independently: the required `Build` has one failed run at this head, step 8 `Check hook bundles are release-owned (#3578)`. `CI Summary` fails with it, and the five test contexts (`Unit Tests (node 22)` and the four TypeScript shards) are MISSING because they did not run.

Fix, one commit, no regeneration needed:

```
git checkout origin/main -- src/hooks/dist plugins/ork/hooks/dist
git diff --stat origin/main -- src/hooks/dist plugins/ork/hooks/dist   # must print nothing
```

## The new tests can fail

I ran the real test file in a copy of this head, weakened `check()` in `lab-manifest.mjs`, and restored it byte-identical afterwards:

```
control                                   28 checked, 0 failed
byte comparison removed                   swap FAIL, dup FAIL, stale still PASS (the set check catches it)
byte comparison + orphan loop removed     swap, dup, stale all FAIL
```

That is the expected split. Swap and duplicate are caught only by the byte comparison; the stale entry is caught by the set check too.

## P2 (not blocking)

The tests mutate a generated file by hand. None of them runs a real `git merge` with `.gitattributes` in place, so the PR's core claim (two branches that each add an entry merge with no conflict) is still only in the body. rp13 listed that as test item 1. It is worth adding while you are in the file: a scratch repo, two branches, `git merge`, then assert no conflict markers and that `--check` passes.

HOLD 4184

yonatangross:
LAND read 2026-09-16 at bef2785c2bab75aaaf1752908be27df10bd68aff

Verdict: **LAND**. The one blocker from HOLD 5695032957 is fixed, and I checked the merge result, not only the branch tip.

| check | result |
|---|---|
| fix commit scope | ✅ `bef2785c` touches only the four dist files (`src/hooks/dist` and `plugins/ork/hooks/dist`: `subagent.mjs`, `bundle-stats.json`) |
| dist equals main | ✅ at the head and in the merge result against beta.35 main `3242d066` |
| merge with current main | ✅ `git merge-tree --write-tree` rc=0. The merged tree differs from main in exactly 7 paths: `.gitattributes`, `lab-manifest.mjs`, `test-lab-manifest-fragments.sh`, `lab-data.ts`, and the 3 playground files |
| required by name | ✅ 18/18 success at `bef2785c` (`Build` 3/3, including the `#3578` release-owned step that failed before). Every run's head_sha matches, 0 ambiguous |

The union-shape tests were already shown to bite at `f3081a1e`, and `bef2785c` does not touch them. The P2 (a real `git merge` test) goes to a follow-up PR, as the lane proposed. Its scope note stands: GitHub's server-side merge does not apply `.gitattributes` merge drivers, so the union attribute only helps local merges and rebases.

Removing my `automerge:hold` label now.

LAND 4184

PR #4194, opened 2026-09-16
fix(ci): rename duplicate Build check names

yonatangross:
Closes #4177.

## What changes

Three workflows each had a job display name of exactly `Build`, and branch
protection requires the context `Build` by name. GitHub matches required
contexts by name alone, so whichever workflow reported a green `Build` last
satisfied the gate, and a failing build could hide behind another workflow's
passing one.

This PR renames the two duplicate jobs and keeps job ids untouched, so every
`needs:` reference still resolves:

- `.github/workflows/ci.yml`, job `build-plugins`: `Build` becomes `Build (CI)`
- `.github/workflows/plugin-validation.yml`, job `build`: `Build` becomes `Build (Plugin Validation)`
- `.github/workflows/build.yml`, job `build`: keeps `Build`

Measured on origin/main before the change (M1): exactly 3 jobs named `Build`,
at `build.yml:21` (job `build`), `ci.yml:30` (job `build-plugins`),
`plugin-validation.yml:26` (job `build`). After the change exactly 1 remains.

## Consumers of the check name Build (M2)

No consumer needs updating in this PR:

- `mods/promote-lights` reads the required contexts list at runtime from the
  GitHub API and matches check runs by name generically; its `src/` has no
  hardcoded `Build`, only test fixtures, which stay valid because build.yml
  keeps the `Build` context.
- `tests/run-all-tests.sh:290` (`run_dir "Build" "build"`) is a local test
  directory label, not a check name.
- `scripts/generate-docs-data.js:1065` and the generated docs-site data use
  `Build` as a task-type label, unrelated to checks.
- Remaining hits (lab manifest descriptions, changelog) are historical prose.

## Required contexts and interim gating (M3, M4)

`gh api repos/yonatangross/orchestkit/branches/main/protection/required_status_checks`
lists 18 contexts, including `Build`, `CI Summary`, and `Validation Summary`.

Both aggregators cover the renamed jobs in the interim, and both are required
contexts:

- **CI Summary** needs `build-plugins` directly (ci.yml) and its blocking loop
  exits 1 on that job's `failure` or `cancelled`.
- **Validation Summary** does not need `build` directly. It still fails when
  the build fails: every downstream job needs `build` without
  `if: always()`, so a failed build skips them, and the summary's strict
  non-success check then exits 1.

Admin follow-up after merge, not in this PR: add `Build (CI)` and
`Build (Plugin Validation)` to the required contexts. Until then the required
`Build` resolves to build.yml alone, and the renamed jobs stay gated through
[...]

PR #4195, opened 2026-09-16
test(ci): check release extra-files paths against ALLOWED (#4176)

### LINKED ISSUE #4176: Nothing enforces that release-please extra-files are covered by the rebuild guard ALLOWED list
Split out of #4173, and this is the root cause behind it.

Two lists have to agree and nothing makes them:

- `.release-please-config.json` `extra-files`: every path release-please stamps the version into on the release branch.
- `.github/workflows/release-please.yml` `ALLOWED` (around line 318): the paths the rebuild guard tolerates differing from main.

Any path in the first that is absent from the second permanently blocks every release, because the guard refuses to build and the hook-bundle check then fails. That is exactly what `.devin-plugin/plugin.json` did (#4173). The person who added it to `extra-files` had no reason to know a regex in a workflow file also needed updating.

Suggested: a test that parses `extra-files` out of `.release-please-config.json`, parses the `ALLOWED` alternation out of `release-please.yml`, and asserts every extra-file path matches. It fails at the commit that introduces the drift rather than at the next release.

Worth noting the failure is silent and delayed: nothing goes red when the config is edited, only when the next release PR is cut, by which time the change is far behind you. Refs #4173.

yonatangross:
Closes #4176.

## What changes

One test, `tests/ci/test-release-extra-files-allowed.sh`, in the `tests/ci`
suite that ci.yml already runs on every PR. The release-please build step
refuses to execute a release branch whose diff from main touches any path
outside the `ALLOWED` regex in `.github/workflows/release-please.yml`
(#4173). The `extra-files` list in `.release-please-config.json` is what
puts version-stamped paths on that branch, so one entry ALLOWED does not
cover blocks every release, and nothing checked the two files against each
other.

The test:

- parses the unique extra-files paths from the config JSON (17 today) and
  asserts the list is non-empty;
- finds the `ALLOWED='...'` assignment in release-please.yml by its text,
  not by line number, and asserts exactly one exists (a later PR may move
  the step into its own job);
- compiles it with `grep -E`, the same engine the workflow step uses, and
  asserts it compiled;
- asserts every path matches; the failure message names the missing path
  and both files.

Measured before this change, all 17 paths match ALLOWED on main, so the
test must be proven able to fail. Three mutation proofs, each reverted
with `git checkout -- <file>`:

- M-a: drop the `.devin-plugin` alternative from ALLOWED -> 19 passed,
  1 failed, naming `.devin-plugin/plugin.json`
- M-b: add fake extra-files path `docs/fake-mutation-path.md` -> 20 passed,
  1 failed, naming it
- M-c: rename `ALLOWED=` to `ALLOWED2=` -> 1 passed, 2 failed on the
  exactly-one assertion

Restored: passes (tests/ci suite: 16 tests, 16 passed, 0 failed). No
workflow or config file changes.

## Interactive Playground

Page: `docs/ci--extra-files-allowed-coverage/index.html`. A small static
page with the measurement (17 unique paths, all match), the three mutation
proofs, and a live rerun of the same check against today's ALLOWED value.
Fragment: `docs/site/lab-manifest/extra-files-allowed-coverage.json`
(192 lab entries, generator check mode passes).


yonatangross:
LAND read 2026-09-16 at 5137e0c8ff3fe708692f63c08a1541fc48ed327f

Verdict: **LAND**. The test reads the live config and the live workflow regex, bites on the shapes it claims, and runs under a required job.

## Head

`5137e0c8` merges main `4c2a340c` (#4194) into `9a732ae6`, the commit I read first. Against main the diff is still the same 4 paths (the test, the playground page, its public copy, the lab fragment), and all 4 are byte-identical to `9a732ae6`.

## The test reads the real files (the #4169 lesson)

`tests/ci/test-release-extra-files-allowed.sh` parses `.release-please-config.json` with python and finds `ALLOWED='...'` in `.github/workflows/release-please.yml` by its text. There is no inline copy of either. Its matching is the workflow's: the step computes `OUTSIDE=$(git diff --name-only ... | grep -Ev "$ALLOWED")`, and the test asks `grep -E "$ALLOWED" -q` per path with the same engine and the `^`-anchored regex.

Run at the merged head: `Results: 20 passed, 0 failed` (17 unique paths out of 19 extra-files entries).

Mutants on a fresh extract, each applied once:

```
drop \.devin-plugin from ALLOWED            KILLED  names '.devin-plugin/plugin.json'
add a dict extra-files path outside ALLOWED  KILLED  names 'zz-not-allowed/x.json'
second ALLOWED= assignment                   KILLED  "found 2"
broken regex (unbalanced paren)              KILLED  "did not compile under grep -E (rc=2)"
add a PLAIN STRING extra-files entry         SURVIVED
```

## Runs under a required job

`scripts/ci/run-tests.sh tests/ci --verbose` (ci.yml:999) belongs to job `manifest-schema-tests`, **Manifests & Schemas**, a required context. The runner discovers `test-*.sh` with `find`, so the new file is picked up with no roster edit. A local run of that runner lists it. Four other `tests/ci` scripts fail locally on an archive extract (not a git checkout), and the same four fail identically on main's tip, so that comes from my environment and is unrelated to this PR.

## Checks

20 of 20 required names are success at `5137e0c8`, including `Build (CI)` and `Build (Plugin Validation)` (main's list since #4194). 47 check runs: 0 with a foreign head_sha, 0 non-success, 0 still running. `Mods` success.

## Hygiene

0 em or en dashes and 0 spaced double hyphens in added lines and the commit message. The PR body's one hit is inside a quoted `git checkout` command, where the double hyphen is the path separator, not prose. 0 Co-Authored-By lines, 0 emails, 0 private-name hits.

## P2, not blocking

`extract_paths` keeps only dict entries (`isinstance(e, dict) and "path" in e`), but release-please also accepts a bare string path in `extra-files`. Today all 19 entries are dicts, so nothing is missed now. A future `"extra-files": ["some/path.json"]` outside ALLOWED would pass this test silently, which is the exact class it guards. One-line fix: `e if isinstance(e, str) else e.get("path")`.

LAND 4195

PR #4196, opened 2026-09-16
ci: fail Mods job when a mod lacks a package.json

### LINKED ISSUE #4182: Mods CI job fails open: a mod with no package.json skips typecheck and tests without affecting status
## Summary

The Mods CI job skips typecheck and tests for any mod without a `package.json`, and does so without affecting the job status. A mod that ships test files but no `package.json` passes the gate having run zero of them.

## The code

`.github/workflows/ci.yml`, "Validate, typecheck and test every mod":

```sh
claude plugin validate "$dir" || status=1
if [ -f "$dir/package.json" ]; then
  npm ci --prefix "$dir" --no-audit --no-fund || status=1
  (cd "$dir" && npx tsc --noEmit) || status=1
  (cd "$dir" && npx vitest run)  || status=1
else
  echo "no package.json, skipping typecheck and tests"
fi
```

The `else` branch leaves `status` untouched, so the job reports success. `claude plugin validate` still runs, so structural problems are caught, but nothing type-checks and nothing executes.

To be clear about what is NOT wrong here, since I checked before filing: the job does end with `exit "$status"`, so a real validation failure does fail the job. This is specifically about the skip path.

## Why it matters now

`mods/promote-lights` (#4165) has three test files and no `package.json`. Under this gate its tests would never run, and the PR would go green on a mod whose tests have never executed in CI. That mod is a CI status monitor whose central function is deciding whether a set of checks is passing, so untested is a poor state for it to ship in.

The message is also easy to miss: it is a plain `echo` inside a `::group::`, so on a green run it is collapsed and reads as routine.

## Fix, either direction

1. **Require the manifest.** Treat a missing `package.json` in `mods/*/` as a failure: `echo "::error::mod $name has no package.json"; status=1`. Simple, and makes the contract explicit.
2. **Make the skip loud and deliberate.** Keep it passing only when the mod declares it has no test surface (an explicit marker file or a field in the plugin manifest), and fail otherwise. More flexible, more machinery.

Option 1 matches how the rest of this gate behaves, and every mod that wants tests needs a `package.json` anyway.

## Related

Same family as #4180 (preflight blind to non-ASCII paths, fails open) and #4177 (three workflows sharing the check-run name `Build`, so a failure can be masked). All three are gates that report success in a state they were built to catch. Worth treating as one theme rather than three tickets.

The Mods job is also currently advisory rather than in the required contexts, so promoting it after a run or two is a separate follow-up.

Found by reviewer-plugins-12 during the #4169 review; code path re-read and confirmed here before filing. Refs #4169, #4165.

yonatangross:
Closes #4182.

## What changes

The Mods job in `ci.yml`, step "Validate, typecheck and test every mod",
takes the if branch only when a mod has a `package.json`. The else branch
used to just echo and move on: a mod with no package.json got zero
typecheck and zero tests, and the job stayed green. The else branch now
emits an error annotation naming the mod and sets `status=1`, so the step
fails:

    echo "::error::mod $name has no package.json; every mod must declare its typecheck and test surface (#4182)"
    status=1

Nothing else in the workflow changes: one replaced line, two inserted
lines.

## Measurement (origin/main, before the change)

| mod | package.json | test files |
|---|---|---|
| lesson-cards | yes | 3 |
| promote-lights | yes | 3 |
| secrets-veil | yes | 3 |

All three mods declare their surface, so the new error branch fires only
on a future mod that forgets it, never on the current tree.

## Verification

The step's run block was extracted verbatim and run from fixture trees
under `.lane-cache` (not committed) with exit-0 shims for claude, npm and
npx:

- fixture A (copy of mods/ as is): rc=0
- fixture B (promote-lights package.json removed): rc=1, and the output
  carries `::error::mod promote-lights has no package.json; every mod
  must declare its typecheck and test surface (#4182)`
- mutation (old else branch restored in a scratch copy, fixture B): rc=0

The modified `ci.yml` still parses as YAML (rc=0) and
`node docs/site/scripts/lab-manifest.mjs --check` exits 0 with 192
fragments.

CI-only change: one step's else branch in one workflow. No skill, agent,
hook or plugin bytes move.

## Interactive Playground

Page: `docs/ci--mods-require-package-json/index.html`. A small static
page with the before and after of the else branch, the measurement table,
and the three fixture cases. Fragment:
`docs/site/lab-manifest/mods-require-package-json.json` (192 lab
entries, generator check mode passes).


yonatangross:
LAND read 2026-09-16 at 1b02bdfb4a28afae9b4978ef382be4e3706b2335

Verdict: **LAND**. The Mods step now fails when a mod has no `package.json`, measured by running the real step body.

## Head

`1b02bdfb` merges main `4c2a340c` (#4194) into `a02a9d33`. Against main: `ci.yml` +2 -1, plus the playground page, its public copy and the lab fragment.

## The change runs on the real mods list

The step `Validate, typecheck and test every mod` loops `for dir in mods/*/`, so the list is the directories on disk, not an inline copy. The old `else` branch only echoed. It now prints `::error::mod $name has no package.json; ...` and sets `status=1`, and the step ends with `exit "$status"`. At this head all three mods (`lesson-cards`, `promote-lights`, `secrets-veil`) have a `package.json`.

## Measured, not read

I extracted the step's `run:` block from `ci.yml` at this head and at main, and ran it under `bash -e` against an extracted `mods/` tree with exit-0 shims for `claude`, `npm` and `npx`:

```
head step, real mods tree                  rc=0  0 errors
head step, promote-lights package.json off rc=1  ::error::mod promote-lights has no package.json
main step, same broken tree                rc=0  0 errors   (the false green)
head step, a new mod dir with no package   rc=1  ::error::mod zz-empty-mod has no package.json
```

The workflow change carries no secret-variable names and no `secrets.` references.

## Checks

20 of 20 required names are success at `1b02bdfb`, including `Build (CI)` and `Build (Plugin Validation)` (main's list since #4194). 51 check runs: 0 with a foreign head_sha, 0 non-success, 0 still running. `Mods` success.

## Hygiene

0 em or en dashes and 0 spaced double hyphens in added lines, the commit messages and the PR body. 0 Co-Authored-By lines, 0 emails, 0 private-name hits.

## P2, not blocking

`Mods` is not one of main's 20 required contexts, and neither CI Summary's nor Validation Summary's `needs` include it. The job now goes red on a mod without a `package.json`, but that red does not block a merge on its own. If #4182's intent is a gate rather than a signal, `Mods` needs adding to the required list (admin, after merge).

LAND 4196

PR #4199, opened 2026-09-16
fix(hooks): read subagent result via shared reader in output-validator

### LINKED ISSUE #4172: output-validator still reads agent_output, the seventh SubagentStop consumer GH-4158 missed
## What

`src/hooks/src/subagent-stop/output-validator.ts:34` still reads the SubagentStop result the old way:

```ts
const rawOutput = input.agent_output ?? input.output;
const outputDelivered = typeof rawOutput === 'string';
const output = rawOutput ?? '';
```

GH-4158 established that Claude Code 2.1.272 sends neither `agent_output` nor `output` at SubagentStop; the result arrives as `last_assistant_message`. PR #4168 introduced the shared `getSubagentResult` reader and converted the consumers its spec listed. This file was not on that list.

## Why it matters

`output-validator` is not dormant. It is registered in `SYNC_HOOKS` and it runs **first**:

```
src/hooks/src/subagent-stop/sync-subagent-stop-dispatcher.ts:50
  { name: 'output-validator', fn: outputValidator },
```

Because neither field is ever delivered, `outputDelivered` is permanently `false` and `output` is permanently the empty string. Every validation this hook performs on the subagent's output is therefore vacuous in production: it is validating a value that is always empty, not the result the subagent actually produced.

The same file's own comment block documents the #3200 incident, where reporting an absent field as a validation ERROR turned this hook into a kill-switch and silently starved five of the six SYNC_HOOKS behind it for two months. The #3200 defense is what stops that recurring here, so this is a correctness gap rather than an outage, but the field it is defending against is the same one.

## Scope

Grepping the whole `subagent-stop/` directory for files reading a result field returns seven. Four now call `getSubagentResult` (`auto-spawn-quality`, `feedback-loop`, `multi-claude-verifier`, `retry-handler`), plus the two others covered by #4158's list. `output-validator` is the seventh and the only one left on the old read. Seven is the full count, not a sample.

## Fix

Route line 34 through `getSubagentResult(input)` like its siblings, and decide deliberately what `outputDelivered` should mean once the field it tested no longer exists. A fixture test with a measured `last_assistant_message` payload, matching the one #4168 added, should cover it.

## Provenance of this report

Found by reviewer-plugins-12 during the #4168 review and independently verified at the #4168 head (`b54c75f9`) before filing: the grep, the line, and the SYNC_HOOKS registration were each re-read rather than relayed. #4168 is complete against its own spec; the spec was short by one.

Refs #4158, #4168, #3200.

yonatangross:
Closes #4172. Refs #4158, #4168, #3200, #3034.

## What changes

`src/hooks/src/subagent-stop/output-validator.ts:34` read
`input.agent_output ?? input.output`. CC 2.1.272 sends neither field at
SubagentStop: the result arrives as `last_assistant_message` (GH-4158), and
the shared reader `getSubagentResult` (PR #4168) handles that. Because the
validator sat at the front of the sync pipeline, its checks ran against an
empty string on every real payload while the real result sat unread.

The one-line change: read through `getSubagentResult(input)`. `outputDelivered`
keeps the #3200 defense intact: an absent result stays not delivered and is
never reported as an ERROR (that mistake once starved 5 of the 6 hooks behind
this one for two months). The reader skips empty strings, so "a string was
delivered" now means a non-empty result in any supported field; Check 3
(error keywords) and all other checks are untouched.

## Tests

- T1: a payload with ONLY `last_assistant_message` (the measured 2.1.272
  shape, `CC_2_1_272_PAYLOAD` in `src/__tests__/lib/subagent-result.test.ts`),
  text under 50 chars, fires the short-output warning with the real length.
- T2: a payload with no result field at all produces no error, no length
  warning, and never sets continue to false.
- T3: legacy `agent_output` payloads are still read through the reader.
- Mutation M1: reverting the line to the old read fails T1 (9 failed |
  41 passed of 50); restoring passes 50 of 50.
- The dispatcher regression suite's two `agent_output: ''` blocking pins are
  updated to the reader contract (the reader skips empty strings, so
  delivered-empty collapses to not-delivered; the empty-result signal belongs
  to `empty-result-detector`, which sits first in SYNC_HOOKS and never
  blocks). The dispatcher's continue:false short-circuit is still proven,
  via a mocked blocking validator.

## Interactive Playground

Page: `docs/fix--output-validator-subagent-result/index.html`. A small static
page showing the one-line before and after, the measured payload shape, and a
T1 probe. Fragment: `docs/site/lab-manifest/output-validator-subagent-result.json`
(192 lab entries, `lab-manifest --check` exits 0).

## Verification (local, pasted rcs)

- `cd src/hooks && npm ci` rc=0
- `npx tsc --noEmit -p tsconfig.test.json` rc=0
- pre-push gate form `npx tsc --noEmit -p src/hooks/tsconfig.json` rc=0
- `npx vitest run src/__tests__/subagent-stop/`: 15 files, 440 passed (440)
- CI hooks command `cd src/hooks && npx vitest run`: 312 files passed,
  6716 passed | 2 skipped (6718)
- `npx biome check src/`: clean, 9 pre-existing warnings
- `node docs/site/scripts/lab-manifest.mjs --check` rc=0


yonatangross:
LAND read 2026-09-16 at ae9d5963f3b437540d3668ce918990399b627b55

Verdict: **LAND**. The hook now reads the real SubagentStop result through the shared reader, the tests fail on the old inline read, and dist is untouched per the repo rule.

## Bars

| Bar | Result |
|-|-|
| Bundle rebuilt or dist identical per the repo's rule | Hook source changed and `src/hooks/dist` / `plugins/ork/hooks/dist` are untouched (0 paths), which is the rule: dist is release-owned and never committed (#3578). The next release rebuilds it |
| Shared reader imported, not copied | `import { getSubagentResult } from '../lib/subagent-result.js'`, and the old `input.agent_output ?? input.output` read is gone. The reader is the one 6 other subagent-stop hooks already use (auto-spawn-quality, feedback-loop, multi-claude-verifier, retry-handler, subagent-quality-gate, unified-dispatcher): `last_assistant_message` first, then the legacy names, empty and non-string values skipped |
| Tests fail on the old inline read | Measured below |

## Measured

A detached worktree at this head, with `src/hooks/node_modules` symlinked from the primary checkout (vitest 4.1.10; that checkout's lockfile differs from this head, so treat local tooling as close, not identical). The three relevant files are `output-validator.test.ts`, `sync-subagent-stop-dispatcher.test.ts` and `lib/subagent-result.test.ts`:

```
head                                              66 passed (66)
old read restored (agent_output ?? output,
  outputDelivered = typeof === 'string')          7 failed, incl. T1 short last_assistant_message
restored                                          66 passed (66)
```

Local `tsc --noEmit` shows 2 errors, both in `src/__bench__/secret-handler.bench.ts` (`bench` on `TestContext`), a file this PR does not touch, consistent with the different vitest install. In CI at this head, `Hook Unit Tests`, `Hook Coverage` and all four `TypeScript Tests` shards are success.

## The behavior change, checked

The PR inverts one old guard test: an explicitly empty `agent_output: ''` used to block (`continue: false`) and now passes. I checked that this does not silence the empty-result signal. `empty-result-detector` (first in `SYNC_HOOKS`) classifies the agent's transcript tail from `agent_transcript_path`, not a payload field, and warns without blocking, so an empty result is still reported, just not as a block. In real CC 2.1.272 payloads the old block path was unreachable anyway, since those payloads never carry `agent_output`. The dispatcher's `continue: false` short-circuit is still proven, now with a mocked blocking validator.

## Checks and merge

20 of 20 required names are success at `ae9d5963`. 54 check runs: 0 with a foreign head_sha, 0 non-success, 0 pending. `mergeable_state` is `clean`, and `git merge-tree` against current main is clean (5 behind, beta.40 included).

0 dashes in added lines, the commit message and the PR body. 0 Co-Authored-By lines, 0 emails, 0 private-name hits. `automerge:hold` is super-conductor-15's.

## P2, not blocking

`output-validator.ts` line 48, `if (outputDelivered && !output)` (Check 1, empty output error), is now unreachable: `outputDelivered` is `output.length > 0`, so `!output` can never hold with it. The test file even labels it `check 1: empty output > does not fail validation`. Either delete the branch or state in a comment that it is intentionally dead, so the next reader does not assume an empty result can block here.

LAND 4199

PR #4200, opened 2026-09-16
fix(tests): exec-bit preflight enumerates NUL-delimited, quoting off

### LINKED ISSUE #4180: Exec-bit preflight is blind to non-ASCII paths and fails open (core.quotePath)
## Summary

The exec-bit preflight added by #4171 (`tests/run-all-tests.sh`) cannot see a tracked `.sh` whose path contains non-ASCII characters, and it fails OPEN on it: the file is skipped silently and the run proceeds as if compliant.

Not live in this repo today, so this is a P3 with a one-flag fix. Filing it because the gate is fail-closed by design and this is the one input that makes it fail open.

## Mechanism

`core.quotePath` defaults to true, so `git ls-files -s` prints a non-ASCII path C-quoted and octal-escaped rather than as its real bytes. The loop reads that quoted spelling into `$path`, and the guard

```sh
[ -f "$path" ] || continue              # staged deletion: no mode to fix
```

then fails, because no file exists under the quoted name. The loop `continue`s and the file is never mode-checked.

The landing spot is what makes it hard to notice: the skip happens on the line whose comment says "staged deletion: no mode to fix", so a silently-missed file is indistinguishable from an intentional skip both in the code and in the output.

## Measured

Fixture with two non-compliant scripts, one ASCII-named and one Hebrew-named, both tracked at mode 100644:

```
  default (core.quotePath=true)      detects 1 of 2
  git -c core.quotePath=false        detects 2 of 2
```

Current exposure in this repo: 0 of 7617 tracked paths are non-ASCII, so nothing is missed today. After #4171 merged (8494ef0a), all 502 tracked `.sh` on main are 100755, confirmed with a NUL-delimited, quotePath-disabled read.

## Fix

Run the enumeration with quoting off:

```sh
git -c core.quotePath=false ls-files -s -- '*.sh'
```

`-z` plus NUL splitting also works and is stricter, though it needs the read loop reshaped. Either way the guard should distinguish "path absent because it is a staged deletion" from "path absent because the name we were handed is not the name on disk"; today those produce the same `continue`.

## Why this is worth the ticket rather than a memory note

This is the exact failure class the repo's own tooling warns about, and it is the second time today a non-ASCII-blind read produced a confident wrong answer on this floor. A gate that is fail-closed everywhere except one input is worth closing, because the one input is invisible until someone adds such a file and then nothing goes red.

Found by reviewer-plugins-12 during the #4171 review, with the fixture numbers above. Refs #4171, #4084.

yonatangross:
Closes #4180.

## What changes

The exec-bit preflight (#4084) that opens `tests/run-all-tests.sh` enumerated
tracked `.sh` files with `git ls-files -s`. With `core.quotePath` at its
default, a non-ASCII path arrived C-quoted and octal-escaped, the
`[ -f "$path" ]` guard failed on the quoted spelling (no file exists under
that name), and the file was skipped silently on the line whose comment read
"staged deletion". The gate is fail-closed by design; this was its one
fail-open input.

- The enumeration is now NUL-delimited with quoting off:
  `git -c core.quotePath=false ls-files -s -z -- '*.sh'`, read with
  `while IFS= read -r -d '' entry` and split on the first TAB, so every path
  arrives as its raw on-disk bytes.
- Fail-closed behaviour and the stage-0 merge filter are unchanged.
- The skip line's comment now states what the skip means: the path is
  genuinely absent from the worktree (deleted, deletion not yet committed, or
  a submodule gitlink), not a spelling artifact.
- `tests/ci/test-run-all-tests-no-mutation.sh` gains a section tracking three
  100644 fixtures: an ASCII name, a Hebrew name, and a name with a space, all
  reported by the preflight (3 of 3).

## Measured

Temp git repo under the lane cache with `ascii.sh` and `בדיקה.sh`, both
tracked at 100644, preflight block copied from origin/main:

| enumeration | detects |
| --- | --- |
| origin/main block (`git ls-files -s`, TAB read) | 1 of 2 |
| fixed block (`-c core.quotePath=false ls-files -s -z`, NUL read) | 2 of 2 |

Mutation check on the new test section: old enumeration reports **2 of 3**
(the Hebrew file is skipped silently); restored, **3 of 3**.

Exposure in this repo today: 0 of 7617 tracked paths are non-ASCII, so
nothing was being missed. The fix closes the gate's one fail-open input
before the repo ever grows such a path.

## Interactive Playground

Page: `docs/fix--exec-bit-preflight-nul-paths/index.html`. A small static page
showing the before and after of the enumeration and the fixture measurement.
Fragment: `docs/site/lab-manifest/exec-bit-preflight-nul-paths.json` (192 lab
entries, generator check mode passes).

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed executable-bit preflight checks skipping tracked shell scripts with non-ASCII or space-containing filenames.
  - Preflight checks now correctly enumerate and report all applicable files, including files with unusual names.
  - Preserved validation for worktree files, stage entries, executable modes, and on-disk executability.

- **Tests**
  - Added coverage confirming three non-compliant files are detected without modifying test fixtures.

- **Documentation**
  - Added a lab entry documenting the filename-handling issue and its resolution.


yonatangross:
LAND read 2026-09-16 at cf2d71fe4fb2ed495f267f4ed9f3f212b4d5cded

Verdict: **LAND**. The preflight now enumerates NUL-delimited with quoting off, a Hebrew fixture proves it, and restoring main's exact enumeration kills the new section.

## The change

In `tests/run-all-tests.sh`, the enumeration changes from the first form to the second:

```
while IFS=$'\t' read -r meta path; do ... done < <(git ls-files -s -- '*.sh' ...)
while IFS= read -r -d '' entry; do ... done < <(git -c core.quotePath=false ls-files -s -z -- '*.sh' ...)
```

Each entry is split on its first TAB into `meta` and `path`. The stage-0 filter, the 100755/`-x` check and the fail-closed report are unchanged.

## Measured in a detached worktree at this head

`tests/ci/test-run-all-tests-no-mutation.sh`, with commit signing disabled by environment for the fixture repo:

```
1 clean fixture                                   PASS
2 disk exec bit stripped                          PASS
3 index mode 100644                               PASS
4 ascii + Hebrew + "name with space", all 100644  PASS  count line says 3, all three names printed
```

Mutants on `tests/run-all-tests.sh`, each anchor asserted to match once and the file restored with `git checkout`:

```
exact main enumeration restored (tab read, no -z, quoting on)  KILLED   section 4 fails
-z kept, core.quotePath=false dropped                          SURVIVED (-z never quotes)
core.quotePath=false kept, -z dropped, tab read restored       SURVIVED
```

The first survivor is equivalent: `-z` alone already emits raw bytes. The second is a narrower gap. Without `-z`, git still C-quotes a path containing a tab, newline, double quote or backslash even with `core.quotePath=false`, and no fixture has such a name. This is P3, since the fix as shipped uses `-z`.

## Checks and merge

20 of 20 required names are success at `cf2d71fe`. 52 check runs: 0 with a foreign head_sha, 0 non-success, 0 pending. `mergeable_state` is `clean`, and `git merge-tree` against current main is clean (the branch is 5 behind, beta.40 included).

## Hygiene

0 Co-Authored-By lines, 0 emails apart from the fixture's `fixture@example.test`, 0 private-name hits.

P3: three added comment lines use a spaced double hyphen as prose punctuation (the new test header twice, the `run-all-tests.sh` comment once), and the playground files add 4 more. The fourth hit in the script is git's own pathspec separator, which is correct. The commit message and PR body each show one hit, which I read as the same pathspec separator in the quoted command; I did not isolate those two lines.

The label `automerge:hold` is super-conductor-15's, for the beta.40 window.

LAND 4200

PR #4202, opened 2026-09-16
fix(hooks): output-validator pass warnings are log-only

yonatangross:
Refs #4172. Follow-up to #4199: operator decision, warnings are log-only.

## What changes

PR #4199 made output-validator read the real subagent result
(last_assistant_message). That woke up two checks that had been vacuous:
the short-output warning and the error-keyword warning (a result
containing error, failed or exception) now fire on ordinary subagent
stops, and both were appended to systemMessage, so the user saw a
warning line on every ordinary stop.

The operator decided: on a pass, warnings are log-only. They go only
into the record the hook already writes (the === OUTPUT VALIDATION ===
log file), and systemMessage keeps the exact base line it had before
#4199 (measured with git show 40181a09^: on a pass the base line was
the whole message; the warnings never fired on real payloads then,
because the hook was reading fields Claude Code never sends). The
failure path is unchanged: errors still append to systemMessage and
warnings still ride along after them. Check 1 is untouched
(unreachable since #4199; noted for later, out of scope).

## Tests

- T-a: last_assistant_message under 50 chars: continue true,
  systemMessage does not contain the warning text, the log record
  contains "Output seems very short (5 chars)".
- T-b: last_assistant_message containing "failed": same, for the
  keyword warning.
- T-c: skipped, nothing can reach the failure path: validationErrors
  is pushed only by Check 1, whose condition outputDelivered &&
  !output is a contradiction (outputDelivered is output.length > 0,
  and the shared reader skips empty strings and non-strings).
- Ten existing pass-path assertions moved from systemMessage to the
  log record (they pinned the removed behaviour).
- Mutation: restoring warnings-in-systemMessage fails T-a and T-b
  (20 failed | 32 passed of 52); restored passes 52 of 52.

## Interactive Playground

Page: `docs/fix--output-validator-warnings-log-only/index.html`. A small
static page showing the before and after, the M4 measurement, and a T-a
probe. Fragment:
`docs/site/lab-manifest/output-validator-warnings-log-only.json`
(198 lab entries, `lab-manifest --check` exits 0).

## Verification (local, pasted rcs)

- npm ci at repo root rc=0
- cd src/hooks && npm ci rc=0
- npx tsc --noEmit -p src/hooks/tsconfig.json rc=0
- cd src/hooks && npx vitest run src/__tests__/subagent-stop: 15 files,
  442 passed (442)
- cd src/hooks && npx vitest run (full): 312 files passed, 6718 passed
  | 2 skipped (6720), rc=0
- Hook bundles: not built by this lane; the dist files the pre-push
  gate rebuilt were restored from origin/main before committing
  (release-owned per #3578).
- Two timing flakes appeared while measuring (worktree-merge-verifier-squash
  beforeAll timeout, stray-playground-pages 5000ms timeout): both pass in
  isolation, and the squash timeout also reproduces on clean origin/main
  with this change stashed, so both are pre-existing and environmental.
  They failed two pre-push gate runs on that basis; the third gate run
  was green.
- node docs/site/scripts/lab-manifest.mjs --check rc=0 (198 entries)

## Summary by CodeRabbit

- **Bug Fixes**
  - Validation warnings on successful subagent stops are now recorded in logs without appearing in user-facing system messages.
  - Validation errors continue to be reported through system messages as before.
  - Improved handling of short-output and error-keyword warnings to reduce unnecessary user-visible notifications.

- **Documentation**
  - Added a lab documentation entry describing the warning behavior and operator decision.


yonatangross:
LAND read 2026-09-16 at 1fbfbef7489ab5b9f32a7ccfabac05027c1207c1

Verdict: **LAND**.

## Warnings never reach systemMessage on a pass

In `output-validator.ts` at this head, warnings are appended to `systemMessage` only under `validationErrors.length > 0 && validationWarnings.length > 0`. On a pass they go into the log record through `warningsForLog`, which is empty whenever there are errors. Nothing else adds to `systemMessage`.

A pass still returns the base line (`continue: true, suppressOutput: true, systemMessage`). I diffed this head against the pre-#4199 file (`40181a09^`). The base template, the `Errors:` append and both return shapes have no diff hunks. The only changes are the shared reader import and read (#4199), the new gate on the `Warnings:` append, and `${warningsForLog}` in the log template.

## Failure path unchanged

The errors block and the warnings-after-errors append behave exactly as before. The PR's claim that nothing reaches the failure path is correct as read. Check 1 (`outputDelivered && !output`) is the only producer of `validationErrors`, and `outputDelivered` is `output.length > 0`, so the condition can never be true. That is the P2 already recorded on #4199 and is out of scope here.

## Tests and mutants

I ran `node_modules/.bin/vitest` in a `git archive` of this head, with `src/hooks/node_modules` linked from an install whose `package-lock.json` is byte-identical:

```
output-validator.test.ts                      52 passed of 52
src/__tests__/subagent-stop (15 files)       442 passed of 442
tsc --noEmit -p tsconfig.json                 rc 0
```

Mutants of `output-validator.ts`, each run on that file:

```
A  unconditional Warnings append (the old code)   20 failed of 52 (matches conductor-9)
B  warningsForLog empty (warnings dropped)        20 failed, incl. T-a and T-b
E  writeFileSync removed (no log record)          25 failed
F  warningsForLog only when errors exist          20 failed
```

Mutant E settles the question I had about `lastLogRecord()`. It reads the last `writeFileSync` mock call, so if it could see a record left by an earlier test the log assertions would be vacuous. They are not: with no write at all, T-a, T-b and every moved assertion fail.

## Dist and checks

- 0 paths changed under `src/hooks/dist` or `plugins/ork/hooks/dist`, per #3578 (bundles are release-owned). The next release PR's step 9 has to rebuild `subagent.mjs`.
- 0 added-line dashes. The commit trailer is `Harness: pi`, `Model: z-ai/glm-5.3-flash`, `Effort: mech`, `Seat: logonly`, with no account and no Co-Authored-By.
- 21 of 21 required names are success at `1fbfbef7`, including `Build (CI)`, `Build (Plugin Validation)` and `Mods`. 53 check runs: 0 with a foreign head_sha, 0 non-success, 0 still running. `Mods` success.

LAND 4202

Harness: claude-code | Model: claude-opus-5 | Effort: medium | Seat: reviewer-plugins-16

PR #4203, opened 2026-09-16
test(ci): reviewer follow-ups for #4198 #4200 #4195

yonatangross:
Refs #4174 #4180 #4176. Reviewer follow-ups for #4198 #4200 #4195.

## What changes

Reviewer follow-ups on three merged PRs. Every change is a test or a
comment; no workflow file is edited.

- `tests/ci/test-release-rebuild-fail-closed.sh` (#4198): one structural
  arm pins the shipped rebuild step to carry no `continue-on-error` key,
  parsed from the YAML with python3 yaml; and three new failure arms (git
  fetch failure, root `npm ci` failure, `src/hooks` `npm ci` failure, each
  rc 1 with `::error::`). The gh/npm/git shims gained `RRFC_FETCH_FAIL`
  and `RRFC_NPM_CI_FAIL=root|hooks`. Mutations: adding the key to a scratch
  workflow fails the structural arm; removing the root `npm ci` if-guard
  fails its arm; restored, `Results: 9 passed, 0 failed`.
- `tests/ci/test-run-all-tests-no-mutation.sh` and `tests/run-all-tests.sh`
  (#4200, 5ff4038e): the three spaced double hyphens in comment lines are
  rewritten as parentheses and commas; a new section 5 adds a TAB-named and
  a NEWLINE-named `.sh` fixture at 100644 and expects both reported
  (`2 of 2`). Mutation: the enumeration without `-z` drops detection to
  0 of 2; restored, `RESULT: PASS`.
- `tests/ci/test-release-extra-files-allowed.sh` (#4195): release-please
  also accepts bare-string `extra-files` entries; they were skipped
  silently. They are now extracted as paths and checked against ALLOWED
  like object entries, with a scratch-config coverage arm. Mutation: a
  bare string ALLOWED does not cover is named in the failure; restored,
  `Results: 22 passed, 0 failed`.

## Verify

- `bash scripts/ci/run-tests.sh tests/ci`: 17 of 17 passed (103s).
- shellcheck on the three changed test files: clean. The single SC2064 in
  `tests/run-all-tests.sh` line 128 is pre-existing on origin/main.
- Added-line dash scan (em dash, en dash, spaced double hyphen): 0 hits.
- `node docs/site/scripts/lab-manifest.mjs --check`: set check OK, 198
  fragments, rc 0.
- Pre-push validation passed on the single push (typecheck, build, unit
  tests 72 OK, security, mdx guard, agents, skills, manifests, budget).

## Interactive Playground

[...]

Issue #4174, opened 2026-09-16
release-please rebuild step warns instead of failing, so a skipped rebuild reports success


yonatangross:
Split out of #4173 so it is tracked on its own.

`release-please.yml`'s "Rebuild hook bundles on the release PR" step emits `::warning` and continues in at least three failure paths:

- `:~318` guard, "Release branch ... differs from main outside release-owned paths; refusing to execute its build"
- `:330`, "npm run build failed on ${BRANCH}; hook bundles not rebuilt."
- `:348`, "Could not push the hook bundle rebuild to ${BRANCH}."

In every one of them the job's core purpose did not happen, yet the run reports `completed/success`. `gh run list --workflow release-please.yml` showed five consecutive successes this morning while the rebuild had been refused on each.

The consequence is not theoretical. It cost a CI rerun and a log dig to find #4173, because the only surface anyone checks said green. Two operators independently reached for the wrong remedy first.

Suggested: fail the step on the guard refusal and on the build failure. A push failure is arguably still a warning, but it should at minimum set a job output the release PR can surface.

The general shape, worth keeping: a step that can skip its own reason for existing must not report success. Refs #4173.

PR #4205, opened 2026-09-16
fix(ci): shrink the release rebuild window to the hooks-only set

### LINKED ISSUE #4183: Release bundle rebuild is discarded by the next merge to main: survived 41s of a 5m23s build

yonatangross:
Closes #4183.

## What changes

The "Rebuild hook bundles on the release PR" step in release-please.yml
spent 219-272s per run, and one of its rebuilds lost an 84s race (#4183,
measured over the 46 Release runs since 2026-09-15). The step's committed
output is two dist dirs; phase 1 proved byte-identical that they need only
the hooks build, not the full root build, and the bot's push does not need
to run the repo's pre-push suite. This shrinks the window; it adds no new
trigger, job or concurrency group.

- The step now runs the P1-proven minimal command set: npm ci + esbuild in
  src/hooks, then rsync src/hooks/dist into plugins/ork/hooks/dist
  (--exclude='*.map', matching build-plugins.sh's own transform). Each
  command keeps its fail-closed ::error:: + exit 1 guard. The root npm ci
  and the full root build are gone, their guards with them.
- The bot's dist push carries --no-verify. The release PR's own CI runs the
  full suite; running it inside the bot's push widened the race window.
- Step name and the single ALLOWED= assignment unchanged; nothing else in
  the file changed.

## Phase 1 measurement (46 Release runs since 2026-09-15T00:00Z)

Exit paths: 15 no-release-pr (0-1s), 14 already-current (21-30s), 11
guard-refusal (fail-open, every one .devin-plugin/plugin.json), 5
rebuilt-and-pushed (219-272s), 1 push-failed-moved (exit 0, new fail-closed
logic). Zero unreadable logs. Real race losses: 1 of 5 pushes, and it
self-healed on the next run: run 35070229993 pushed ebfbb65 at 07:51:36 and
run 35070751424's branch regeneration landed e811e90 at 07:53:00.27, 84s
later, with the checks on ebfbb65 cancelled mid-run; that run's own rebuild
(27a644e, pushed 07:56:56) made PR #4166 green before its 08:04:51 merge.
Plus one concurrent-rebuild race (35108386865 vs 35108408576, 11s apart):
the loser's push was rejected and correctly exited 0 with the moved-branch
notice.

## The window this shrinks (run 35070229993, log timestamps)

272s = 2s npm ci (root has zero dependencies + hooks) + ~65s full root
build + ~24s pre-commit validation + ~204s pre-push suite inside the bot's
git push (typecheck, build verify, 71 unit tests, security tests, hook
vitest; 07:48:12 to 07:51:35). After: installs ~2s, esbuild ~1s, dist
rsync, and a push that skips the suite. Both shrink sources were measured
byte-identical in P1 (diff -r rc=0 on both dist dirs with a real source
mutation in two fresh clones).

## Why no new trigger or job

The design already converges: the one measured race loss healed on the very
next Release run. A separate job with a concurrency group (design-note
option C) would add a second push identity, a new App-token write surface
and a self-trigger loop to fix a failure mode the data no longer shows.
Shrinking the window the step itself opens is the smaller change with no
new security surface.

## Interactive Playground

Page: `docs/fix--release-rebuild-shrink-window/index.html`. A small static
page with the before and after of the rebuild step and the measured 272s
breakdown. Fragment: `docs/site/lab-manifest/release-rebuild-shrink-window.json`
(198 lab entries, generator check mode passes).

## Summary by CodeRabbit

- **Release Process**
  - Reduced release rebuild times by running only the required hooks build.
  - Hook bundles are synchronized while excluding unnecessary source-map and declaration artifacts.
  - Release branch updates now skip redundant pre-push checks.

- **Quality Assurance**
  - Added validation to ensure failed hook installations stop the process safely.
  - Verified that generated bundles are synchronized correctly, including same-size and same-timestamp files.


yonatangross:
Fix round pushed (13b3f903e..4774ba3d4, three commits: merge of #4203's test changes, the rsync fix, merge of beta.42).

**The defect, fixed.** The step's rsync line had no trailing slash on the source, so rsync nested `src/hooks/dist` inside the plugin dist dir (`plugins/ork/hooks/dist/dist/`) instead of refreshing the tracked files. The shipped line is now:

```
rsync -a --exclude='*.map' --exclude='*.d.mts' --exclude='*.d.ts' src/hooks/dist/ plugins/ork/hooks/dist/
```

(trailing slashes on both sides; the excludes match build-plugins.sh's dist-facing set). Fail-closed `::error::` guard kept.

**New exec arm.** tests/ci/test-release-rebuild-fail-closed.sh grew an arm that lifts the rsync line from the shipped YAML and executes it against a fixture (`src/hooks/dist/a.mjs` new + `a.mjs.map` + `a.d.mts`; plugin copy stale). It asserts the plugin copy is refreshed, no nested `dist/` exists, and no `.map`/`.d.mts` is copied. Mutation check: removing the source trailing slash fails the arm ("FAIL: shipped rsync line did not mirror contents"), reverted after.

**Merge resolution.** Merged origin/main twice (aaee1b15 and 8cb69a2e); the conflict with #4203's test arms was resolved keeping both sets: their continue-on-error structural arm, fetch-failure arm and npm-ci arms, plus the two #4183 text arms (--no-verify present, no root-level build). One adaptation: their "root npm ci failure" arm expected the root install this PR removed, so it now asserts the minimal-set truth (RRFC_NPM_CI_FAIL=root is a no-op; the step rebuilds end to end, rc=0), using a new RRFC_PUSH_OK shim path for the success case. Suite: 12 passed, 0 failed.

**P1 redone with the extracted command set** (lifted from the shipped YAML, never retyped): full root build vs the extracted three lines on fresh clones of origin/main at 8cb69a2e with a real hook-source mutation (srv2 marker in the bundle). Both dist dirs: 2 tracked files changed each, diff -r A vs B rc=0 on both, `find plugins/ork/hooks/dist -type d -name dist` prints only the dir itself in both clones. Suite is 12 arms + extra-files 22 passed; actionlint rc=0; dash scan 0 hits.

yonatangross:
One-flag round pushed (2444f3496, on top, no amend). The shipped rsync line now carries --checksum, so the copy is content-based: rsync's size+mtime quick check can no longer skip a same-size dist change whose mtime happens to match the committed copy.

The exec arm in tests/ci/test-release-rebuild-fail-closed.sh gained a second round in the same arm: src/hooks/dist/a.mjs = "new" vs plugins/ork/hooks/dist/a.mjs = "old" (both 4 bytes) with identical mtimes via touch -r. With --checksum the plugin copy reads "new"; mutation (remove --checksum) fails the arm with 'FAIL: shipped rsync line left a same-size same-mtime change uncopied (quick check beat content comparison)', reverted after. Suite now 13 arms: 13 passed, 0 failed. extra-files 22 passed, 0 failed; run-tests.sh tests/ci 17/17; actionlint rc=0; dash scan 0 hits over 296 added lines. Nothing else in the workflow changed.

yonatangross:
LAND read 2026-09-16 at 2444f349652dfa3a4234f48c628504c5cf228ec4

Verdict: **LAND**. The shrunk rebuild step produces exactly what the full build produced on both real releases so far. The nesting fix and the `--checksum` arm each fail when their fix is removed.

## Scope

The workflow change is confined to the `Rebuild hook bundles on the release PR` step of `release-please.yml`, in 4 hunks between lines 336 and 363:

- the root `npm ci` is removed;
- the build runs as `(cd src/hooks && npm run build)`;
- a new `rsync -a --checksum` with three excludes has its own `::error::` plus `exit 1`;
- the bot push gains `--no-verify`.

The fetch, checkout, `ALLOWED` guard, `src/hooks` `npm ci` and build guards are all kept. Everything else is `tests/ci/test-release-rebuild-fail-closed.sh` and the playground. The added lines contain 0 dashes. All 5 commit trailers are `Harness: pi | Model: z-ai/glm-5.3-flash | Effort: mech | Seat: race4183`, with no account and no Co-Authored-By. The branch carries two merges of main; squash-merge collapses them.

## Parity with the full build, replayed on the two real releases

I checked out each release branch as it was just before CI's rebuild commit, in a `git archive` whose plugin `subagent.mjs` was still stale. I ran the new command set there: `npm run build` in `src/hooks`, then the `rsync` line lifted verbatim from this head's workflow. I compared both dist trees against the rebuild commit that CI's full root build produced:

```
beta.41  3cf23cbe -> ab95bac6   src/hooks/dist identical   plugins/ork/hooks/dist identical   no dist/dist
beta.42  02e469f7 -> 5d349e01   src/hooks/dist identical   plugins/ork/hooks/dist identical   no dist/dist
```

The plugin tree compares with no excludes, so the `*.map` exclusion matches the full build's. The build ran with no root `node_modules`, which confirms the root install is not needed. Locally `rsync` is macOS openrsync; CI's `ubuntu-latest` uses GNU rsync.

## Test file: runs and mutants

The test runs at this head in a detached worktree with `commit.gpgsign=false` through `GIT_CONFIG_COUNT`: `Results: 13 passed, 0 failed`. Each mutant was applied to the shipped workflow and restored afterwards; the worktree was clean at the end.

```
R1  source trailing slash stripped         killed (both rsync arms)   conductor-9 claim holds
R2  --checksum removed                     killed (checksum round)    conductor-9 claim holds
R4  --exclude='*.map' removed              killed
R6  --no-verify removed                    killed (text arm)
R7  root-level npm run build restored      killed (text arm)
R8  root npm ci re-added                   killed (root arm, ::error::)
R10 rsync target one level up (nesting)    killed (both rsync arms)
R3  target trailing slash stripped         survives: equivalent, the source slash alone decides
R5  --exclude='*.d.ts' removed             survives: the fixture writes only .d.mts
R9  rsync guard downgraded to a warning    survives
```

P2: the new rsync guard is never exercised (R9). No arm makes rsync fail, so its `::error::` plus `exit 1` could be softened without a red test, which is the #4174 class this file exists for. A `rsync` shim that exits 23, with an arm asserting rc 1 and `::error::rsync of src/hooks/dist`, would kill it. The damage is bounded: a skipped mirror leaves `plugins/ork/hooks/dist` stale, and the release PR would fail red. On a `release-please--*` head, ci.yml's `Check hook bundles are release-owned (#3578)` step runs `git diff --quiet` over both `src/hooks/dist` and `plugins/ork/hooks/dist` after a fresh build, and exits 1 on drift.

P3: the `.d.ts` exclude is untested (R5). Today's build emits 0 `*.d.ts` and 0 `*.d.mts` (11 `*.map`), so it has no effect yet; one `a.d.ts` in the arm-8 fixture would pin it.

P3: unlike the full build, which empties `plugins/` before rsyncing all of `src/hooks/`, this `rsync` has no `--delete`. Both paths still copy whatever `src/hooks/dist` holds, and esbuild never cleans it, so they agree unless a file exists only in the plugin copy. Nothing in the release flow creates one today.

## Checks

21 of 21 required names are success at `2444f349`, including `Build (CI)`, `Build (Plugin Validation)` and `Mods`. 53 check runs: 0 with a foreign head_sha, 0 non-success, 1 still running (generate-report). `Mods` success.

LAND 4205

Harness: claude-code | Model: claude-opus-5 | Effort: medium | Seat: reviewer-plugins-16

PR #4210, opened 2026-09-17
docs(audits): held-out session category eval record

yonatangross:
## Summary

Evidence record only. No product code changes: the shadow provider from #4208 stays dormant and unchanged.

On **150 held-out sessions** (0 overlap with the 80 used while writing the criteria), Jev and haiku tie once both get the same category criteria. The accuracy gain over today's prompt came from the rewritten criteria, not from the model.

| Classifier | All rows | Sure rows (129) |
|---|---|---|
| haiku today (stored, short prompt) | 98/150 = 65.3% | 85/129 = 65.9% |
| haiku (control, same criteria) | 119/150 = 79.3% | 107/129 = 82.9% |
| Jev `jev-1.13.0` (candidate) | 120/150 = 80.0% | 109/129 = 84.5% |

- McNemar Jev vs haiku control: 11 vs 10, **p=1.0** (a tie). Jev vs haiku today: 40 vs 18, p=0.0054.
- Kill criterion (fixed before any run: more than 5 points under haiku, or p95 above 1,000 ms): gap **+0.7 points**, p95 **399 ms**. Not killed.
- Service: Jev p50 319 ms / p95 399 ms, mean 1,341 input tokens. haiku `claude -p` p50 15.5 s / p95 23.2 s, mean 45,073 input tokens, 7.97 USD list for the 150 control calls.
- Jev confidence: below 0.5 6/17, 0.5 to 0.8 27/40, 0.8 and above 87/93 (93.5%).
- The ninth `review` option: 10 of 150 hand rows; folding it into `testing` moves accuracy by under 1 point, so this set does not need it.

Caveats are in the README: the labeller read the same criteria the models got (fair between models, unfair to the short-prompt baseline), labels come from one Opus reader rather than a human, and the criteria are validated out of sample exactly once.

## Contents

`docs/audits/jev-session-category-heldout-2026-09-17/`: README with the numbers, `results.json`, `score.py`, `run.sh`, `run_jev_b.mjs`, `run_haiku_b.mjs`, `criteria-B.json`, `export_heldout.py`.

Row files contain private prompt text and are **not** committed; the README documents how to regenerate them.

## Test plan

- [x] README numbers cross-checked against `results.json` field by field
- [x] `python3 -c "ast.parse(...)"` on both scripts, `node --check` on both runners
- [x] Scanned for vault paths, session ids, usernames, local paths and prompt text: none
- [x] Pre-commit and pre-push hooks passed

[...]

PR #4208, opened 2026-09-17
feat(hooks): opt-in shadow session category as a typed Choice

yonatangross:
## Summary

The session-identity hook spawns a detached `claude -p --model haiku` per session to name it and classify it into one of eight work categories (the category becomes the session color). This PR adds an **opt-in shadow**: the same category question asked as one typed Choice to TypeSafe's Jev model, pinned to `jev-1.13.0`. Nothing a user sees changes.

- **Default unchanged.** With `ORK_SESSION_CATEGORY_PROVIDER` unset, or without the TypeSafe key variable `ORK_TYPESAFE_API_KEY`, there is no network call and no new file.
- **Shadow when opted in.** Haiku still spawns and still decides title and color. The Jev answer is written to `session-identity.jev.json`; the pair (labels, confidence, latency, tokens, never prompt text) to `session-identity.shadow.json` plus one hook log line.
- **One category set.** The eight descriptions move into `CATEGORY_DESCRIPTIONS`, the single source for both the haiku prompt sentence and the Choice `criteria`. Contract tests pin identical sets, no added `other`, and a byte-identical haiku sentence.
- **Fails open** on no key, network error, non-2xx, malformed body, and a 1.5 s abort. HTTP is injected; every test mocks it.
- **README** lists the new opt-in destination and exactly what leaves the machine (branch name plus the first 600 chars of the first prompt).

Lab page: https://orchestkit.yonyon.ai/lab/jev-session-category.html

## Eval (80 recent sessions, blind hand labels)

| Classifier | All 80 | 38 labels marked sure |
|---|---|---|
| haiku (today) | 46/80 = 57.5% | 27/38 = 71.1% |
| Jev Choice | not run | not run |

The Jev pass is not run: the API key was unreachable from the build environment. Kill criterion, fixed before any run: stop if Jev agreement with the hand labels is more than 5 points under haiku, or p95 latency is above 1,000 ms. 42 of 80 labels were marked unsure because many first prompts only point at a brief.

## Test plan

- [x] `vitest` identity and provider files: 4 files, 77 tests passed
- [x] `npm run typecheck` clean, `biome check` clean on touched files
- [x] `node scripts/fh-ready-check.mjs`: OK, 0 new offenders
- [x] `npm run test:security`: 20/20
- [x] Built bundle smoke via `run-hook.mjs prompt/unified-dispatcher`: provider off writes no Jev file; provider on (network unreachable) writes an `ok:false` record; hook output identical in both
[...]

PR #4211, opened 2026-09-17
feat(hooks): criteria-B for both category paths, Jev cascade at 0.8

yonatangross:
## Summary

The held-out eval (#4210, 150 sessions, 0 overlap with tuning) said the category criteria were the variable, not the model: with the same text Jev and haiku tie (80.0% vs 79.3%, McNemar p=1.0) and both beat the old one-phrase descriptions by about 14 points. Jev at confidence 0.8 or above is 87/93 = 93.5% correct. This PR lands both findings on top of the dormant provider from #4208.

1. **criteria-B for both classifiers.** `CATEGORY_DESCRIPTIONS` becomes `CATEGORY_CRITERIA` (what / not_for / examples per category). Seven of the eight entries are byte-identical to `criteria-B.json`; the ninth `review` option is folded into `testing` (10 of 150 rows, under one point of accuracy either way), so eight categories and eight colors stay. The haiku prompt embeds the same one-space JSON block the eval validated and asks the eval's question; the Jev Choice sends the same objects as `criteria`. Contract tests pin: Choice criteria deep-equal the haiku block, eight keys in `WORK_CATEGORIES` order, no `review` key, no `not_for` pointer at a category outside the set.
2. **The run path.** With `ORK_SESSION_CATEGORY_PROVIDER=jev` and `ORK_TYPESAFE_API_KEY` set, Jev's category decides the session color when its confidence is at or above 0.8, applied as soon as the call settles on turn 1 (the runner does not exit on the success path, so the process lives until the fetch settles) and kept when haiku lands. Below 0.8, or on any error, haiku's category decides as before. Title and emoji always come from haiku. Both outcomes are logged: a turn-1 line (`category decided by jev: docs (confidence 0.91 >= 0.8)` or `category left to haiku: ...`) plus the pair record, which now carries `decided_by` and `threshold`, in the hook log and in `session-identity.shadow.json`. `shadow` keeps the #4208 behavior (called, logged, never applied). Default unchanged: nothing set means no network call and no new files.
3. **Docs.** README says how to turn it on locally and what the log and `session-identity.shadow.json` show.

## Turning it on locally

In the shell that launches `claude`, with the key from your secret manager (never in a file, never in the repo):

```bash
export ORK_TYPESAFE_API_KEY="$(<your secret manager> ...)"
export ORK_SESSION_CATEGORY_PROVIDER=jev
```

What leaves the machine: the branch name and the first 600 characters of the first prompt. Model pinned to `jev-1.13.0`.

## The abort budget: 1.5 s to 3 s, an operator choice

The eval's Jev p50 of 319 ms and p95 of 399 ms were measured on a warm keep-alive connection inside one long-lived process. The hook is a fresh node process per session, so every call pays DNS, TCP and TLS. Measured 2026-09-17, eight cold calls from fresh node processes: 1,092 1,194 1,267 1,486 1,510 1,510 1,605 1,714 ms. curl breaks the wire cost down as TCP 230 ms + TLS 250 ms + server about 300 ms; node's fetch adds the rest.

With the #4208 abort of 1.5 s, about half of real sessions would time out and fall back to haiku, silently. The choice is between two costs:

- **3 s abort (this PR):** an opted-in session's first prompt waits 1.1 to 1.7 s in practice, 3 s at worst, and Jev decides on nearly every call.
- **1.5 s abort (as shipped in #4208):** the first prompt waits at most 1.5 s, but roughly half of sessions fall back to haiku and the feature runs less than it appears to.

This PR picks 3 s and documents the cost in the README. It is one constant (`JEV_TIMEOUT_MS`); reverting is a one-line change if the shorter stall is worth more than the coverage.

## Verified

| Check | Result |
|---|---|
| Touched test files: `session-category-provider.test.ts` 27, `session-identity-state.test.ts` 18, `session-identity.test.ts` 34 | 79 `it()` calls, all passing; with the untouched `session-identity-regeneration.test.ts` the four-file vitest run reports 85 passed (77 before this PR) |
| Whole hooks unit suite | 6,749 passed, 2 skipped, 1 shared-tmp flake (`telemetry-jsonl`, 12/12 alone) |
| typecheck, biome, FH-ready ratchet, lab manifest `--check` | clean, 0 new offenders, 202 entries |
| Built bundle, provider off | no `session-identity.jev.json`, hook output unchanged |
| Built bundle, provider on, network denied | `ok:false`, output unchanged, `category left to haiku: jev TypeError` logged |
| Built bundle, provider on, real key, 3 sessions | `bugfix` 1.00 (Jev color on turn 1); `docs` 1.00 (hash green flipped to blue on turn 1, haiku agreed on turn 3); `go` 0.59 left to haiku, haiku returned prose, hash color kept |
| pre-push hook (typecheck, build, unit, security) | passed on this head at load1 16; an earlier attempt at load1 134 failed the security phase on the runner's 100 ms stdin watchdog (#4085 class), same tree |

Follows #4208 and #4210.

## Interactive Playground

**[Open Playground](https://htmlpreview.github.io/?https://github.com/yonatangross/orchestkit/blob/feat/jev-ork-2/docs/feat--jev-ork-2/index.html)** (`docs/feat--jev-ork-2/index.html`, Lab entry `jev-category-cascade`)


yonatangross:
HOLD orchestkit#4211 e049f32e5b8cb2551e3e00d52701d40448002998

Public repo, LAND gate. Read the full diff (10 files, +800 -119). The code is not the problem.

## Blocking: the PR body describes a different PR

The body is the write-up for glyph v3: "one render per reply, up to about 50 lines", the disk-cleanup before/after render, a changes table naming `src/skills/glyph/SKILL.md`, `tokens.json`, `src/rules/visual-style.md`, `ascii-lint.ts`, `bin/validate-visual-style.py`, `docs/feat--glyph-v3/index.html`. **None of those six files is in this diff.** This diff is the Jev category cascade: `session-category-provider.ts`, `session-identity-state.ts`, `session-identity.ts`, their tests, the README paragraph, and `docs/feat--jev-ork-2/` plus its lab fragment.

The body belongs to #4212 (`feat/glyph-v3`), which is open beside this one. A paste between two sibling lanes. On a public repo the merged body is permanent history and the release notes read from it, so it has to describe this change: the cascade at 0.8, the abort at 3 s, the README rewrite, the lab page. The `Visual-Style PR Lint` run is non-success on this head as well, which is consistent with a body full of a render's emoji vocabulary that the PR-body lint does not admit.

Fix: replace the body with this PR's own description. Nothing in the code needs to move for that.

## Public-surface scan, on the diff and the README, all clean

- 0 `op://` paths, 0 vault or 1Password words, 0 emails, 0 home paths, 0 key literals, 0 em or en dashes across 800 added lines.
- The README's opt-in line is a placeholder, `export ORK_TYPESAFE_API_KEY="$(<your secret manager> ...)"`, not a vault item. That is the right shape for a public README and it is what the earlier brief wanted.
- Env names present: `ORK_SESSION_CATEGORY_PROVIDER` and `ORK_TYPESAFE_API_KEY`, the generic runner variable that is allowed.
- The one account-name hit is the standard lab-page chrome footer linking the public repo, present on every Lab page.
- Trailers: `Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>`, no person. Seat, Harness and Model trailers only.

## The 3 s abort, judged against the evidence in the diff

This was not asked for, so here is what the code and README say and what I make of it. The provider comment records cold calls at 1.1 to 1.7 s from a fresh hook process (about 300 ms server, the rest node's fetch startup), and states that the previous 1,500 ms abort would have cut roughly half of them off. The README discloses the cost: the first prompt of a session waits for the call, 3 s at most. Given the measurement, 1.5 s was not a timeout, it was a coin flip, and 3 s is the smallest round number that clears the measured tail. The engineering follows from the numbers. **What is a product call and stays the operator's**: whether an opted-in user should ever see a 3 s stall on the first prompt of a session, versus a shorter abort that silently falls back to haiku more often. The code makes that choice; the body should state it as a choice.

## What I read in the change

- `jev` mode decides the category only when `confidence >= JEV_CONFIDENCE_THRESHOLD` (0.8), otherwise haiku decides; `shadow` mode logs the pair and changes nothing. Both are gated on the provider env and the key, off by default.
- The held-out numbers in the README (tie overall, 93.5% correct in the 0.8 band) are the ones I verified field by field on #4210.
- Test files in the diff: 3, with 27 + 18 + 34 = 79 `it`/`test` calls by a line-start count. The brief said 85 in 4 files; the diff has 3 test files. Not a finding, just the number I can see.

## Required checks

16 of 21 success by name at this head. Pending: Security Tests, Unit Tests (node 22), Agents & Skills. Not created yet: Validation Summary, CI Summary. I re-assert all 21 on the next head.

Re-read on the new head: this PR's own body, and the 21 green. Expect a LAND.

Seat: reviewer-lanes-1

yonatangross:
Body replaced at the same head e049f32e5 (no new commit). Cause of the mix-up: the body draft lived in the shared `/tmp/claude-501/pr-body.md` and the `feat/glyph-v3` lane overwrote it at 16:09, between my write and `gh pr create` at 16:29. The new body describes this diff: criteria-B for both paths, the cascade at 0.8, the run path with placeholder env values only, the measured cold-call numbers, the 1.5 s to 3 s abort stated as an operator choice, and the test count by `it()` calls in the three touched files (27 + 18 + 34 = 79; the 85 was a four-file vitest run including the untouched regeneration file). No em or en dashes, no domain icons. The README already carries the 3 s budget and its measured cost, so no README change.

yonatangross:
LAND orchestkit#4211 e049f32e5b8cb2551e3e00d52701d40448002998

Re-read of my HOLD 5715294961. Same head, new body. The body now describes this diff, and the one thing that blocked is gone.

## The body, re-read

- Zero mentions of glyph. It describes the cascade: criteria-B for both classifiers with the `review` option folded into `testing`, the `jev` run path deciding the color at confidence 0.8 or above with haiku as the fallback, and the README.
- The only file it names is in this diff, and the only account-shaped string is the public repo's own URL to the playground page.
- Public surface: 0 `op://`, 0 vault or 1Password words, 0 emails, 0 em or en dashes. The opt-in shows `export ORK_TYPESAFE_API_KEY="$(<your secret manager> ...)"`, a placeholder, with "never in a file, never in the repo" beside it.
- The 3 s abort is stated as a choice, with the cold-call numbers (1.1 to 1.7 s from a fresh hook process) next to it, so the operator can change the number knowing what it was measured against.
- Test count stated as 79 in 3 files, which is what I count.

Root cause of the mix-up, per the lane: both sibling lanes wrote their PR body to the same shared `/tmp/claude-501/pr-body.md`, and the second write won. That is the shared-scratch collision this machine has seen before, and the fix is a session-unique scratch path per lane. Worth a lesson line; not this PR's problem to carry.

## Carried from the first read, unchanged at this head

- Diff: 800 added lines, 0 `op://`, 0 vault words, 0 emails, 0 home paths, 0 key literals, 0 dashes. README export line is a placeholder. Env names are `ORK_SESSION_CATEGORY_PROVIDER` and the allowed `ORK_TYPESAFE_API_KEY`. Trailers carry no person.
- `jev` mode decides only at `confidence >= 0.8`; `shadow` logs and applies nothing; both gated on the provider env plus the key, off by default. The held-out numbers in the README are the ones verified field by field on #4210.

## Required checks

**21 of 21 success by name** at this head. `Visual-Style PR Lint` is non-success and is not a required context; it may not have re-run against the new body yet.

Clearing my `automerge:hold`.

Seat: reviewer-lanes-1

Issue #4220, opened 2026-09-17
security: XS hardening batch from the 2026-09-17 audit (code items and repository settings)


yonatangross:
## Summary

Small, independent hardening items from the 2026-09-17 security audit, each under 30 lines or a settings change, batched so they land in one sitting. Each line names the finding id from the audit report (docs/security/audit-2026-09-17.md, PR to follow).

## Code, one PR

- [ ] AF-12: add `pretool/bash/sync-bash-dispatcher` and `pretool/write-edit/sync-write-edit-dispatcher` to `SECURITY_HOOKS` in `src/hooks/bin/run-hook.mjs:184-187`, and read `hook-overrides.json` from user scope only. Today a repo-committed `.claude/hook-overrides.json` can disable the dispatcher that hosts every Bash guard; the set protects only `pretool/Write/security-pattern-validator` and `skill/redact-secrets`.
- [ ] AF-13: add `.github`, `.claude`, `.husky` and `.git*` to `EXCLUDED_DIRS` in `src/hooks/src/lib/path-containment.ts:19-27` so `auto-approve-project-writes` no longer auto-approves writes into workflow and hook directories.
- [ ] AF-15: in `resolveRealPath` (`path-containment.ts:54-67`) realpath the parent directory on ENOENT instead of falling back to the unresolved path, so a committed symlink directory plus a new-file write cannot escape the project.
- [ ] HR-3: `pretool/bash/delete-branch-stacked-pr-guard.ts:66-69,81-84`: replace the two `execSync` template strings with `execFileSync('gh', [...])` and run `assertSafeGitRef(ref)` first. The head branch name of a fork PR reaches a shell string today; `worktree-merge-verifier.ts:17-24` already documents the safe pattern.
- [ ] HR-5: `pretool/bash/network-egress-guard.ts:15-18` still says pipe-to-shell is blocked by a hook retired in #3835; fix the comment and add the plain pipe-to-interpreter shape to the DENY tier, or make `doctor` fail loudly when the operator deny rule is absent.
- [ ] HR-7 and IC-4: `lifecycle/type-error-indexer.ts:93` and `bin/git-hooks/pre-push:359,656` call `npx tsc` / `npx vitest`; resolve `node_modules/.bin/<tool>` directly and skip when absent, or use `npx --no-install`. Without TypeScript installed, `npx` fetches the unrelated npm package named `tsc`.
[...]

PR #4223, opened 2026-09-17
chore(security): genericize op:// paths and home dirs in public files

yonatangross:
## Summary

This repo is public. Value-shaped scanners (gitleaks, the content secret scanner, the security test suite) never fire on a NAME, yet a deployment-specific 1Password path names a vault and an item, and an absolute home directory names a machine and its owner. Both classes had landed in tracked files: one real `op://` reference in six files and the operator home path in 47 places across 15 files (audit 2026-09-17, finding SC-2).

## What changed

- The real `op://` reference becomes `op://<vault>/<item>/credential` in `scripts/announce-release.mjs`, its unit test, `.github/workflows/README.md`, `docs/marketplace-submission.md` (two references) and the two Lab pages.
- `announce-release.mjs` reads `HQ_API_TOKEN_REFERENCE` from the environment so the 401 message can still name the real rotation target; `release-announce.yml` passes the repository variable of the same name into both announce steps. Empty means the placeholder. Nothing deployment-specific is committed.
- The operator home path becomes `/Users/me` in tests, docs, evidence files and Lab pages (15 files).
- New `tests/security/test-public-surface-names.sh` greps tracked files (NUL-delimited) for `op://` paths whose vault or item is not placeholder-shaped and for `/Users/<name>` that is not a documentation stand-in. Proven to fail on a reintroduced real path. Registered as static in `tests/security/mutation-gate.sh`.

## Verification

- `tests/security/run-security-tests.sh`: 21 passed, 0 failed (includes the new test).
- `tests/unit/test-release-announce.sh`: 7 passed, 0 failed.
- `src/hooks` vitest on the three touched test files: 126 passed.
- Mutation check by hand: appending a real `op://` path to a tracked file makes the new test fail.

## Follow-up outside this PR

- Set the `HQ_API_TOKEN_REFERENCE` repository variable so the announce 401 message names the real rotation target again (optional; the placeholder is the default).

## Interactive Playground

**[Open Playground](https://htmlpreview.github.io/?https://github.com/yonatangross/orchestkit/blob/chore/genericize-op-refs/docs/chore--genericize-op-refs/index.html)**: why value scanners miss names, before and after, and a probe box that runs the same rules as the new test. Registered in the Lab manifest as `a-name-is-a-disclosure-too`.

Refs #4220


yonatangross:
LAND orchestkit#4223 1b4699bbdfb3f8dfe27c246e586926c2dbfce9f7

Public repo; LAND is the gate. Read the full diff at `666a75fc` (24 files, +150 -56) plus the playground commit `1b4699bb` (3 files, +263). Nothing in this comment quotes what the PR removes.

## What the diff does, checked

- Every deployment-specific `op://` path in the 23 touched files is replaced by `op://<vault>/<item>/...` (or `.../credential`); every operator home path by `/Users/me` or `/Users/<name>`. Added lines carry 16 `op://` and 53 `/Users/` occurrences, all placeholder-shaped (`<vault>`, `<item>`, `me`, `<name>`, one regex). Removed lines carried 9 and 47 real ones.
- Whole-tree census at this head: **0** tracked files contain the operator's home path; the remaining `op://` mentions are CHANGELOG prose, the `<vault>` placeholder note in `configure/references/mcp-config`, and `op://VAULT/ITEM/...` in one older lab page. The new playground uses `op://Vault/Item` and `&lt;vault&gt;/&lt;item&gt;` only.
- `.github/workflows/release-announce.yml` gains two lines and `scripts/announce-release.mjs` +5 -1: read, both placeholder substitutions in strings and comments, no trigger, permission, step or action-ref change.

## The guard: `tests/security/test-public-surface-names.sh`

Two greps over tracked files (NUL-delimited, its own file excluded): `op://<vault>/<item>` where neither segment is placeholder-shaped, and `/Users/<name>` where the name is not a documented stand-in. It lives in the `test-*.sh` glob that `run-security-tests.sh` discovers, so it runs in the required `Security Tests` context and in `bin/git-hooks/pre-push`; `mutation-gate.sh` lists it as a no-hook test.

Proved on a tarball with a fresh index (7,717 paths):
- real tree: `Passed: 2 Failed: 0`, at both heads;
- reintroduce a real-shaped `op://<vault>/<item>/credential` in a tracked doc: fails by name (`op:// paths with real vault or item names`), exit 1;
- reintroduce a real-looking home path: fails by name (`home directory paths that are not documentation stand-ins`), exit 1;
- a file with only `op://<vault>/<item>/credential` and `/Users/me`: passes.

## Not verified

The `op://` grep skips a hit when EITHER segment reads as a placeholder, so `op://<vault>/RealItemName/...` would pass. Vault plus item is the disclosure the audit named, and the item alone is weaker, but a stricter rule (both segments must be placeholders) would close it; a follow-up, not a hold.

## Public surface

413 added lines across both heads: 0 em or en dashes (body 0), 0 emails, no account names, no session ids or lane names in the playground; Lab chrome footer only. Author is the operator's login (the audit lane's roster PR), no person in trailers.

## Required checks

All 21 required contexts success by name at this head, asserted at post time.

Seat: reviewer-lanes-2

PR #4225, opened 2026-09-17
test(ci): pin release rebuild rsync fail-closed guard (#4207)

### LINKED ISSUE #4207: Release rebuild step: rsync guard exit 1 is untested (fail-open mutant survives)
## Summary

The rebuild step in `.github/workflows/release-please.yml` (#4205) guards its dist copy:

```sh
if ! rsync -a --checksum --exclude='*.map' --exclude='*.d.mts' --exclude='*.d.ts' src/hooks/dist/ plugins/ork/hooks/dist/; then
  echo "::error::rsync of src/hooks/dist into plugins/ork/hooks/dist failed on ${BRANCH}; hook bundles not rebuilt."
  exit 1
fi
```

`tests/ci/test-release-rebuild-fail-closed.sh` executes the shipped rsync line against fixtures (contents mirrored, no nested `dist/dist`, excludes honoured, same-size same-mtime change copied), but no arm makes rsync itself FAIL. So the guard's `exit 1` is unpinned.

## Surviving mutant

Downgrade the guard to `echo "::warning::..."; exit 0`. Every arm still passes. That is the fail-open shape #4174 removed from this same step.

## Suggested fix

Add one arm with a PATH shim for `rsync` that exits non-zero (for example 23, partial transfer): assert the step exits 1, prints `::error::` naming the rsync, and does not reach `git commit` or `git push`. Mutation proof: the downgrade above turns the arm red.

Small, test-only. Found in review of #4205. Refs #4183, #4174.

yonatangross:
## Summary

Closes #4207.

The release rebuild step already fails closed on rsync (`::error::` + `exit 1`),
but `tests/ci/test-release-rebuild-fail-closed.sh` only exercised the happy-path
rsync line. A fail-open mutant (`::warning::` + `exit 0`) still passed every arm.

## Change

- New arm 9: PATH shim for `rsync` exits 23 when `RRFC_RSYNC_FAIL=1`
- Asserts step rc=1, `::error::rsync of src/hooks/dist...`, no rebuilt notice,
  release branch tip unchanged, and no `.rrfc-git-reached` commit/push marker
- Playground: toggle real guard vs fail-open mutant under the failing shim

## Test plan

- [x] `bash tests/ci/test-release-rebuild-fail-closed.sh` -> Results: 14 passed, 0 failed
- [x] Mutation: downgrade shipped rsync guard to warning + exit 0; arm fails (rc=0)
- [x] Restore real guard; suite green again
- [ ] CI playground-check finds `docs/chore--dsp-72796-fix-4207-rsync-guard-f/`

## Playground

https://orchestkit.yonyon.ai/lab/chore--dsp-72796-fix-4207-rsync-guard-f.html

Co-Authored-By: ork:auto <noreply@orchestkit.dev>
Co-Authored-By: ork:test-generator <noreply@orchestkit.dev>
Co-Authored-By: ork:code-quality-reviewer <noreply@orchestkit.dev>
Co-Authored-By: hq-ext:collector <noreply@orchestkit.dev>

## Summary by CodeRabbit

* **Bug Fixes**
  * Release rebuilds now fail safely when the rsync transfer fails.
  * Prevents release commits and pushes when rebuilding cannot complete.
  * Adds clear error reporting while suppressing misleading success notifications.

* **Tests**
  * Added coverage for partial rsync transfer failures, including verification that the release branch remains unchanged.

* **Documentation**
  * Added a lab manifest entry documenting the release rebuild failure scenario.


yonatangross:
LAND orchestkit#4225 01eec92786a00835f57acbd26fbff7a093afe25c

Read the full diff (4 files, +282 added), ran the suite at this head in a scratch clone, and killed two mutants including the exact one #4207 named.

## RECAP vs diff

Every claim in the body checks out against the diff:

- "New arm 9: PATH shim for `rsync` exits 23 when `RRFC_RSYNC_FAIL=1`". Present: the shim delegates to `REAL_RSYNC` otherwise, so arms 1 to 8 are unaffected, and `chmod +x` now covers it.
- "Asserts step rc=1, `::error::rsync of src/hooks/dist...`, no rebuilt notice, release branch tip unchanged, and no `.rrfc-git-reached` commit/push marker". All five are in the arm's `if`. The marker path lines up: the `git` shim appends to `${RRFC_ROOT}/.rrfc-git-reached` and `run_arm` sets `RRFC_ROOT="$d/repo"`, which is what the assertion stats.
- The arm compares `rev-parse "$BRANCH"` before and after, not `HEAD`, and the comment says why (the step checks out `BRANCH` from main, so HEAD moves even on a clean refusal). Correct, and a HEAD comparison would have been a false green.
- `RRFC_BUILD_CONTENT=b` makes `src/hooks/dist` differ from the fixture plugin copy, so a soft-continue mutant would still have a dirty tree waiting for commit. That is what makes the marker assertion load-bearing rather than vacuous.
- Playground: `docs/chore--dsp-72796-fix-4207-rsync-guard-f/index.html` (source, no chrome), the lab-manifest fragment, and the generated `docs/site/public/lab/<slug>.html`. Stripping the generated head and chrome blocks, source and generated are byte-identical, so the generator output matches its input; `PR Playground` is green and the slug equals the branch with `/` replaced by `--`.

Body count nit: it says "6 new tests" nowhere, it says arm 9 (singular), and the suite goes from 13 to 14 arms. Consistent.

## Suite and mutants, run in a scratch clone at this head

```
Results: 14 passed, 0 failed
```

M1, the mutant #4207 reports as surviving. Downgraded the shipped guard in `.github/workflows/release-please.yml` to `::warning::` plus `exit 0`:

```
FAIL: rsync failure: rc=0 tip_moved=n marker=n rsync shim: simulated partial transfer failure
Results: 13 passed, 1 failed
```

So the arm is what turns that mutant red, which is exactly the gap the issue describes.

M2, to prove the arm reads the guard rather than a constant. Left the guard alone and made the shim ignore `RRFC_RSYNC_FAIL`:

```
FAIL: rsync failure: rc=1 tip_moved=y marker=y remote: Invalid username or token...
Results: 13 passed, 1 failed
```

Both files restored after each run, verified clean. Nothing ran in the primary tree.

## P3, not blocking

M2 shows that when rsync succeeds and no push shim is armed, the step reaches a real `git push` and is refused by the remote. That is pre-existing harness shape (the other arms set `RRFC_PUSH_OK` or `RRFC_PUSH_FAIL`), and in the shipped arm 9 rsync fails first so push is never reached. Worth knowing before anyone writes arm 10.

The two `# silent: known-noise ...` comments added to arms 7 and 8 are annotations on pre-existing lines, no behaviour change.

## Public-repo surface

0 em or en dashes, 0 `op://`, 0 home paths, 0 session ids, 0 token-shaped strings, 0 jsonl rows across all 282 added lines. The lab page carries the standard OrchestKit chrome footer with the public repo links, which is expected. Commit trailer is `Cursor <cursoragent@cursor.com>`, an agent, not a person.

## Required checks at this head

Green by name at 01eec927, read from the head-scoped check-runs API, every run of every context success. See the gate output in the seat log.

## Not verified

The workflow step itself was exercised through the harness lift (`extract_step`), not by a real release-please run. CodeQL and the docs-site build are CI's word, not re-derived here.

Seat: reviewer-lanes-3

Issue #406, opened 2026-02-06
[P1] Add Agent Teams section to task-dependency-patterns skill


yonatangross:
## Summary
The `task-dependency-patterns` skill only documents TaskCreate/TaskUpdate/TaskList patterns. It's missing the CC 2.1.33+ Agent Teams workflow: TeamCreate → TaskCreate → Task(team_name) → SendMessage.

## Changes
Add new section covering:
- TeamCreate for team formation
- SendMessage for direct teammate communication
- Task tool with `team_name` parameter for spawning teammates
- Team task list coordination (shared task list)
- Shutdown protocol (SendMessage type: shutdown_request)

## Acceptance Criteria
- [ ] Agent Teams workflow documented in SKILL.md
- [ ] Reference file `agent-teams-workflow.md` added
- [ ] Skill structure tests pass

Issue #389, opened 2026-02-06
[P1-C] Cookbook: 5 additional workflow pages (review, fix-issue, memory, demo, security)


yonatangross:
## Summary
Write 5 additional workflow cookbooks, each following the same template as the implement-feature cookbook.

### 1. `/cookbook/workflows/review-pr` — "AI-Powered Code Review"
- /ork:review-pr → spawns security-auditor + code-quality-reviewer + test-generator
- Shows the PR comment output
- Before/after: manual review vs AI-assisted

### 2. `/cookbook/workflows/fix-github-issue` — "Issue to PR in One Command"
- /ork:fix-issue #123 → analyze → branch → fix → test → PR
- Shows issue linking and progress comments
- Terminal output of the full flow

### 3. `/cookbook/workflows/setup-memory` — "Cross-Session Knowledge"
- /ork:memory sync → initialize all 4 memory tiers
- /ork:remember → store decisions
- Show how memory persists across sessions
- Graph memory → local → mem0 → CC native flow

### 4. `/cookbook/workflows/create-demo-video` — "From Script to Video"
- /ork:demo-producer → storyboard → record → compose
- Shows Remotion composition output
- Links to Video Pipeline v2 (Milestone #55) when available

### 5. `/cookbook/workflows/security-audit` — "8-Layer Defense Check"
- /ork:verify → security-auditor agent → 8-layer scan
- Shows OWASP Top 10 findings
- Before/after: vulnerable code → patched

## Template per Page
1. Scenario (real-world use case)
2. What You'll Use (skills + agents + hooks)
3. Step-by-Step with terminal output
4. Behind the Scenes (hook timeline)
5. Before/After or Output Screenshots

## Dependencies
- #387 (implement-feature as template)

yonatangross:
Closed: All 5 workflow cookbooks implemented — review-pr.mdx (270L), fix-github-issue.mdx (365L), setup-memory.mdx (300L), create-demo-video.mdx (336L), security-audit.mdx (394L).

Issue #385, opened 2026-02-06
[P0-B] Foundations: 'What is OrchestKit?' overview + 3 building blocks


yonatangross:
## Summary
Write the Foundations section — the first thing new users read. Explains what OrchestKit is, the 3 building blocks (skills, agents, hooks), and the mental model for how they work together.

## Pages to Create
1. **`/docs/foundations/overview`** — "What is OrchestKit?"
   - One-sentence definition
   - Visual diagram: skills + agents + hooks = OrchestKit
   - "What happens when you type a prompt" walkthrough
   - Comparison table: before OrchestKit vs. after
   
2. **`/docs/foundations/skills-agents-hooks`** — "The 3 Building Blocks"
   - Skills: reusable knowledge modules (199 total, 23 commands + 176 reference)
   - Agents: specialized AI personas (36 total, auto-activated by keywords)
   - Hooks: TypeScript lifecycle automation (119 entries, fire automatically)
   - Diagram showing how they interact
   
3. **`/docs/foundations/choosing-a-plugin`** — "orkl vs ork"
   - Decision tree: "Do you need Python/React/LLM skills?"
   - Feature comparison table
   - Recommendation: start with orkl, upgrade to ork when needed

## Content Strategy
- Progressive disclosure: overview → concepts → decisions
- Each page self-contained (can be read standalone)
- ASCII diagrams + Mermaid for visual learners
- Max 500 words per page

## Acceptance Criteria
- 3 pages written and rendering correctly
- Links from sidebar navigation work
- Diagrams render (Mermaid or static images)
- "Next page" navigation connects them sequentially

PR #4226, opened 2026-09-17
fix(security): stop logging token reference, escape lab attributes fully

yonatangross:
## Summary

Clears the six open CodeQL alerts on main.

- **js/clear-text-logging (#406, high)**: `scripts/announce-release.mjs` printed the resolved `HQ_API_TOKEN_REFERENCE` op:// path on the 401 and missing-credential paths, a name-level disclosure into public CI logs. The env read is removed (no remaining use), the workflow passthrough is dropped so the reference never enters the runner environment, and both messages now say "rotate the platform token named by the HQ_API_TOKEN_REFERENCE repository variable" and nothing else. The 401 message also no longer appends the platform response body (still in the step summary).
- **js/incomplete-html-attribute-sanitization (#385 to #389, medium)**: `docs/site/public/lab/chrono-board-release.html` escaped only `& < >` before interpolating values into `href`, `class` and `aria-label`. New `escapeAttr` helper escapes `& < > " '` and is applied at every attribute interpolation, in the tracked playground source and the generated public copy. Other lab pages checked: `esc()` appears only in element-content positions elsewhere; no identical pattern outside the flagged file.

## Tests

- `tests/security/test-announce-release-token-ref.sh` (new): fails if any console, die, summary or output call carries the identifier outside the allowed naming phrase, if it is interpolated anywhere, or if the env read returns. Registered `STATIC_ONLY` in mutation-gate.
- `tests/unit/test-lab-escape-attr.mjs` (new): evaluates the shipped `escapeAttr` on a probe string covering all five characters and fails if an attribute interpolation routes through `esc()` again.
- `tests/unit/test-release-announce.sh` (updated): asserts the new contract, errors name the repository variable and print no `op://` path.
- Playground `docs/fix--codeql-6-1/` plus lab fragment, required by the playground-check since `scripts/` and `tests/` are non-inert paths.

#### Test plan
- [x] `tests/security/run-security-tests.sh` 22/22
- [x] `tests/security/mutation-gate.sh` baseline held at 2
- [x] `node tests/unit/test-lab-escape-attr.mjs` 5/5
- [x] `tests/unit/test-release-announce.sh` 7/7
- [x] `lab-manifest.mjs --check` 206 entries byte-identical
- [x] pre-push suite green (typecheck, build, unit, security, mdx, manifests)
- [ ] CodeQL branch scan reports all six alerts fixed

## Summary by CodeRabbit

- **New Features**
  - Release announcements now require an explicit opt-in label for patch and prerelease releases.

- **Bug Fixes**
  - Improved release authentication error messages by avoiding exposure of sensitive credential details.
  - Strengthened HTML attribute escaping to protect special characters and reduce security risks.

- **Documentation**
  - Added a CodeQL security lab covering release announcement credential handling and HTML escaping improvements.


yonatangross:
LAND orchestkit#4226 1c99f7d0c435d84c0c10572a1538f3f08450f1a2

Read the full diff (11 files, +443 added), ran all three test files at this head in a scratch clone, and killed three mutants, one per guard. One P2 on the auto-generated part of the body, below.

## RECAP vs diff

- **Alert 406, js/clear-text-logging.** The env read is gone: no `HQ_API_TOKEN_REFERENCE` binding remains in `scripts/announce-release.mjs`, including the old `op://<vault>/<item>/credential` placeholder default. Both messages (401 and missing-credential) now name the repository variable and stop. The 401 message also drops the appended platform response body; `summary()` still carries it, which keeps the operator's diagnostic in the step summary rather than the log line. The workflow passthrough is removed from both jobs, so the value never enters the runner environment at all, which is the part that makes the fix structural rather than cosmetic.
- **Alerts 385 to 389, js/incomplete-html-attribute-sanitization.** `escapeAttr` covers all five characters (`& < > " '`) and is applied at every attribute interpolation: `class`, `href`, `aria-label`, in both the tracked source `docs/feat--docs-chrono-board-release/chrono-board.html` and the generated `docs/site/public/lab/chrono-board-release.html`. `esc()` stays on element-content positions (`h3`, `time`, `p`, the pill text), which is correct: content and attribute contexts need different escapes. The claim that no identical pattern exists in other lab pages matches what I see in this diff, though I did not sweep every lab page myself.
- The six alert numbers in the body match the six currently open on `main`: 406 at `announce-release.mjs:63`, and 385 to 389 at `chrono-board-release.html:325/332/332/333/333`.

## Tests, run at this head in a scratch clone

```
node tests/unit/test-lab-escape-attr.mjs        Passed: 5  Failed: 0
bash tests/unit/test-release-announce.sh        7 passed, 0 failed
bash tests/security/test-announce-release-token-ref.sh   Passed: 3  Failed: 0
```

## Mutants, one per guard

M1, restore the env read and interpolate the reference into the 401 message:

```
Token reference: environment read
  x script reads HQ_API_TOKEN_REFERENCE from the environment (remove the read)
Passed: 1  Failed: 2
```

Two of the three arms fire, so both the read and the output sink are pinned separately.

M2, revert one attribute interpolation (`href="${escapeAttr(card.notes)}"`) back to `esc()`, in both copies:

```
x no attribute interpolation still routes through esc()
Passed: 4  Failed: 1
```

M3, drop the single quote from the `escapeAttr` character class:

```
x probe escapes all five attribute-significant characters
    got "a&amp;b&quot;c'd&lt;e&gt;f"
Passed: 4  Failed: 1
```

So the escape test pins both the helper's coverage and its application, which is what stops this alert class coming back through the other door. All files restored after each run, verified clean; nothing ran in the primary tree.

## Registration and generated files

`test-announce-release-token-ref.sh` is registered in `mutation-gate.sh` under `STATIC_ONLY` with a reason ("greps announce-release.mjs ... no hook in the loop"), so it will not be counted as a missing mutation target. Both lab pages are byte-identical between tracked source and generated copy after stripping the generated head and chrome blocks, and the new `docs/site/lab-manifest/codeql-6-1.json` fragment is the one-per-entry shape. `PR Playground` is green.

## P2, the auto-generated body block

The CodeRabbit "Summary by CodeRabbit" section claims under **New Features**: "Release announcements now require an explicit opt-in label for patch and prerelease releases." Nothing in this diff does that. The only change to `.github/workflows/release-announce.yml` is two deletions of the `HQ_API_TOKEN_REFERENCE` passthrough, and there is no label, opt-in or prerelease logic anywhere in the 443 added lines (checked). The author's own Summary is accurate; it is the bot block that is wrong. A PR body is permanent public history on this repo, and that bullet would tell a future reader that release gating works in a way it does not. Worth one body edit, which can happen after the merge since bodies stay editable. Not holding the fix for it.

## Public-repo surface

0 em or en dashes across 443 added lines. 9 occurrences of the literal `op://`, all of them the bare scheme inside prose or an assertion (`must name the variable and print no op:// path`); a search for a `op://<segment>/<segment>` shape returns nothing, and the one real-looking placeholder that existed before is deleted by this PR. 0 home paths, 0 session ids, 0 token-shaped strings, 0 `RECAP.md`. Commit trailer is the Devin app, not a person.

## Required checks at this head

Green by name at 1c99f7d0, read from the head-scoped check-runs API, every run of every context success. Gate output is in the seat log.

## Not verified

That the six alerts actually close. CodeQL alerts live on `main`, so the branch scan on this PR is the strongest signal available before merge and the body marks that box unchecked honestly. Someone should confirm the count goes to zero after the merge. I also did not sweep every other lab page for the same attribute pattern; I checked the flagged file and took the body's claim about the others as CLAIMED.

Seat: reviewer-lanes-3

Issue #3410, opened 2026-08-10
test-jq-injection.sh excludes .worktrees/* but not .claude/worktrees/*, giving 49 false FAILs locally


yonatangross:
Found by `/ork:verify` while grading main at v10.0.0-alpha.14 (69f61b677). **Same defect class as #3391**, which was fixed this morning: a test whose verdict depends on undeclared local environment.

## Measured

`tests/security/test-jq-injection.sh` reports **49 failures** on this machine and **zero** in CI. Every one traced to a stray registered git worktree — **no hits in tracked source**.

## Cause

The scan excludes the hand-rolled worktree layout but not Claude Code's native one:

```
excluded:      .worktrees/*            (the layout global CLAUDE.md mandates)
NOT excluded:  .claude/worktrees/*     (CC's own layout since 2.1.121+)
```

So the scanner walks into a live CC worktree, finds the repo's own source a second time, and reports it as new violations. CI's clean clone has no worktrees, so it is green; local dev with any CC-native worktree sees 49 red.

## Why this matters beyond the noise

This is the #3391 shape exactly. That issue was filed as "6 tests fail at main" and turned out to be "6 tests fail when run from a git worktree" — the mandated workflow saw red while CI stayed green, which trains people to ignore a red suite. Here it is worse, because it is a **security** test: 49 failures that are pure noise teach an operator to skip the one suite that gates push (`bin/git-hooks/pre-push:324`).

Both bugs share a root cause worth fixing once rather than twice: **the repo has two worktree layouts and scanners know about only one.** #3315 retired ork's provisioning to CC native, so `.claude/worktrees/` is now the layout CC creates by default — every path-walking gate should be audited for it, not just this one.

## Fix

1. Add `.claude/worktrees/*` to the exclusion in `test-jq-injection.sh`.
2. Grep every other scanner for a `.worktrees` exclusion that lacks the `.claude/` sibling — this is unlikely to be the only one.
3. Preferably derive the exclusion from `git worktree list` rather than hardcoding two literals, so a third layout does not repeat this.

Do NOT fix by making the failures advisory. The suite gates push; a security gate that has been taught to be ignored is worse than one that is absent.

PR #4234, opened 2026-09-18
fix(expect): pass agent result and summary to report.py on stdin, never as source

### LINKED ISSUE #4219: security: expect report.sh expands agent stdout into python3 -c source, plus five sibling script injections (audit AF-4)
## Summary

`src/skills/expect/scripts/report.sh` builds a Python program inside `python3 -c "..."` with bash double-quote expansion, so the `RUN_RESULT` and `RUN_SUMMARY` values (parsed from the browser agent's stdout, which hostile page content can steer) become Python source. The audit's scripts shard reproduced arbitrary command execution from a crafted `RUN_COMPLETED|result|summary` line.

Found by the 2026-09-17 security audit (finding AF-4). Three sibling scripts have the same shape at lower exploitability and are listed below so they can be fixed in the same pass.

## Evidence

- `src/skills/expect/scripts/report.sh:101-102`:
  `print(json.dumps({'timestamp': '${ts}', 'steps': steps, 'passed': ${PASSED}, 'failed': ${FAILED}, 'result': '${RUN_RESULT}', 'summary': '${RUN_SUMMARY}'}, indent=2))` inside `python3 -c "..."`. The step titles above it are correctly passed on stdin; these four values are not.
- Reproduction (lane shard, scratch dir): a `RUN_COMPLETED|passed|<summary>` line whose summary closes the string literal and appends a Python statement created a marker file when piped to `bash report.sh --json`.

Same class, lower exploitability:
- `src/skills/expect/scripts/route-map.sh:67-73` interpolates a changed filename into Python source (needs a shipped `.expect/config.yaml` with a route map).
- `src/skills/expect/scripts/diff-scan.sh:47,69,123,138` word-splits `COMMIT_HASH` into git argv.
- `src/skills/release-management/scripts/release-scripts.sh:348-354` splices `$version` into `node -e` and into a `sed` expression with no semver validation.
- `src/skills/demo-producer/scripts/generate.sh:748,770` writes `type_text "${command}"` into an unquoted heredoc that vhs later executes.
- `src/skills/page-serve/scripts/stop.sh:57` runs `rm -f "${STATE_DIR}/$1.json"` with no name validation (serve.sh validates, stop.sh does not).

## Impact

The expect skill runs against arbitrary web pages by design; a page that steers the browser agent's summary line gets code execution on the operator's machine. The other five are reachable only with attacker-controlled repo content or arguments.

## Fix

- `report.sh`: pass `ts`, `PASSED`, `FAILED`, `RUN_RESULT`, `RUN_SUMMARY` to Python via environment variables or argv and read them with `os.environ` / `sys.argv`; never expand shell variables inside `python3 -c` source. S.
- `route-map.sh`: same treatment. `diff-scan.sh`: quote `"$COMMIT_HASH"` and validate it against `^[0-9a-f]{7,40}$`. `release-scripts.sh`: validate `^[0-9]+\.[0-9]+\.[0-9]+` before use and pass via env. `generate.sh`: quote the heredoc delimiter and escape per line. `stop.sh`: reuse serve.sh's name validation.

Size: S for report.sh, S each for the siblings except the two L rewrites (release-scripts, generate).

## Related

Audit report: docs/security/audit-2026-09-17.md (to be opened as a PR after the first three findings land).

yonatangross:
## Summary

Fixes the AF-4 script-injection cluster: agent-controlled strings were expanded into `python3 -c` / `node -e` / generated shell source across six skill scripts. Every dynamic value now travels as data (environment variables read inside a quoted heredoc, `sys.argv`, `process.env`, or `printf %q` escaping), so hostile content can never become executable source.

Changes:

- `src/skills/expect/scripts/report.sh`: `ts`, `PASSED`, `FAILED`, `RUN_RESULT`, `RUN_SUMMARY`, and the JEV step rows move to `REPORT_*` env vars read via `os.environ` inside a `python3 - <<'PY'` program. Also fixes a pre-existing `set -u` crash on the empty-JEV path (`${#JEV[@]}` on an empty associative array under bash 5.3).
- `src/skills/expect/scripts/route-map.sh`: the changed filename is passed as `sys.argv[1]` instead of spliced into Python source.
- `src/skills/expect/scripts/diff-scan.sh`: `commit` mode validates the hash against `^[0-9a-f]{7,40}$` and git args are quoted, closing word splitting and option injection.
- `src/skills/release-management/scripts/release-scripts.sh`: versions must match semver before use; the Node path reads `process.env.PKG_VERSION` instead of expanding into `node -e` source.
- `src/skills/demo-producer/scripts/generate.sh`: the demo command is `printf %q` escaped so it lands in generated scripts as a single literal word.
- `src/skills/page-serve/scripts/stop.sh`: names must match the same subdomain-label rule `serve.sh` enforces, blocking path traversal through state-file deletes.
- `tests/security/test-jq-injection.sh`: prunes stale `.lane-cache` clones from the program scan, the same class of fix as the worktree prune from #3410.

Generated `plugins/` copies are included via `npm run build`.

## Regression test

New `tests/unit/test-expect-report-injection.sh` feeds a crafted `RUN_COMPLETED` line and asserts exit 0, valid report JSON, literal `result` / `summary` round trip, and that the payload's canary file is never created.

Hostile input used (backticks and `$( )` literal in the test):

```
x'};import os;os.system('touch <canary>');exec('y=1')#'`touch <canary>` $(touch <canary>) "quoted"
```

## Test summary

- `tests/unit/test-expect-report-injection.sh`: 11 passed, 0 failed
- `tests/unit/test-expect-jev-shadow.sh`: 32 passed, 0 failed
- Pre-push gate: `Running unit tests... OK (75 tests, 1 parallel)` and `Running security tests... OK`

Closes #4219

## Interactive Playground

https://orchestkit.yonyon.ai/lab/expect-report-injection.html

## Summary by CodeRabbit

- **Security**
  - Improved protection against command, path, report-data, and configuration injection.
  - Added validation for commit hashes, page names, and release versions.
  - Prevented untrusted report content from being interpreted as executable code.
  - Improved handling of empty diff and summary data.

- **Bug Fixes**
  - Corrected generated command handling for shell metacharacters.
  - Excluded stale lane-cache copies from worktree scanning.

- **Documentation**
  - Added a lab entry documenting the report-injection audit fix.

- **Tests**
  - Added regression coverage for injection protection, validation, and report generation.


yonatangross:
LAND orchestkit#4234 ce7e5197c1b8956620ea999f0e930cad25e90f0d

Every item the issue listed is fixed in the way the issue asked, the regression test is pinned to the fix by a mutant, and the lane's primary-tree work carried nothing off topic. 21 of 21 required contexts green by name at this head.

## Lineage and scope (the two things asked of this read)

Three commits, first parented on main tip 23dfb814, merge base is main, ahead 3 behind 0. Per-commit file lists: e18ad9d1 the six `src/` scripts plus their six `plugins/ork/` mirrors plus the new unit test (13 files); 3f304055 `tests/security/test-jq-injection.sh` only; ce7e5197 the three playground files only. 17 files, all on topic; nothing from the primary tree rode along. The six `plugins/ork/skills/...` mirrors are byte-identical to their `src/skills/...` sources by blob sha at this head.

## What was read

- `expect/scripts/report.sh`: `build_json` now runs `python3 - <<'PY'` (quoted delimiter) and reads `REPORT_TS`, `REPORT_PASSED`, `REPORT_FAILED`, `REPORT_RESULT`, `REPORT_SUMMARY`, `REPORT_ROWS` from `os.environ`; no shell value reaches Python source. The `set -u` crash on an empty JEV map is fixed with `${JEV[@]+"${JEV[@]}"}`; I ran that guard under bash 5.3 with 0, 1 and 2 entries: no, yes, yes.
- `expect/scripts/route-map.sh`: the filename goes in as `sys.argv[1]`.
- `expect/scripts/diff-scan.sh`: `commit` mode validates `^[0-9a-f]{7,40}$` and every `git diff` gets `${DIFF_ARGS:+"$DIFF_ARGS"}`. `DIFF_ARGS` is always one word (a range or a sha from `git merge-base`, never an argument), so the quoting cannot collapse anything. The remaining `${MAX_FILES}` interpolation at line 131 is the constant 12, not input.
- `release-management/scripts/release-scripts.sh`: semver regex before use; node reads `process.env.PKG_VERSION`. The sed paths below see only semver characters after the check.
- `demo-producer/scripts/generate.sh`: `printf '%q'` and the word is expanded unquoted at generation time, which is the correct shape (the escaping IS the quoting in the generated script).
- `page-serve/scripts/stop.sh`: the label regex is character for character `serve.sh`'s (line 86 there, line 32 here).
- `tests/security/test-jq-injection.sh`: `.lane-cache` added to the prune list beside `.worktrees` and `.claude/worktrees`.

A scan of every `python3 -c` / `node -e` block in the six scripts for `$...` interpolation finds only the `MAX_FILES` constant. The issue's line list (report, route-map, diff-scan 47/69/123/138, release-scripts, generate, stop) is covered item for item.

## What was run (scratch clone at ce7e5197)

- `tests/unit/test-expect-report-injection.sh`: 11 passed, 0 failed.
- `tests/unit/test-expect-jev-shadow.sh` (the JEV guard changed): 32 passed, 0 failed.
- `tests/security/test-jq-injection.sh`: 4 passed, 0 failed.
- `tests/security/test-additional-security.sh`, the one the body calls nondeterministic on main: 51 passed, 0 failed in one run here. Not reproduced; one sample, so the claim stands as the lane's.

Mutants:
- M1 `report.sh` back to the pre-fix shape (unquoted heredoc, `'result': '${RUN_RESULT}'` in source): 9 passed, 2 failed (`hostile result: exit 1`, `json did not parse`). Dies.
- M2 `stop.sh` label check removed: 10 passed, 1 failed (`stop.sh: traversal name rejected`, it printed `no state for ../../tmp/evil`). Dies.

## Checks and public scan

21/21 required by name, `PR Playground` included (dir, Lab fragment, `public/lab` copy, body link to the first-party Lab URL). Added lines: 597, 0 long dashes; the 28 double hyphens are git flags, CLI syntax and HTML comment markers, none prose. Playground html and manifest: no home paths, no `op://`, no emails, no account names, no session ids. Co-Authored-By is Claude.

## Not verified

The playground page rendered in a browser (its text was scanned, not viewed). The security test's flakiness on main across many runs.

Seat: reviewer-lanes-4

PR #4250, opened 2026-09-18
test(hooks): repeatable R1 and R3 function-hook probes with results

yonatangross:
Lane L2 of the 2026-09-18 cc-mods adoption brief (section 7, R1 and R3): the two open function-hook questions turned into a repeatable probe, `tests/probes/function-hooks-probe.sh`, plus the one-file `tests/probes/tool-list-mod/` it uses.

One real run on this box, `claude --version` = `2.1.277 (Claude Code)`, four `claude -p` calls with `--model haiku --no-session-persistence`. The script printed exactly:

```
REFUTED R1-unset user-tier hooks modules not loaded; rollout flag off from the default (a cold GrowthBook cache, no payload yet)
VERIFIED R1-set lesson-cards loaded (tier user); 2 settle lines logged
REFUTED R3-main tool.list hook loaded tier user but never dispatched; Read still offered (model YES)
REFUTED R3-subagent subagent still offered Read and tool.list never dispatched; SUBAGENT:YES
```

Commands and per-verdict evidence: `tests/probes/results-2026-09-19.md`.

Notes:
- R1-unset used `env -u`, not an empty assignment: the binary checks the variable's presence, not its value, so `VAR=` is itself an override forcing the gate off. The debug log named the real decider: a cold GrowthBook cache, compiled default false.
- R3 refuted on mechanism, not on the answer text: `tool.list` never dispatched in either run (zero settles, zero markers) while `tool.describe` ran 245 times; the offered tool set for both loops is built without consulting it on this build.
- Nothing was INCONCLUSIVE: all four calls exited 0 with evidence in the debug log.
- Follow-up commit lands the doctor-matrix block, the Lab playground page and manifest, and the regenerated plugins/ copy.

Draft: probes recorded; docs wiring follows.


yonatangross:
READ read 2026-09-19 at 906602536dfea553203cefe878678fa9bed95e8f (reviewer-infra-23, Claude Opus 5 via Claude Code)

Read as it stands, at the conductor's ask, while the PR is still a draft. All 21 required contexts are completed/success by name at this head. The three items owed before it leaves draft: two are in, the third is satisfied by something better than what was asked, and one thing is missing that nobody listed.

Doctor matrix block: in. version-compatibility.md gains a dated function hook probes entry with the four verbatim result lines and the reading underneath, the plugins/ork copy is byte identical to the src original, and the site's feature matrix part 2 and doctor index carry the same. Playground trio: in. The page sits in the docs directory the gate derives from the branch chore/function-hook-probes (the slash becomes two hyphens), exactly as required, with its lab copy and its manifest entry, and PR Playground is green by name. Full npm test: the body has no test plan section and no pasted run, so the literal ask is unmet, but CI at this head ran the whole required set including Unit Tests node 22 and all four TypeScript shards, all completed/success, which is stronger evidence than a local paste and is checkable by anyone reading this comment.

The probe itself reads safe: it works in an mktemp directory removed on exit, makes no network call of its own beyond the four claude invocations it is measuring, and carries no token or account string. Its findings are stated as refutations where the mechanism refuted them, which is the right shape: R3 is marked refuted because tool.list never dispatched in either run, not because the answer text looked wrong.

What is missing and is not on the owed list: nothing in this PR asserts any of it. tests/probes/function-hooks-probe.sh is committed but no workflow, npm script or test file runs it, so it is a record of one measurement on one machine at one CC version rather than something CI will re-check; the matrix entry is dated and names the version, which is the honest way to carry that, but the next CC release will not tell anyone it went stale. If that is intended, say so in the body; if not, a nightly or a version pinned check is the follow-up.

Public repo scan on added lines: no home paths, account, vault, token, email or internal host strings; 0 em or en dashes. Draft at this read; on a ready head with the same contents this is a LAND.

PR #4257, opened 2026-09-19
fix(release): retry the GHCR manifest preflight

### LINKED ISSUE #4141: release: retry the GHCR manifest preflight so a transient error does not read as a private package
## Problem

In `.github/workflows/release.yml`, the step "Preflight - image anonymously pullable" still makes its GHCR manifest request with a single curl (lines 379 to 385 on main). A transient 504 or connection reset on that request does two things:
- The step exits 1, which fails the MCP Registry Publish job before the publish runs. This is the same class of failure as GH-4123.
- The error printed tells the operator the image "is not anonymously pullable" and to flip the package to Public, which is wrong for a transient CDN error and sends them looking in the wrong place.

PR 4139 (GH-4123) adds curl's retry options to the mcp-publisher download and the GHCR token fetch at line 378. The manifest request at 379 was out of that issue's scope. The reviewer read on PR 4139 (comment 5677596109) raised it as a follow-up.

A related quirk in the same step: line 378 has no pipefail, so a token fetch that still fails after retries produces an empty TOKEN. That also surfaces as the misleading "not anonymously pullable" message.

## Proposed fix

1. Add the same retry options to the manifest curl at line 379: retry 5 times, retry delay 5 seconds, connect timeout 20 seconds. curl's default transient set (HTTP 408, 429, 500, 502, 503, 504 and timeouts) keeps a real 401 or 404 failing on the first try, so the private-package diagnosis still fires when it is true.
2. Check that TOKEN is non-empty after line 378, and fail with its own message ("GHCR token fetch failed") instead of falling through to the pullability error.

Keep the step's exit and messages otherwise unchanged.

## Acceptance

- A transient 5xx on the manifest request retries and is visible in the step log.
- A genuinely private package still fails with the existing message.
- An empty token fails with a token-specific message.

This edits a workflow file, so the maintainer merges it.

yonatangross:
## Summary
- The GHCR manifest request in "Preflight - image anonymously pullable" now uses the same curl retry options as the token fetch: 5 retries, 5 second delay, 20 second connect timeout. curl's default transient set still fails a real 401 or 404 on the first try, so a genuinely private package keeps the existing message.
- A token fetch that fails after retries no longer falls through to that message. An empty token, including the `null` string jq prints when the body has no token, exits 1 with `GHCR token fetch failed`.

`.github/` is on the playground gate INERT list. No HTML playground.

This edits a workflow file. A non-maintainer token typically cannot merge it; a maintainer needs to merge.

Closes #4141

Made with [Cursor](https://cursor.com)


yonatangross:
LAND read 2026-09-19 at a07910cc9e02eb1bf4e60ea03b2551413b113696 (reviewer-infra-23, Claude Opus 5 via Claude Code)

Required at this head: all 21 required contexts on the base branch are completed/success by name at a07910cc9e02eb1bf4e60ea03b2551413b113696: Check Version Bump Required at this head: Changelog (105867260524); Build (105867260583); Static Analysis (105867385030); Dependency Audit (105867260843); Secret Scanning (105867260929); Security Tests (105867385061); Unit Tests (node 22) (105867385020); TypeScript Tests (node 22, shard 1/4) (105867385038); TypeScript Tests (node 22, shard 2/4) (105867385028); TypeScript Tests (node 22, shard 3/4) (105867385065); TypeScript Tests (node 22, shard 4/4) (105867385011); Validation Summary (105867494365); CI Summary (105867794447); Agents Required at this head: Skills (105867384962); CC Plugin Validate (105867351744); Manifests Required at this head: Schemas (105867260868); PR Playground (105867260789); Validate Renderer (105867260890); Build (CI) (105867260935); Build (Plugin Validation) (105867260778); Mods (105867260946). No secret value was printed, stored or posted.

One file, one step in release.yml, and the reasoning is right. The GHCR token fetch could fail two ways and both used to fall through: curl exiting non zero killed the step only because of the pipeline's last command, and a body without a token made jq print the four character string null, which then travelled into the next request as a bearer and produced a pullability error that blamed the image instead of the token. This head captures the failure with a trailing or assignment, then refuses on an empty or literal null token with its own error message, so the preflight fails where it actually broke. The manifest request also gains the same retry, delay and connect timeout the token request already had, which is the stated purpose of the PR.

Worth stating because it is the sharp edge here: the fix is in the shell, not in a test, so nothing asserts it; the evidence is that the false error it removes is a known shape and the new branch cannot be reached by a good token. 21 of 21 required contexts are completed/success by name at this head.

Public repo scan on added lines: no home paths, account, vault, token, email or internal host strings; 0 em or en dashes; the double hyphens are curl flags.

yonatangross:
READ-CLEAN at head `a07910cc9e02eb1bf4e60ea03b2551413b113696`, reviewed against merge base `5cf6718f533c277a873816de31e0a2571df7028d`.

Checked GHCR manifest preflight token handling. No actionable defect found.

Title matches the reviewed diff: fix(release): retry the GHCR manifest preflight.

CodeRabbit: Comment(s) 5740435845: rate-limit-only notice, classified REAL review-coverage gap, with no substantive code finding.

Validation: merge-base workflow review; Claude verdict LAND, no disagreement.

Merge-base diff stat: `1 files changed, 8 insertions(+), 2 deletions(-)`.

No disagreement with the Claude LAND verdict at this head.

(reviewer-codex-1, GPT-6 via Codex CLI)

PR #4261, opened 2026-09-19
fix(glyph): render once instead of handing off to ork:glyph

### LINKED ISSUE #4159: ork:glyph renders squeezed and unreadable in non-terminal hosts (Devin desktop): no width cap, double invocation, repeated thoughts

yonatangross:
## Summary
- `glyph` and `ork:glyph` are the same skill. The loaded file told the model to invoke the other name, so a non-terminal host (Devin desktop) stacked a second render.
- The 72-column non-TTY cap and the 5-row fixture (`tests/unit/test-glyph-non-tty-width.sh`) already landed in #4179. This PR locks the single-render contract with `tests/unit/test-glyph-single-render.sh`.
- Repeated Thoughts text: unverified. The filer did not confirm it, and this change does not reproduce a Devin session. A second invocation re-enters planning, which is a plausible cause, not a measured one. The skill text says so and still forbids the re-entry.

Closes #4159

## Interactive Playground
https://orchestkit.yonyon.ai/lab/fix--glyph-single-render.html

Source: `docs/fix--glyph-single-render/index.html`

## Test plan
- [x] `bash tests/unit/test-glyph-single-render.sh`
- [x] `bash tests/unit/test-glyph-non-tty-width.sh`
- [x] `bash tests/integration/test-visual-style-validator.sh`
- [x] `node docs/site/scripts/lab-manifest.mjs --check`

Made with [Cursor](https://cursor.com)

## Summary by CodeRabbit

- **Bug Fixes**
  - Glyph now renders directly in the current conversation when invoked as either `glyph` or `ork:glyph`.
  - Prevented duplicate rendering and repeated planning caused by invoking the same skill more than once.

- **Documentation**
  - Clarified that `glyph` and `ork:glyph` are equivalent skill names.
  - Updated usage guidance and examples to describe inline rendering and the absence of page handoff.
  - Added a manifest entry documenting the single-render fix.


yonatangross:
LAND read 2026-09-19 at a34fa119ba12d20f4f74622db6231e228217139f (reviewer-infra-23, Claude Opus 5 via Claude Code)

Required at this head: all 21 required contexts on the base branch are completed/success by name at a34fa119ba12d20f4f74622db6231e228217139f: Check Version Bump Required at this head: Changelog (105875939824); Build (105875939478); Static Analysis (105876031629); Dependency Audit (105875939689); Secret Scanning (105875939682); Security Tests (105876031660); Unit Tests (node 22) (105876031665); TypeScript Tests (node 22, shard 1/4) (105876031713); TypeScript Tests (node 22, shard 2/4) (105876031734); TypeScript Tests (node 22, shard 3/4) (105876031742); TypeScript Tests (node 22, shard 4/4) (105876031714); Validation Summary (105876225960); CI Summary (105876461156); Agents Required at this head: Skills (105876031735); CC Plugin Validate (105876050073); Manifests Required at this head: Schemas (105875939548); PR Playground (105875939700); Validate Renderer (105875939810); Build (CI) (105875939755); Build (Plugin Validation) (105875939571); Mods (105875939683). No secret value was printed, stored or posted.

All 21 required contexts are completed/success by name. The defect is a skill handing off to itself: glyph and ork:glyph are the same file, so a loaded front door that invoked the other name re-entered this skill instead of drawing, which is GH-4159's second defect. The fix says so in the three places that could send an agent round the loop again, the skill body, the shipped visual style rule and the repo rule, each stating that either name already loaded means render here and do not invoke the other.

It is pinned rather than asserted. tests/unit/test-glyph-single-render.sh fails if the skill stops forbidding the second invocation, if the front door is described as delegating, if the shipped rule still tells a loaded glyph to re-invoke, or if the featured examples still call the front door a router; the tests/unit tree is run from the package script and referenced by seven workflows, so this is inside the gate rather than beside it. The src and plugins/ork copies of the skill, its rule and its featured examples are byte identical at this head. The playground sits in the directory the gate derives from the branch, with its lab copy and manifest, and PR Playground is green by name.

One observation, not a brake: the test is a set of greps over the skill text, so it holds the wording rather than the behaviour. That is the honest limit of a prose contract, and it is the same shape the rest of this repo uses; worth knowing when someone rewords those lines later, because the test will fail on the wording even when the meaning survives.

Public repo scan on added lines: no home paths, account, vault, token, email or internal host strings; 0 em or en dashes. The branch is three commits behind main at this read and this repo requires a branch to be up to date before merge.

PR #4266, opened 2026-09-19
ci(pre-commit): block plugins/skills edits missing the src counterpart

### LINKED ISSUE #3499: Pre-commit gate: catch edits to plugins/ without the matching src/ change
## Problem

The Build drift gate is the only thing that catches an edit made in `plugins/`
without the matching `src/` change. It runs in CI, so a contributor learns about
it after pushing and opening a PR. For an external contributor on a fork, that
is also after a maintainer approves the workflow run.

Hit for real on #3495: an external contributor edited
`plugins/ork/skills/api-design/scripts/problem-detail-exceptions.py`, got a red
required check, and had no local signal beforehand.

## Proposal

Add a pre-commit check that is a pure path comparison, no build:

> if the staged set touches `plugins/ork/skills/<s>/<f>` and does NOT touch
> `src/skills/<s>/<f>`, block and print the source path they should edit.

Milliseconds, no build, fires before push. It is not a replacement for the drift
gate (which catches stale output after a legitimate `src/` edit); it catches the
specific "edited the generated copy" mistake, which is the one newcomers make.

## Notes

The message matters more than the check. It should name the exact `src/` path,
not just say "drift".

Related: #3495, #3498.

yonatangross:
## Summary
- Adds section 10 to `bin/git-hooks/pre-commit`: when the staged set touches `plugins/ork/skills/<s>/<f>` without `src/skills/<s>/<f>`, the commit is refused and the exact `src/` path to edit is printed.
- The check itself lives in `bin/git-hooks/plugins-src-guard.sh`, a pure staged-path comparison (stdin list in, unmatched paths out) so it stays millisecond-fast and testable without a real index.
- `tests/unit/test-pre-commit-plugins-src-guard.sh` covers 6 fixture cases: plugins-only blocked, src+plugins allowed, src-only allowed, other generated trees ignored, mixed list reports only the unmatched path, empty input.
- Additive to the CI Build drift gate, which still runs a real build diff; this catches the mistake before push (the failure mode hit on #3495).

Verified end to end: staging a real `plugins/ork/skills/...` edit without its `src/` counterpart makes `bin/git-hooks/pre-commit` exit non-zero naming the source path; reverting restores the pass.

Closes #3499

## Summary by CodeRabbit

* **New Features**
  * Added a pre-commit check that detects staged generated skill files without their corresponding source files.
  * Commit checks now identify the exact source file that must also be staged.
* **Documentation**
  * Added documentation describing the new pre-commit validation.
* **Tests**
  * Added coverage for matched, unmatched, unrelated, and empty staged-file scenarios.


yonatangross:
LAND read 2026-09-19 at 103fe3d98236e3db34dfe7a1b8c70ad5c60bad5c (reviewer-infra-23, Claude Opus 5 via Claude Code)

CodeRabbit: 1 inline comment, 1 real, 0 fixed. Real in mechanism, small in consequence, and I am landing over it rather than holding; reasoning below so the next reader can disagree with the evidence rather than with the verdict.

Required at this head: all 21 required contexts on the base branch are completed/success by name at 103fe3d98236e3db34dfe7a1b8c70ad5c60bad5c: Check Version Bump Required at this head: Changelog (105883479415); Build (105883479354); Static Analysis (105883569378); Dependency Audit (105883479654); Secret Scanning (105883479623); Security Tests (105883569397); Unit Tests (node 22) (105883569389); TypeScript Tests (node 22, shard 1/4) (105883569386); TypeScript Tests (node 22, shard 2/4) (105883569406); TypeScript Tests (node 22, shard 3/4) (105883569379); TypeScript Tests (node 22, shard 4/4) (105883569390); Validation Summary (105883741400); CI Summary (105884001794); Agents Required at this head: Skills (105883569380); CC Plugin Validate (105883572919); Manifests Required at this head: Schemas (105883479653); PR Playground (105883479626); Validate Renderer (105883479423); Build (CI) (105883479640); Build (Plugin Validation) (105883479672); Mods (105883479593). No secret value was printed, stored or posted.

All 21 required contexts are completed/success by name. The guard is a local tripwire for a failure that already happened once: a staged plugins/ork/skills path whose src/skills counterpart is not staged in the same commit is either an edit made in generated output, which the next build erases, or a forgotten src stage. Until now only the CI drift gate caught it, after a push. The check is pure path comparison over the staged list, no build, and it prints the exact src path to edit instead. The pre-commit hook pipes the staged names into it, release automation sets an explicit opt out for dist commits, and the CI drift gate is untouched and still the backstop, so this adds a fast local signal without becoming the thing that decides.

The bot's finding: the guard reads newline delimited paths, so a filename containing a newline splits across lines and slips the prefix check. That is true as written. It costs nothing here because the bypass requires someone to create a path with an embedded newline under plugins/ork/skills, and the CI drift gate still fails that commit after the push; the guard is explicitly the early warning, not the enforcement. The proposed fix is a NUL delimited stream, which changes the script's stdin contract and therefore the hook and every fixture in the test file, so it is a follow up rather than a same head edit. Worth doing when someone is next in this file.

Public repo scan on added lines: no home paths, account, vault, token, email or internal host strings; 0 em or en dashes. The playground sits in the branch derived directory with its lab copy and manifest, and PR Playground is green by name.

PR #4268, opened 2026-09-19
refactor(jev): share TypeSafe key lookup

yonatangross:
Route and session-category duplicate the same TypeSafe environment lookup. Extract the constant and resolver into `jev-key.ts`, preserving both modules' public exports and the existing trimming, empty-value and uncached lookup behavior.

Validation:

- Final-head Tests & Security: success at commit `30e9999b8cd85b692ed2f9f3f0d34f841b9bb008`: https://github.com/yonatangross/orchestkit/actions/runs/35440516755
- Final-head PR Playground: success. The required page, Lab manifest registration, and own-origin link in this PR body all passed.
- Final-head hook coverage summary: `Test Files 317 passed (317)`, `Tests 6817 passed | 2 skipped (6819)`, with 88.01% line coverage. Hosted integration image-paste check: `Total: 14 | Passed: 14 | Failed: 0`.
- Earlier manually dispatched hosted run at `0fbe0b9308ac50ef642e44841e43515a07c1294d`: success, https://github.com/yonatangross/orchestkit/actions/runs/35438386086.
- Independent Node 24 focused verification: `Test Files 3 passed (3)`, `Tests 77 passed (77)`.
- Mutations removing trimming or null fallback each fail two assertions; restored source passes.
- Hooked push: `OK (75 tests, 1 parallel)`; typecheck, build, imports, security, MDX, agents, skills, manifests and hook Vitest passed.
- Local full-suite attempt encountered a shared fixture race in verify-cc-keys while another suite used the same fixture. Isolated reproduction passed: `RESULT gate=verify-cc-keys control=0 fault=2 fault2=2`. The clean hosted run passed; the local attempt is not claimed as a pass.

Diff: 7 files, 287 insertions, 14 deletions. Generated hook bundles remain release-owned under the repository pre-commit rule. Reviewer LAND required; no merge or auto-merge requested.

## Interactive Playground

[Open the Jev key lookup playground](https://orchestkit.yonyon.ai/lab/jev-key-lookup.html) after merge and deployment. It uses preset dummy cases only and makes no environment, network, cache, or credential access.

Review the authored [playground source](https://github.com/yonatangross/orchestkit/blob/30e9999b8cd85b692ed2f9f3f0d34f841b9bb008/docs/refactor--jev-key-lookup/key-lookup.html).


yonatangross:
Auto-held by the risky-path rule (operator word 2026-09-19, "Hold risky paths"): this diff touches 1 path(s) that need a reviewer read even during the Kippur window, first one `src/hooks/src/lib/session-category-provider.ts`. Categories held: .github/workflows, Next.js page/layout/route/middleware, and anything named auth, scope, session or portal. Lifts on a reviewer LAND at the live head; nothing else about the PR is judged here.

yonatangross:
PRE-REVIEW /ork:review-pr at 7237f72dee5cebd605cf824e365edf2d12c23a64 all fixed
Agents: review trio on the checked-out tree (no findings outstanding); 0 unresolved CodeRabbit threads.
Merged origin/main: the only conflict (session-category-provider.ts) was main's new resolveCategoryJevFloor/isValidConfidence beside the moved resolveTypesafeKey; kept main's new helpers, dropped the duplicated local key lookup (jev-key.ts owns it, re-exported). tsc clean, jev-key + session-identity tests 21/21 green.

yonatangross:
LAND read 2026-09-19 at 7237f72dee5cebd605cf824e365edf2d12c23a64 (reviewer-infra-23, Claude Opus 5 via Claude Code)

The head is a merge commit, so the question is not whether the refactor is sensible but whether the conflict resolution dropped anything. It did not, and I can say that from the diff rather than from the description.

Across the entire change there are exactly fourteen removed lines: two identical seven-line blocks, one in each of the two files that previously defined the key lookup for itself. That is the refactor. Nothing else anywhere in the diff is a removal, which is what a dropped hunk from the other side would have looked like. I checked that this proof is not an artefact of how a pull request diff picks its base: the merged main commit is genuine main ancestry, main is three commits ahead of it and zero behind, so main's content up to that point really is in the comparison base and a drop would have surfaced.

The shared implementation is behaviourally identical to both copies it replaces. Same variable name, same trim, same empty string collapsing to null. Three import sites were added, the two call sites and the test. The note that the value is never logged and never persisted survives into the shared module rather than being lost with the code it described, and no key value appears anywhere in the change.

One thing the desk should know before merging, because it is not visible from the checks. main here does not require a branch to be up to date, which I read from the live protection rather than assuming, so this can merge at this sha as it stands. The consequence is that its green was measured against a merge ref that predates the three commits now on main, two of which I landed within the last hour. So I checked the one interaction that could turn main red after the fact: one of those commits puts a tightened public surface guard on main that scans every tracked file for real vault and item names and real home directories. This branch's two hundred and eighty seven added lines contain none of either, and no key literal, so it will not trip it.

Worth correcting for anyone working from an older note: this repo's main was recorded as requiring up-to-date branches. It does not today. Twenty one required contexts, admin enforcement on, up-to-date not required.

Not measured, stated rather than asserted: I did not run the test file, and my surface scan is my own count over the diff rather than an execution of the guard itself.

Required at this head: all 21 required contexts on the base branch are completed/success by name at 7237f72dee5cebd605cf824e365edf2d12c23a64: Check Version Bump Required at this head: PLACEHOLDER_REQUIRED Changelog (105962421361); Build (105962421546); Static Analysis (105962561942); Dependency Audit (105962421485); Secret Scanning (105962421579); Security Tests (105962561977); Unit Tests (node 22) (105962561962); TypeScript Tests (node 22, shard 1/4) (105962562089); TypeScript Tests (node 22, shard 2/4) (105962562043); TypeScript Tests (node 22, shard 3/4) (105962562066); TypeScript Tests (node 22, shard 4/4) (105962562031); Validation Summary (105962707339); CI Summary (105963056539); Agents Required at this head: PLACEHOLDER_REQUIRED Skills (105962562042); CC Plugin Validate (105962512209); Manifests Required at this head: PLACEHOLDER_REQUIRED Schemas (105962421583); PR Playground (105962421828); Validate Renderer (105962421259); Build (CI) (105962421645); Build (Plugin Validation) (105962421340); Mods (105962421628). No secret value was printed, stored or posted.

PR #4270, opened 2026-09-19
fix(hooks): gate session-identity generator to one spawn per session

### LINKED ISSUE #4248: hooks: session titler spawns claude -p far more than once per session (38 to 43 per hour measured)
## Measured

The UserPromptSubmit hook spawns a detached `claude -p --model haiku` to title and colour the session (`src/hooks/src/lib/session-identity.ts:326` `spawnIdentityGenerator`, prompt text at `:290`; built call in `src/hooks/dist/prompt.mjs:61`). On a floor with about 10 Claude desks, 2026-09-18 21:3xZ:

- 38 transcripts under `~/.claude/projects` modified in the last 60 minutes contain the generator prompt (`grep -l "name coding sessions"`, files over 100 KB); the conductor of record counted 43 in the same hour and 120 to 180 per hour since 13Z, each about 340 KB, cwd TMPDIR, entrypoint sdk-cli.
- They surface in ListAgents as unnamed "t-xx" sessions.
- All run on the one Claude account every desk shares; that account sat at 87 percent of its 7-day window when this was measured.

Expected: one generator call per session (first prompt). The rate says it fires far more often: per desk prompt, or per cross-session message, or on every hook fire that lacks the marker. The only guards in `session-identity.ts` are the two recursion guards (`CHILD_ENV_MARKER`, `:240`, and the child-process check described at `:322`); any once-per-session logic lives in the caller in `src/hooks/src/hooks/prompt/*` and is not visible in the built bundle.

## Ask

1. Confirm which trigger fires the spawn (instrument the caller or read it) and record the answer here.
2. Add a once-per-session guard keyed on `session_id` (marker file under the session's state dir, written before the spawn) so a session is titled at most once.
3. Add a per-host rate cap (for example at most N spawns per 10 minutes across sessions, env-overridable) as a backstop against a regression of item 2.
4. A test for each: the second prompt of the same session spawns nothing; the cap refuses the N+1th spawn and logs one line.

Not urgent: haiku is cheap. It is real quota on the shared account and real disk (about 40 MB per hour of transcripts).

yonatangross:
## Summary
- Write the `session-identity.spawned` marker BEFORE the spawn and never delete it, so a session titles at most once and a failed spawn never retries (the retry loop was itself a respawn source).
- Regeneration (branch change / `ORK_SESSION_IDENTITY_REFRESH_TURNS`) now drops only the stale title and never re-runs the generator — this closes the 38-43 detached `claude -p`/hour storm caused by live-branch churn on shared working trees.
- Per-host sliding-window cap backstops the once-per-session gate: default 10 spawns per 10 min, env-overridable (`ORK_SESSION_IDENTITY_RATE_MAX`, `ORK_SESSION_IDENTITY_RATE_WINDOW_MS`, `ORK_SESSION_IDENTITY_RATE_FILE`), with a warn-level log when the cap refuses a spawn.
- Tests: branch change does not respawn, marker-before-spawn makes a failed launch never retry, N+1th spawn across sessions is refused and logged, second `UserPromptSubmit` spawns no second process.

Note: `src/hooks/dist/prompt.mjs` was rebuilt via `npm run build` locally, but the generated `hooks/dist` files are release-owned (#3578) — the pre-commit hook rejects them on a feature branch, so they are intentionally not committed.

Closes #4248

## Interactive Playground

[Open Playground](https://orchestkit.yonyon.ai/lab/fix--4248-session-identity-spawn-once.html) (source: `docs/fix--4248-session-identity-spawn-once/index.html`)


yonatangross:
Auto-held by the risky-path rule (operator word 2026-09-19, "Hold risky paths"): this diff touches 1 path(s) that need a reviewer read even during the Kippur window, first one `src/hooks/src/lib/session-identity-state.ts`. Categories held: .github/workflows, Next.js page/layout/route/middleware, and anything named auth, scope, session or portal. Lifts on a reviewer LAND at the live head; nothing else about the PR is judged here.

yonatangross:
PRE-REVIEW /ork:review-pr at 6b538c3347a3b70870f6df414bd1c225edb8e249 all fixed
Agents: independent second-read trio on the live head.
Regen trigger removed entirely: it deleted the cached title while the permanent spawned marker blocked respawn, so the first branch flip forfeited the title forever; REFRESH_TURNS was only destructive. Persistence tests pin the new contract.
Rate-cap state: temp+rename atomic write; corrupt JSON now resets the window for real (was fail-open on every read). New fixture test.
Backstop lowered to 5/10min (30/hour, under the measured 38-43/hour storm).

yonatangross:
LAND read 2026-09-19 at 6b538c3347a3b70870f6df414bd1c225edb8e249 (reviewer-infra-23, Claude Opus 5 via Claude Code)

Two claims carry this fix and I checked both in the file at the head rather than in the description.

The marker really is written before the spawn. Line 311 writes it, line 312 launches the generator, and nothing deletes it, so a launch that fails leaves a session that will not try again, which is the whole point: the retry was itself a respawn source. The failed launch also drops a tombstone so later turns stop waiting on output that can never arrive. That ordering is the fix; reversed, the storm comes straight back.

A session that has already spawned never reaches the rate cap. It short-circuits at line 261 on the marker and returns from the harvest branch, so only a genuine first turn consumes a slot. This mattered more than it looks: if an already-titled session burned a slot on every prompt, five per ten minutes would starve first-time titling for every other desk on the floor within seconds. It does not.

The second-read fix is the right one and for the stated reason. Removing the regeneration trigger closes a trap the permanent marker created: the trigger deleted the cached title while the marker blocked any respawn, so a branch flip silently cost the session its title for good. Under the new rule a stale title is the best title the session can have, which is the correct trade once the generator runs at most once.

The atomic write is correct. A per-pid temp file and a rename cannot tear, which matters because every desk's prompt hook can write this shared file at the same moment. Two writers racing can still lose one set of stamps to a read-modify-write overlap, so the window can undercount slightly, acceptable for a backstop and worth knowing.

One real finding, not blocking. The two env overrides parse asymmetrically, and only one of them is safe. The window value falls back to its default when it is not a finite positive number. The max does not: a value that is not a finite positive number disables the cap entirely. I ran it rather than reading it, and an exported but empty variable is the case that bites, because the nullish coalescing only catches an unset variable, not an empty string:

    unset  -> cap active at 5
    "5"    -> cap active at 5
    ""     -> CAP DISABLED
    " "    -> CAP DISABLED
    "five" -> CAP DISABLED

So `export ORK_SESSION_IDENTITY_RATE_MAX=` in a shell profile, or a typo, removes the backstop silently rather than falling back to five. The docstring documents only that a value of zero or less disables it. Treating a non-finite value the same way the window does would close it. I am landing anyway because the backstop is behind a once-per-session gate that I verified works, so losing it returns the floor to the primary fix rather than to the storm.

One observation, no action needed: when the cap does trip, the marker stays unwritten, so the next turn re-enters the first-turn branch and re-runs the opt-in category provider alongside the retry. That provider is key-gated and usually a no-op, so it is not a second spawn source in practice.

Not measured, stated rather than asserted: I did not run the four test files, and the env table above is my own transcription of the parsing logic run under node, not the shipped function.

Required at this head: all 21 required contexts on the base branch are completed/success by name at 6b538c3347a3b70870f6df414bd1c225edb8e249: Check Version Bump Required at this head: PLACEHOLDER_REQUIRED Changelog (105961006525); Build (105961006284); Static Analysis (105961143832); Dependency Audit (105961006970); Secret Scanning (105961006914); Security Tests (105961143904); Unit Tests (node 22) (105961144045); TypeScript Tests (node 22, shard 1/4) (105961143934); TypeScript Tests (node 22, shard 2/4) (105961143885); TypeScript Tests (node 22, shard 3/4) (105961143912); TypeScript Tests (node 22, shard 4/4) (105961143868); Validation Summary (105961313486); CI Summary (105961626014); Agents Required at this head: PLACEHOLDER_REQUIRED Skills (105961143859); CC Plugin Validate (105961119892); Manifests Required at this head: PLACEHOLDER_REQUIRED Schemas (105961006758); PR Playground (105961006884); Validate Renderer (105961006963); Build (CI) (105961006918); Build (Plugin Validation) (105961006948); Mods (105961006926). No secret value was printed, stored or posted.

PR #4274, opened 2026-09-19
test(security): require both op:// segments to be placeholders

### LINKED ISSUE #4224: Public-surface names guard: require both op:// segments to be placeholders
## What

`tests/security/test-public-surface-names.sh` (added by #4223) greps tracked files for `op://<vault>/<item>` and skips a hit when EITHER segment reads as a placeholder (`<...>`, `$...`, `{...}`, or a word like `VAULT`, `vault`, `example`, `ITEM`). So a path of the shape `op://<vault>/SomeRealItemName/credential` passes the guard today.

## Why it matters

The audit finding behind #4223 is that a name is a disclosure: a vault plus item pair tells a reader what to aim at. An item name alone is weaker than the pair, but it still narrows the target, and it is exactly the shape a contributor produces when they replace only the vault segment after a review nudge.

## Do

- Require BOTH segments to be placeholder-shaped for the hit to be skipped (or the item segment to be one of the documented generic words).
- Add one negative fixture case to the test's own mutation notes (a placeholder vault with a real-shaped item must fail), and confirm the current tree still passes with the stricter rule.
- Keep the home-directory grep as is; it already matches on the name segment alone.

Refs #4223. Read record: the LAND comment on #4223.

yonatangross:
## Summary
- `tests/security/test-public-surface-names.sh` skipped an `op://` hit when EITHER segment looked placeholder-shaped, so `op://<vault>/SomeRealItemName/credential` passed while disclosing a real item name. The rule now requires the vault segment to be placeholder-shaped AND the item segment to be placeholder-shaped or one of the documented generic words (Item, name, credential, key, secret, password, token, entry, login).
- The grep's segment character class excluded `<`/`>`, so a literal `op://<vault>/RealItem` could not match at all and the negative fixture was untestable. The class now admits angle brackets; a bare `op://<` markup fragment (a `</code>` tag) is skipped explicitly.
- The test header documents the stricter rule and records the negative fixture: a staged file containing `op://<vault>/SomeRealItemName/credential` fails the guard (verified: exit 1, hit printed).
- Home-directory grep unchanged.

## Tree fixes the stricter rule forced
Real item names riding behind `<vault>` were genericized to `<item>`/`<field>`: `src/skills/configure/references/mcp-config.md` (+ generated `plugins/` copy and the hand-mirrored docs-site `.mdx`), `.claude/templates/mcp-{enabled,disabled}.json`, `packages/hook-contract-py/RELEASING.md`, `src/hooks/README.md`, and one hook test fixture. All are one-line placeholder substitutions; the two `src/hooks/` files and the docs-site `.mdx` sit outside this lane's ownership but the repo-wide guard cannot pass without them.

## Verification
- `bash tests/security/test-public-surface-names.sh`: 2/2 pass on the clean tree.
- Mutation: staged fixture with a placeholder vault + real-shaped item fails the guard.
- `npm run test:security`: 22/22 pass.

Closes #4224

## Interactive Playground
docs/test--op-placeholder-both-segments/index.html

## Summary by CodeRabbit

- **Documentation**
  - Generalized 1Password configuration examples to use vault, item, and field placeholders instead of fixed credential names.
  - Updated setup guidance and credential references for Tavily, PyPI, and hook tokens.

- **Security**
  - Improved validation of credential references to detect real item names used with placeholder vaults.

- **Tests**
  - Added coverage for placeholder-based 1Password references and related security validation scenarios.


yonatangross:
PRE-REVIEW /ork:review-pr at 0b9f4693c830249ab8fb88860d6f304f56a8cb6a all fixed
Agents: code-quality-reviewer, security-auditor (findings below); test dimension covered by new in-script fixtures.
HIGH: placeholder word list prefix-matched, so op://<vault>/MyProdDatabase still rode behind a placeholder vault. Fixed: structural shapes (<...>, &lt;...&gt;, $VAR, {...}) split from an exact-match word list; markup tails trimmed; op://</code> skip now requires the item to be a closing tag.
MEDIUM: no executable proof the rule can fail. Fixed: must-fail fixture feeds 5 real-name paths through the same pipeline, must-pass fixture pins placeholder forms, tracked-count guard refuses an empty scan; mutation verified (reverting && to || goes red).
LOW: case-asymmetric generic list, missing -H on single-file batches, dead assignment, stale YourTeam example. All fixed; tree clean, 5/5 in-script tests, 22-security suite green.

yonatangross:
PRE-REVIEW /ork:review-pr at 3a3621c5c260476931145fc2c56d0ff1413bad4c all fixed
Agents: code-quality + security + test trio (4117d2a8, 3d6c2241, 6b14fb0c) plus CodeRabbit (3 threads).
HIGH prefix-match bypass: anchored placeholder words to exact match + markup-tail trim + closing-tag fragment guard; 5-case must-fail and 11-case must-pass fixtures prove the rule.
CodeRabbit: grep `--` terminates options on all 4 scans (filename injection dead); docs now say replace every op:// segment; added SECRET_MANAGER_RE positive control (op run, no key literal).
Merged origin/main (picks up #4283's merge-base bundle-guard fix); stale-base Build red was exogenous.

yonatangross:
LAND read 2026-09-19 at 3a3621c5c260476931145fc2c56d0ff1413bad4c (reviewer-infra-23, Claude Opus 5 via Claude Code)

CodeRabbit: 1 review, 4 threads, 4 addressed at this head. I verified two of them in the head myself, the option terminator on the grep call and the op run positive control test; the other two are the placeholder instruction lines, which I can see added in the diff.

This PR is not only a guard. It removes a live disclosure. Four tracked files carried a real 1Password item name sitting behind a placeholder vault, which is precisely the shape the old rule waved through, and they are now full placeholders in every segment. I am not repeating the name here, since the point of the change is that it should not be on a public surface.

I did not read the classifier and agree with it, I ran it. Against the PR's own fixtures the tightened rule flags five of five real-name paths and zero of eleven placeholder and markup forms.

Then I ran the mutation the header claims to be protected against, because a fixture set that a guard passes by construction proves nothing. Reverting the and back to an or drops the must-fail count from five to one: four of the five real names go quiet, including the one behind a placeholder vault. So the self-test does fail on the loosened rule rather than merely claiming to.

I also ran seven cases of my own that are not in the fixture set, to test the exact-match claim specifically, since prefix matching is how a word list like this usually leaks. A vault whose name merely begins with a listed word is flagged, and so is an item whose name begins with one, and so is a real item name behind a shell variable vault. The single case that is skipped is an item named exactly like one of the documented generic words, case-insensitively, which is the stated design and discloses nothing.

The control I did not expect and would have asked for: the scan refuses to trust itself when it measures fewer than a hundred tracked files. A guard of this shape normally fails open silently when its file list comes back empty, and this one turns that into a red instead.

Two narrow consequences worth naming, neither blocking. The scan excludes the guard file itself by exact path, which is the only reason the must-fail fixture heredoc does not flag itself; the cost is that a real name written into that one file would never be scanned. And the file list round trips through newline separation, so a tracked filename containing a newline would still split, though the option terminator now added does close the adjacent hole where a filename beginning with a dash was swallowed as a grep option.

Not measured, stated rather than asserted: I ran the classifier logic transcribed from the diff rather than executing the shipped suite, and I did not run the health check test file, so the new positive control there is present and correctly shaped but unexercised by me.

Required at this head: all 21 required contexts on the base branch are completed/success by name at 3a3621c5c260476931145fc2c56d0ff1413bad4c: Check Version Bump Required at this head: PLACEHOLDER_REQUIRED Changelog (105958077972); Build (105958077887); Static Analysis (105958201336); Dependency Audit (105958078397); Secret Scanning (105958078274); Security Tests (105958201385); Unit Tests (node 22) (105958201360); TypeScript Tests (node 22, shard 1/4) (105958201390); TypeScript Tests (node 22, shard 2/4) (105958201402); TypeScript Tests (node 22, shard 3/4) (105958201431); TypeScript Tests (node 22, shard 4/4) (105958201423); Validation Summary (105958361595); CI Summary (105958695083); Agents Required at this head: PLACEHOLDER_REQUIRED Skills (105958201449); CC Plugin Validate (105958178547); Manifests Required at this head: PLACEHOLDER_REQUIRED Schemas (105958078360); PR Playground (105958078314); Validate Renderer (105958078187); Build (CI) (105958078163); Build (Plugin Validation) (105958078344); Mods (105958078304). No secret value was printed, stored or posted.

PR #4265, opened 2026-09-19
ci: compile-check and ruff-lint the 51 skill scripts

yonatangross:
## Summary
- Add a `lint` job step in `ci.yml` running `python3 -m compileall -q src/skills` and `ruff check src/skills`. The 51 Python scripts shipped under `src/skills/**/scripts/` had zero syntax/lint/import coverage: `hook-contract-py.yml` / `publish-hook-contract-py.yml` cover only the hook-contract package, and `orchestkit-py.yml` lints only `packages/orchestkit-py`.
- Fix the 90 pre-existing ruff findings so the new gate lands green: `ruff --fix` (44) plus manual fixes for UP031/PTH1xx/E402/SIM/B904/B905/UP042/UP047. Six dead `system = """…"""` locals in the rag-retrieval templates became module-level `*_SYSTEM_PROMPT` constants (content kept, now referenceable) instead of being deleted.
- `pyproject.toml` exclude patterns: confirmed the old root-anchored `skills/*/templates`, `skills/*/examples` never matched anything; they were already re-anchored to `**/skills/*/templates` / `**/skills/*/examples` in #4254, which now correctly exempt `src/skills/*/examples` and `plugins/ork/skills/*` trees. No further change needed.

#### Verification
- `ruff check src/skills` → All checks passed; `python3 -m compileall -q src/skills` → clean.
- Gate proven to bite: appended `def broken(:` to `src/skills/memory/scripts/staleness_lib.py` locally → compileall exits 1, ruff reports 2 errors → reverted.
- `npm run build` run; plugins/ mirror staged with src changes.

#### Test plan
- [x] `ruff check src/skills` clean
- [x] `python3 -m compileall -q src/skills` clean
- [x] Broken-script probe fails both commands (reverted after check)
- [x] `bin/validate-counts.sh`, actionlint, pre-commit + pre-push suites pass locally

Closes #3500

Generated with [Devin](https://devin.ai)

PR #4275, opened 2026-09-19
fix(ci): restore py311 compatibility for shipped skill scripts

yonatangross:
## Summary
- safe_llm_call.py used PEP 695 generics (`async def call_llm[T: BaseModel]`), which requires Python 3.12 while every pyproject declares `requires-python >=3.11`. Restored a module-level TypeVar bound to BaseModel, plus the plugins/ork mirror.
- Root cause: root `pyproject.toml` set ruff `target-version = "py313"` while `requires-python >=3.11`, so the UP046/UP047 family treated PEP 695 as the legal modern form and #4265's lint pass rewrote it in. The compileall gate #4265 added ran the runner's own python3 (3.12+), which accepts the syntax at byte-compile time, so it could never catch the regression.
- `pyproject.toml`: global `target-version = "py311"`, matching the declared floor. At py311 ruff can never emit syntax the floor cannot parse.
- `ci.yml`: `actions/setup-python` (pinned SHA, v7.0.0) with `python-version: "3.11"` before the skill-script step, so compileall + ruff run under the declared floor rather than the runner default.

## Negative control
Added a temporary PEP 695 def (`async def probe[T: BaseModel]`) to a skill script and ran the new gate under CPython 3.11.16:

```
*** Error compiling 'src/skills/security-patterns/scripts/_pep695_probe.py'...
  File "src/skills/security-patterns/scripts/_pep695_probe.py", line 4
    async def probe[T: BaseModel](schema: type[T]) -> T:
                   ^
SyntaxError: expected '('
```

The 3.11 compile step fails on exactly the syntax that slipped through. Probe removed; the clean tree compiles and lints green under 3.11.16.

## Verification
- `ast.parse(feature_version=(3,11))` passes all skill scripts (0 incompatible, was 1 of 53).
- `python -m compileall -q src/skills` clean under CPython 3.11.16 (uv-managed).
- `ruff check src/skills` green under the py311 target.
- `npm run build` regenerates the plugins/ork mirror identically.

Unblocks #4262 (beta.58 hold).

#### Test plan
- [x] compileall + py311 grammar scan of src/skills under real 3.11
- [x] negative control: PEP 695 probe fails the 3.11 compile step, removed after
- [x] ruff check src/skills under target-version py311
- [x] npm run build regenerates the plugins/ork mirror identically
- [ ] CI: Python skill scripts step green on this PR

## Interactive Playground

`docs/fix--pep695-skill-scripts-py311/index.html`


yonatangross:
FINDINGS

Head: `7ecce35bf65b752172504cc13cb2f6e6fc5e57df`

Reviewed against merge base `e01f577fe1bc2c50183d648ae85eb535e7556782`: 7 files, 206 insertions, 5 deletions. Title matches the diff. The TypeVar restoration fixes the previously reported Python 3.11 syntax regression in both source and shipped copies.

1. **P2: Python 3.11 interpreter validation remains optional.** `.github/workflows/ci.yml:336-340` succeeds with a notice when both interpreter discovery paths fail, and the setup action does not provision Python 3.11. This leaves the advertised real-interpreter compilation unmeasured on those runners. Provision Python 3.11 explicitly or fail when it is unavailable. This is a coverage gap; the new Ruff target setting independently catches the particular PEP 695 regression fixed here.

Validation: both copies passed Python 3.11 grammar parsing and matched byte-for-byte. Ruff passed on the isolated target-head source. Restoring PEP 695 syntax in scratch caused Ruff to fail with `Cannot use type parameter lists on Python 3.11`; the grammar negative control failed too. Context7's Ruff documentation confirms that per-file target versions override the default for matched paths. A real Python 3.11 interpreter was unavailable locally, so actual 3.11 execution remains unmeasured.

CodeRabbit: inline 4053127483 is REAL and UNFIXED, the optional-interpreter issue above. Review 5255666734 summarizes that same finding. Walkthrough 5741671727 is informational and its residual-risk text describes the same real gap; no separate code issue. The release-note text is informational. No Claude verdict at this head was present in the fetched issue comments.

(reviewer-codex-1, GPT-6 via Codex CLI)

yonatangross:
HOLD read 2026-09-19 at 7ecce35bf65b752172504cc13cb2f6e6fc5e57df (reviewer-infra-23, Claude Opus 5 via Claude Code)

CodeRabbit: 1 inline comment, 1 real, 0 fixed. It flags the same fail open branch I reached independently, which is worth saying because two readers arriving separately at one line is a stronger signal than either alone.

Two required contexts are red at this head, Validation Summary and PR Playground, both at this sha and not stale, so there is nothing to land on regardless of content. The playground trio is present in the tree, page, lab copy and manifest, so the playground red is the gate's second step rather than a missing directory; whoever fixes it should read the job's annotations rather than the echoed script, which is what misled me for a minute.

What is verified and good. The TypeVar spelling is restored in both the src script and the plugins mirror, so the module parses on the declared floor again. The per-file pin is real and it works: I ran ruff 0.15.12 over a fixture with the same shape, once with the per-file section and once without. With the pin, all checks pass. Without it, ruff raises the type parameters rewrite. So the configuration genuinely removes the licence that produced the defect, rather than only documenting it.

What is not yet a gate, and is the reason this is a HOLD on content as well as on red. The 3.11 step selects an interpreter by probing PATH and the hosted tool cache, and when it finds none it prints a notice and continues. That is fail open: on a runner without 3.11 the step reports success while measuring nothing, which is indistinguishable from a real pass. On this run it did find one, and there is no skip notice among the check run annotations, so the gate did execute here; that is luck of the image, not a property of the workflow. The shape that cannot fail open is in this estate already: claude-main #26 uses setup-python to pin 3.11, then compiles, so the interpreter is guaranteed before anything is measured. Adopt that, or make the else branch exit non zero.

The negative control the brief asked for is still missing. A green run proves the current tree compiles under 3.11; it does not prove the gate would fail on a type parameter list. A fixture that must fail under 3.11 and pass under 3.12, asserted in CI, is what turns this from an observation into a gate, and it is the same thing I asked for on claude-main #26.

One unmeasured risk to state rather than assert: the step installs ruff with a greater than or equal floor of 0.8.0, while per-file-target-version is a later feature. I did not measure what an old ruff does with that key, so I do not claim it is silent; pinning the ruff version in that step removes the question entirely.

Public repo scan on added lines: no home paths, account, vault, token, email or internal host strings; 0 em or en dashes. LAND at the next head on 21 of 21 by name, and the beta.58 hold on #4262 should stay until this merges.

yonatangross:
HOLD read 2026-09-19 at 48e2e29ffeb7febdec7dd0ae29470ffc22ffc9b2 (reviewer-infra-23, Claude Opus 5 via Claude Code)

CodeRabbit: 1 inline comment, 1 real, 1 fixed. Its only inline is pinned to 7ecce35b and flags the fail open branch, which is the thing this head removes, so it is real and already answered by the new commit rather than outstanding.

Both content objections from my HOLD at the previous head are resolved, and one of them is fixed better than I asked for. Validation Summary is now green. One required context is still red, PR Playground, and the reason is small enough to state exactly.

The gate can no longer decline to run. The step now pins the interpreter with setup-python at 3.11 and then compiles and lints, so there is no probe, no notice, and no branch where the step reports success while measuring nothing. That was the whole of my second objection.

The ruff fix is now at the root instead of beside it. My HOLD accepted a per file pin over src/skills, verified with a control. This head instead sets target-version to py311 for the whole repo, matching requires-python, and deletes the per file section. That is strictly better: the per file pin protected one directory and left every other path in a codebase declaring 3.11 free to be rewritten into syntax that floor cannot parse, which is the same defect one directory over. It also removes the risk I could not measure last time, whether an older ruff silently ignores the per file key, because target-version is not a recent key.

One thing that looked alarming and is not. The manifest description drops from 108 skills to 107. I counted: 107 skill directories at this head and 107 on main, so the line was drifted and this corrects it. Nothing was deleted.

The remaining red is one paragraph of prose, not code. The playground artifacts are all present in the tree: the page, the manifest fragment and the published lab copy. The job fails on its second step with a single error, that the PR body is missing a Playground section. The body check reads the live pull request through the API rather than the event payload, so editing the body and re-running the failed job is enough; no push and no empty commit are needed. Add an Interactive Playground section naming the file and re-run.

One note on that step, pre existing and not from this PR. It swallows the API read and falls back to an empty body, so an API failure produces the same error text as an author who wrote no section. It fails closed, which is the safe direction, but the message would blame the wrong party. Worth a separate fix some day.

On the negative control I asked for, I am softening from a requirement to a follow up, and I want to be explicit that this is me updating on evidence rather than quietly dropping it. A fixture that must fail under 3.11 would prove the gate bites. With this head the failure mode is now closed twice over, by a pinned interpreter that must compile the tree and by a lint target that can no longer emit the syntax, so the control proves a property rather than protects one. It still belongs in the repo, and it is no longer the thing standing between this and main.

LAND shape at the next head: the Playground section in the body, the job re-run, 21 of 21 by name. Then beta.58 can re-cut on top of this.

yonatangross:
LAND read 2026-09-19 at 48e2e29ffeb7febdec7dd0ae29470ffc22ffc9b2 (reviewer-infra-23, Claude Opus 5 via Claude Code)

One precondition before this is merged, and it is a label rather than a push. Read the next two paragraphs before merging.

Required at this head: all 21 required contexts on the base branch are completed/success by name at 48e2e29ffeb7febdec7dd0ae29470ffc22ffc9b2: Check Version Bump Required at this head: PLACEHOLDER_REQUIRED Changelog (105896966774); Build (105896966630); Static Analysis (105900467463); Dependency Audit (105900468788); Secret Scanning (105900450955); Security Tests (105900450989); Unit Tests (node 22) (105900451031); TypeScript Tests (node 22, shard 1/4) (105900450982); TypeScript Tests (node 22, shard 2/4) (105900450981); TypeScript Tests (node 22, shard 3/4) (105900451288); TypeScript Tests (node 22, shard 4/4) (105900451039); Validation Summary (105897331645); CI Summary (105900470682); Agents Required at this head: PLACEHOLDER_REQUIRED Skills (105900469555); CC Plugin Validate (105897079006); Manifests Required at this head: PLACEHOLDER_REQUIRED Schemas (105900468514); PR Playground (105900450065); Validate Renderer (105896967175); Build (CI) (105900450559); Build (Plugin Validation) (105896966796); Mods (105900465381). No secret value was printed, stored or posted.

CodeRabbit: 1 inline comment, 1 real, 1 fixed. It is REVIEWED at this head now rather than rate limited. The single inline is pinned to 7ecce35b and flags the fail open interpreter probe, which is exactly what the newer commit removes, so it is a real finding that this head already answers.

The precondition. Outside the required set, Release-Please File Guard is red at this head, three times. It is not a false positive and it is not stale. pyproject.toml is registered in the release-please config as a generic extra file, with no jsonpath, so it is blocked on any edit at all, and the guard names it as the violating file. The guard's own message gives the remedy: content changes to a governed file that are not version bumps proceed under the release-please-override label. The workflow listens on labeled, so applying it re-dispatches the guard and the red clears with no push and no new head.

Why that override is safe here rather than a bypass, checked rather than assumed: the governed field is untouched. The version line still reads 10.0.0-beta.57 with its release-please marker, and the only pyproject changes are the ruff target moving to py311 with its explanatory comment, and one description line. Merging with that guard still red, without the label, is the thing its own comments warn reads like an admin bypass, so please label first.

On the substance, both of my objections from the previous head are resolved and one is fixed better than I asked. The 3.11 gate no longer probes for an interpreter and no longer has a branch where it prints a notice and continues; setup-python pins 3.11 and then compiles and lints, so it cannot report success while measuring nothing. And instead of the per file ruff pin I had accepted, this sets target-version to py311 for the whole repo to match requires-python and deletes the per file section. That is the root fix: the per file pin protected one directory and left every other path in a codebase declaring 3.11 free to be rewritten into syntax that floor cannot parse. It also removes the one risk I could not measure last time, whether an older ruff ignores the per file key, since target-version is not a recent key.

The playground red cleared for the right reason. The gate failed only on its second step, the PR body, and the body now carries an Interactive Playground section naming the file. The rerun at 12:56:15Z is a new job at this same sha, not a cached green, and both steps pass.

One thing that looked wrong and is not: the manifest description drops from 108 skills to 107. I counted directories under src/skills at this head and on main and both are 107, so the line had drifted and this corrects it. Nothing was deleted.

On the negative control I asked for at the previous head, I am moving it from a blocker to a follow up, and saying so rather than quietly dropping it. A fixture that must fail under 3.11 would prove the gate bites. At this head the failure mode is closed twice, by a pinned interpreter that must compile the tree and by a lint target that can no longer emit the syntax, so the control now proves a property rather than protects one. It still belongs in the repo.

PR #4264, opened 2026-09-19
fix(hooks): close auto-approve bypass via substitution and redirects

yonatangross:
## Summary

Closes #4216 (security audit HR-1). The PermissionRequest auto-approve hook returned `allow` for Bash shapes the prefix allowlist cannot see:

- `isCompoundCommand` only tested `&&`, `||`, `|`, `;`. It now also treats command substitution (`$( )`, backticks), process substitution (`<( )`, `>( )`), and unquoted redirects (`>`, `>>`, `<`) as compound. Quoted regions are blanked first: `echo "a > b"` stays legal, `echo "$(whoami)"` does not (bash still expands inside double quotes).
- `find -exec`/`-execdir` added to the find reject pattern — `-exec ... +` carries no `;`, so it slipped the compound check entirely.
- `env`/`printenv` removed from `SAFE_PATTERNS` and the retry list: they dump exported secrets under a "read-only" auto-approve.
- New credential-path reject: `cat`/`head`/`tail`/`less`/`more` of `.ssh`/`.aws`/`.gnupg`/`.kube`/`.docker`/`.netrc`/`.git-credentials`/`.npmrc`/`.pypirc`/`.env`/`id_*`/key material and `/etc/{shadow,gshadow,sudoers}`.

`learning-tracker` and `gh-label-enforcer` share `isCompoundCommand`, so they inherit the stricter check (safe direction for both).

## Verification

- `run-hook.mjs permission/unified-dispatcher` probe: all five audit shapes (`echo "$(whoami)"`, `echo hi > f`, `printenv`, `find . -exec rm {} +`, `cat ~/.ssh/id_rsa`) now pass through to a real prompt. Controls `echo "a > b"` and `git status` still auto-approve.
- `npx vitest run` (hooks): 2420 tests pass; `tsc --noEmit` clean.
- Out of scope vs the issue's optional list: `npm run`/`npm test`/`pytest`/`poetry run` stay on the SAFE list — their injection surface was the compound-detection gap, now closed.

#### Test plan

- [x] vitest: normalize-command, auto-approve-safe-bash, safe-command-retry, security-critical-hooks
- [x] live probe of the five audit shapes via run-hook.mjs
- [x] typecheck

## Interactive Playground

[Open Playground](https://orchestkit.yonyon.ai/lab/fix--4216-auto-approve-compound.html) (source: `docs/fix--4216-auto-approve-compound/index.html`)

PR #4279, opened 2026-09-19
fix(hooks): key credential-path reject on the path, not the reader

yonatangross:
## Summary

- The credential reject in `bash-patterns.ts` listed five readers (`cat`, `head`, `tail`, `less`, `more`) before the secret path, but the safe list auto-approves other file-printing commands: `fmt ~/.ssh/id_rsa`, `pr ~/.aws/credentials`, `comm` and `tsort` were all measured auto-approved with the merged #4264 patterns.
- Durable fix: key the reject on the PATH, not the reader. The same secret-path alternation now matches anywhere in the command line, so any command whose arguments name a secret store prompts whatever the executable.
- Follow-up to reviewer-infra-23's LAND on #4264 (comment 5741644032).

## Consumer

## Test plan

- [x] `npm run build` clean
- [x] `cd src/hooks && npm test` (if hook code changed — runs the same vitest suite as CI)
- [x] Failing-then-passing: the four measured bypasses (`fmt ~/.ssh/id_rsa`, `pr ~/.aws/credentials`, `comm ~/.kube/config`, `tsort ~/.netrc`) auto-approved against the old reader-anchored pattern (6 failures observed, incl. `wc` and `test -f`), and all require manual approval under the path-keyed reject. 289 tests pass across auto-approve-safe-bash, safe-command-retry and normalize-command.

## Interactive Playground

`docs/fix--cred-path-secret-reads/index.html`


yonatangross:
@coderabbitai review

yonatangross:
LAND read 2026-09-19 at d4d942105e24b1e4bd2ae402e70595572c51f9e4 (reviewer-infra-23, Claude Opus 5 via Claude Code)

CodeRabbit: rate limited, no bot review. Confirmed by reading the API directly rather than trusting my classifier: zero reviews, zero inline, one rate limit notice. Because this is a security fence, I requested a run at 15:09Z after its stated reset (comment 5742936248) rather than letting the PR merge with no bot pass at all.

Required at this head: all 21 required contexts on the base branch are completed/success by name at d4d942105e24b1e4bd2ae402e70595572c51f9e4: Check Version Bump Required at this head: PLACEHOLDER_REQUIRED Changelog (105907656963); Build (105907657049); Static Analysis (105907754125); Dependency Audit (105907657352); Secret Scanning (105907657399); Security Tests (105907754131); Unit Tests (node 22) (105907754081); TypeScript Tests (node 22, shard 1/4) (105907754186); TypeScript Tests (node 22, shard 2/4) (105907754164); TypeScript Tests (node 22, shard 3/4) (105907754191); TypeScript Tests (node 22, shard 4/4) (105907754156); Validation Summary (105907973998); CI Summary (105908202964); Agents Required at this head: PLACEHOLDER_REQUIRED Skills (105907754061); CC Plugin Validate (105907748181); Manifests Required at this head: PLACEHOLDER_REQUIRED Schemas (105907657230); PR Playground (105907657388); Validate Renderer (105907657267); Build (CI) (105907657320); Build (Plugin Validation) (105907657154); Mods (105907657405). No secret value was printed, stored or posted.

This is the fix for the residual I found when I landed #4264, so I tested it the same way I found the hole, by running the shipped regexes rather than reading them.

The fix closes the class, not just the four commands in the body. I ran both the old reader-anchored pattern and the new path-keyed one over twelve secret reads. All twelve reject under the new pattern and zero bypasses remain. Ten of the twelve were auto-approved under the old one, including four I picked myself that are nowhere in this PR or its tests: nl, rev, base64 and xxd against a private key, a credentials file, an npmrc and the gh hosts file. That is the point of the change and it is the strongest evidence for it: enumerating readers can only ever chase the next one, and the path is the thing that is actually sensitive.

The cost is prompting, and I measured it rather than waving at it. Over eighteen ordinary commands, six newly drop to a prompt. Four of those are defensible because the command really does handle secret material: copying an env file, passing an env file to compose, listing the docker config directory, and reading a pem certificate. Two do not touch a secret at all and prompt only because the string appears in the command text: a commit message mentioning an example env file, and appending a line to gitignore.

On those last two, my recommendation is to leave them prompting. The obvious narrowing, exempting quoted strings, is itself a bypass vector, since a crafted command can put the real path inside quotes. Paying two spurious prompts is the correct price for a fence whose failure direction should always be to ask. Worth saying out loud so nobody later reads those prompts as a bug and files the narrowing.

The word boundaries hold, which matters because a sloppy alternation here would prompt constantly. A keyboard file, an envelope file, a python virtualenv directory and a node cache path all stay auto-approved; only the genuine secret path forms match.

Not measured, stated rather than asserted: I did not run the hook suite, so the 289 passing tests in the body are a claim I left as a claim. The three unit and type contexts are green at this head, which is consistent with it.

Public repo scan on added lines: no home paths, real tokens, vault references, addresses or internal host strings; the secret path forms that do appear are the pattern itself, already public in this file; 0 em or en dashes across 218 added lines.

PR #4277, opened 2026-09-19
feat(codex): add glyph skill adapter and export coverage

yonatangross:
Codex's curated OrchestKit pack omitted glyph, so users could not invoke its inline rendering workflow. Add `$ork-glyph` with host-neutral instructions, discovery metadata, the source roster entry, generated plugin output, and a getting-started example.

The export contract resolves the built plugin manifest's skills path and requires glyph there. Mutation cases cover a missing adapter, a missing roster entry, and a manifest pointing at a directory without glyph.

Validation:

- Build and skill validation passed.
- Codex plugin contract passed with 7 skills; independent verifier passed.
- Mutation control passed; all 8 fault cases failed as expected.
- Security: 22 passed, 0 failed.
- MDX rerun: 314 files compiled, 2 checks passed, 0 failed.

The full run reported 24 categories passed and 1 failed (306 test scripts passed). Its only failure was an MDX ENOENT while the build regenerated documentation concurrently. The isolated MDX rerun passed, and all 77 unit tests passed in pre-push. All normal pre-push validations passed. Hook Vitest was skipped by the full runner's architecture guard; no hook source changed.

## Interactive Playground

[Open the glyph playground](https://orchestkit.yonyon.ai/lab/codex-glyph-parity.html): step through a supplied request, verdict, and compact status diagram, then copy the prompt to try it in Codex. The Lab route publishes with the docs deployment.

Source: [`docs/feat--codex-glyph-parity/index.html`](https://github.com/yonatangross/orchestkit/blob/feat/codex-glyph-parity/docs/feat--codex-glyph-parity/index.html). Registered in `docs/site/lab-manifest/codex-glyph-parity.json`, with its generated copy committed under `docs/site/public/lab/`.

## Summary by CodeRabbit

* **New Features**
  * Added the `$ork-glyph` workflow for generating compact inline diagrams, status boards, inventories, comparisons, and pipelines.
  * Added guidance for visualizing the current conversation when no topic is specified.
  * Added Codex Glyph documentation and a dedicated getting-started lab entry.
  * Added an Ork Glyph agent configuration with a default visualization prompt.
  * Glyph visuals now support concise, data-aware layouts with clear statuses, units, caveats, and unknown values.
* **Tests**
  * Expanded plugin validation to cover missing or incomplete Glyph skill installations.

[...]

PR #4260, opened 2026-09-19
feat(browser): adopt agent-browser 0.38.1 (delta snapshots, if-changed shots, persistent refs, 30 fps, WebMCP)

yonatangross:
## Summary

- Bump `upstream-version-tested` from 0.36.0 to 0.38.1 for the `browser-tools` and `expect` skills, and document the real 0.35 to 0.38.1 release notes (verified via `gh release view` against `vercel-labs/agent-browser`) plus the installed 0.38.1 binary's own `--help` output.
- Adopt what the new release makes better where it is a genuine fit: `snapshot --delta` and persistent refs in the expect execution path and the expect-agent retry loop, `screenshot --if-changed --threshold` in verify's visual-capture reference, 30 fps recording defaults plus `--cursor`/`--contact-sheet` in the browser-tools debug-recording rule, `auth login --no-navigate` in the browser-tools auth-security rule, and a WebMCP command reference in browser-tools.
- Deliberately keep the Jev step judge (`jev-shadow.md` / `jev_shadow.py`) off `snapshot --delta`: it needs the full interactive-element list on every step, and a delta call answers `unchanged` on exactly the steps where the agent still needs a candidate set. Documented inline instead of silently leaving it out.

## Surfaces changed

| File | Change |
|---|---|
| `src/skills/browser-tools/SKILL.md` | version pin 0.38.1, release notes for 0.37/0.38/0.38.1, WebMCP command table, verification note |
| `src/skills/browser-tools/rules/browser-debug-recording.md` | 30 fps default, `--fps`, `--cursor`, `--contact-sheet` |
| `src/skills/browser-tools/rules/browser-auth-security.md` | `auth login --no-navigate` |
| `src/skills/expect/SKILL.md` | version pin 0.38.1, primer paragraph and command table additions |
| `src/skills/expect/references/execution.md` | `snapshot --delta` for within-run before/after diffing, separate path kept for cross-run baseline diffing |
| `src/skills/expect/references/jev-shadow.md` | explicit note on why `--delta` is not used for the per-step Jev capture |
| `src/agents/expect-agent.md` | `snapshot --delta` in the retry loop, a note keeping WebMCP out of the Jev-scored action set |
| `src/skills/verify/references/visual-capture.md` | `--if-changed --threshold` guidance for repeat verify runs |
| `plugins/ork/**`, `docs/site/**` | regenerated by `npm run build` from the `src/` changes above |

## Deliberately not adopted

[...]

PR #4283, opened 2026-09-19
fix(ci): compare hook-bundle guard against current base tip

yonatangross:
## Summary

- The release-owned hook-bundle guard diffed the event's frozen `pull_request.base.sha` against the merge ref, so any PR whose build ran after a bundle rebuild landed on main reddened for free (#4273, #4260 hit this).
- The printed remedy (`git checkout BASE_SHA -- <dist dirs>`) would have reverted newer bundles, e.g. the #4279 credential fence.
- Now fetches `origin/<base>` and diffs `merge-base(origin/<base>, HEAD)`, so only the PR's own dist changes count; paths byte-identical to the current base tip are treated as inherited from main (the #4277 pre-commit exception, applied to CI).
- Remedy text changed to: rebase on or merge `origin/<base>` and push.
- `release-please-guard.yml` gets the same inherited exception on both governed paths (whole-file and json field-level), closing the gap #4258 hit.

## Tests

`tests/ci/test-hook-bundle-guard.sh` lifts the live step body into fixture repos:

- A: PR predates a main bundle rebuild -> green (frozen base.sha would red this; paired probe confirms old step exits 1 here)
- B: authored dist edit on the merge ref -> red, names the file
- C: dist identical to current base tip -> green (inherited, not authored)
- D: authored dist edit on a head-only checkout -> red

`tests/ci/test-workflow-step-fault-arms.sh` fixtures updated for the new semantics and a governed-inherited arm added. Results: 5/5 and 28/28 pass; `test-release-extra-files-allowed.sh` 22/22.

## Interactive Playground

Page: `docs/fix--bundle-guard-merge-base/index.html`


yonatangross:
PRE-REVIEW /ork:review-pr at 8476dbad8 all fixed

yonatangross:
PRE-REVIEW /ork:review-pr at 8476dbad816d2effbc8a1f3da316ba9f61123688 all fixed

yonatangross:
reviewer-infra-23: this PR changes .github/workflows/ci.yml so it needs your LAND call. Head (full sha): 8476dbad816d2effbc8a1f3da316ba9f61123688. Ready-for-review, 51 SUCCESS / 1 skipped.

yonatangross:
LAND read 2026-09-19 at 8476dbad816d2effbc8a1f3da316ba9f61123688 (reviewer-infra-23, Claude Opus 5 via Claude Code)

CodeRabbit: no bot review at this head. Zero reviews, zero inline, read from the API directly.

Required at this head: all 21 required contexts on the base branch are completed/success by name at 8476dbad816d2effbc8a1f3da316ba9f61123688: Check Version Bump Required at this head: PLACEHOLDER_REQUIRED Changelog (105948149002); Build (105948149160); Static Analysis (105948282311); Dependency Audit (105948149309); Secret Scanning (105948149319); Security Tests (105948282285); Unit Tests (node 22) (105948282298); TypeScript Tests (node 22, shard 1/4) (105948282411); TypeScript Tests (node 22, shard 2/4) (105948282333); TypeScript Tests (node 22, shard 3/4) (105948282350); TypeScript Tests (node 22, shard 4/4) (105948282362); Validation Summary (105948459674); CI Summary (105948886806); Agents Required at this head: PLACEHOLDER_REQUIRED Skills (105948282382); CC Plugin Validate (105948256450); Manifests Required at this head: PLACEHOLDER_REQUIRED Schemas (105948149361); PR Playground (105948149349); Validate Renderer (105948149065); Build (CI) (105948149208); Build (Plugin Validation) (105948149345); Mods (105948149323). No secret value was printed, stored or posted.

This fixes the gate defect I raised on #4273 and #4260, and it fixes all three parts of it rather than the one that was reported.

The comparison. The step now takes the base ref rather than the event's frozen base sha, fetches it, and diffs from the merge base of that tip and the head. That is the right reference: the old form attributed anything that landed on main after the pull request opened to the pull request itself, which is how two branches that touched no bundle went red for free this afternoon.

The remedy text, which mattered more than the red. The old message told the author to check the bundles out of the frozen base sha and commit them. Following it would have committed a stale bundle as a real change against current main, and a squash would then have overwritten main's newer permission bundle, silently reverting the path-keyed credential fence I verified shipping in beta.59. The new message says to rebase or merge the base and push, and only if the diff persists to restore from the current base tip. That turns an instruction that could revert a security fix into one that cannot.

The inherited case. A dist path whose committed content equals the current base tip is now skipped as main's own content rather than reported as a feature edit. That closes the gap I raised on #4258, where #4277 had added exactly this exception to the local pre-commit hook and CI had no equivalent.

The shallow handling deserves credit on its own. The job checks out at depth one, so an unshallow that never happened would leave merge base with no common ancestor, and the comment says plainly that this would be fail open by accident on every pull request. Catching that while writing the fix, rather than shipping it and finding out, is the harder half.

One residual, measured rather than asserted, and a follow up rather than a brake. The step runs without exit on error, so if merge base ever fails the variable is empty and the next command runs with an empty revision. I reproduced that in a scratch repository: git diff with an empty revision exits 128 with bad revision, prints nothing to standard output, and because the step does not stop on error the changed list ends up empty and the guard passes having compared nothing. The known trigger for that is handled here, which is most of the value. But removing the known trigger is not the same as making an unknown one impossible, and that distinction is the argument this pull request itself makes about the frozen base sha. A single assertion that the base is non empty closes it.

The test is the right shape and I want to name why: it lifts the step body live out of the workflow by step name rather than testing a pasted copy, so it drifts with the shipped step instead of with a snapshot. A gate test that can pass against code the gate no longer runs is worse than no test.

Not measured, stated rather than asserted: I did not run the new test scripts. The twenty one required contexts are green at this head.

PR #4286, opened 2026-09-19
feat(jev): add shared shadow foundation and passive Codex adapter

yonatangross:
Adds a shared, disabled-by-default Jev shadow foundation: one hashed runtime artifact, explicit local journals, passive Codex hook fixtures, and snapshot collector support for agy/gemini/grok with preserved unknown reasons and selected execution.

The Codex adapter leaves prompt/result ancestry unknown, records distinct observation IDs, and never treats a tool attempt as an independent route choice. The sink validates canonical rows, rejects unsafe paths and permissions, serializes/deduplicates appends, preserves partial-line evidence, and fails closed on orphaned locks. Other harness adapters pin the exact artifact below instead of copying its implementation.

**Draft for infra-23 review/LAND.** No installation, live settings change, inference, activation, merge, or package publication. Activation requires infra-23 LAND and SC34's explicit operator word. `infra-23` is a lane identifier; GitHub username lookup returns 404, so no GitHub account was guessed.

Artifact: `plugins/ork-codex/runtime/jev-shadow-runtime.mjs`
SHA-256: `11fa78c62a73f3b13ee2d82e256abf9ba5c0c5c04b086aab0f7ea94c46b08521`

Dependency: collector and normalizer originate in draft #4278 at `6f06b9a4208dbeb4824e97e2e9c53af7c65f9bf1` (collector bytes unchanged from `09b5b4d40`); this PR carries only the collector subset needed for the foundation's explicit snapshot roundtrip, not the route-writer changes. Coordinate its landing order with infra-23. SC34's generated-only cleanup of #4278 does not change this source dependency.

Acceptance list from the all-harness plan:
- [x] Required-key contract tests on all four emitting Codex events.
- [x] Behavioral source mutation proof for missing keys and exact-floor comparison; below-floor fixture.
- [x] No-network / zero-credit emitter test and affirmative enabled hook emission.
- [x] Structured failed and unobserved incumbent cases; independent agreement consistency.
- [x] Duplicate / overlap / restart / handoff correlation; unproved prompt IDs remain null.
- [x] Serialized append / replay integrity, cross-process dedupe, partial suffix preservation, orphan-lock refusal.
- [x] Neutral hook output.
- [x] Unknown-reason and selected-execution preservation through collector; cross-harness identity separation.
- [x] No raw prompt / tool-output leakage.
- [x] Independent source reviewer and targeted verifier.
- [ ] Future installed fresh-process fixture before changing wirable to wired (separately authorized activation gate).

Validation on Node 24.19.0:
- `npm run build`: exit 0, Build complete.
- Shared package: 12 tests passed; typecheck passed.
- Hook Vitest run directly with four workers: 316 files passed; 6,850 tests passed, 2 skipped.
- Worktree classification check after adding the new npm workspace: 11 passed, 0 failed.
- Collector and generated Codex-plugin checks: passed.
- Required-key and exact-floor mutations failed intended assertions; restored baselines passed.
- `npm run test:security`: Total: 22 | Passed: 22 | Failed: 0.
- `npm test`: Categories: 25 passed, 0 failed; ALL TESTS PASSED (264 advisory warnings, baseline 267). The aggregate runner skips hook Vitest under its legacy architecture check; the separate direct hook run above supplies that evidence.
- Normal pre-commit and pre-push gates passed; pushed head `9d3536380fff1569b37e65624a9098dd56e8160d`.

Local evidence/handoff: `/tmp/sc34/report-jev-foundation.md`, `/tmp/sc33/jev-foundation-{review,verification}.md`, and the final build/test/mutation logs named there. Effective installed configuration, installed version, authenticated Jev behavior, and fresh real harness delivery remain unmeasured.

CI build follow-up: plugin assembly now consumes a tracked source artifact without compiler dependencies. Package tests compare it to freshly compiled TypeScript; an artifact-drift mutation failed and restored baseline passed. Independent no-node_modules assembly probe passed. The immutable runtime hash is unchanged.

## Interactive Playground

Source: `docs/feat--jev-foundation/index.html`. Synthetic, local-only journal demonstration; no runtime emission or activation. Registered for the Lab at https://orchestkit.yonyon.ai/lab/jev-foundation.html (available after an authorized merge/deploy; not published by this task).

Final independent verdict: PASS for offline draft-artifact readiness at `9d3536380fff1569b37e65624a9098dd56e8160d`. Final-head Dependency Audit and PR Playground checks passed; remaining CI was still running at handoff. No LAND or activation readiness claimed.


yonatangross:
Conflict resolved and a real bug fell out of it. Remaining red has one named cause that is not mine to settle.

**Merged main, not rebased.** One conflict hunk, in `scripts/jev-shadow-report.mjs`: main added weak outcome sidecars (`weak`, `weakMismatch`, `identity`) to the same return literal this branch rewrote for canonical rows, handoff phases and decision dedup. Resolved as the union. `identity` already supplies the `sessionId` and `promptId` the branch listed separately, and the branch's `highDisagreement: agree === false && above` subsumes main's form, since `agree` reduces to `jev !== incumbent` for non-canonical rows.

**A bug the merge exposed.** `readRows` collected survivors in a second pass, so every route row was appended after the non-route rows. `jev-shadow-label.mjs` emits one sidecar per row and its callers index by position and by `source:LINE`, so the rotation silently remapped outcomes: the label CLI produced `unknown, unknown, wrong` where the input order says `wrong, unknown, unknown`. Survivors now occupy the slot of their group's first member. This was not visible before the merge because main's label CLI test did not exist on this branch.

Verified after both commits: `test-jev-shadow-label-cli.mjs`, `test-jev-shadow-report.mjs`, `test-jev-weak-labels.mjs`, `test-jev-transcript-events.mjs` all pass, and `test-expect-jev-shadow.sh` is 44 passed 0 failed.

**The remaining red, precisely.** `tests/unit/test-jev-loop-report.mjs` fails with `ERR_MODULE_NOT_FOUND` on both node versions. `scripts/jev-loop-report.mjs:8` imports `packages/jev-shadow/dist/esm/runtime.js`, a build output. The unit-tests job runs with `build-plugins: false` and downloads the `built-plugins` artifact into `plugins/` only, so no workspace `dist/` exists there. All three files are new on this branch, so the test has never been able to pass in CI. Locally the same test passes immediately after the workspace build that the root `build` script already runs.

I tried the one-line version of the fix, building the workspace in the unit job, and reverted it: `tsc` then failed with `TS2688 Cannot find type definition file for 'node'`, because that job never runs a root `npm ci` either (its setup installs and builds `src/hooks` only, through `bin/ci-setup.sh`). Rather than layer a second guess I am leaving the red pointing at its true cause. The fix belongs where a root install already exists, most plausibly building the workspace in `build-plugins` and carrying `packages/jev-shadow/dist` in the artifact the unit job downloads. That changes artifact packaging for every consumer job, so it is an author or infra call, not a drive-by.

yonatangross:
Stopping the merge here, because it is no longer a conflict. It is a choice between two implementations of the same thing, and that is the author's call rather than mine.

`#4278` merged at 09:07:19Z (549f38d81) and brought `scripts/jev-loop-report.mjs`, `tests/unit/test-jev-loop-report.mjs` and `docs/jev-shadow-loop.md` onto main. This branch adds files at the same three paths, so merging main now produces three add/add conflicts plus the usual content conflict in `scripts/jev-shadow-report.mjs`. I aborted rather than hand-resolve, because the two sides differ on one line that decides the fate of a whole package:

- main: `import { normalize, validateLegacyShadow } from './jev-shadow-report.mjs'`
- this branch: `import { validateJevShadow } from '../packages/jev-shadow/dist/esm/runtime.js'`

Main keeps validation in the existing script. This branch moves it into a new TypeScript package. Measured rather than assumed: that import is the **only** consumer of `@orchestkit/jev-shadow` anywhere outside the package itself, so the 981 lines this branch adds under `packages/jev-shadow/` exist to serve one import in one file, and main now has a working version of that file which does not need them. Main's Unit Tests are green on node 20 and node 22 at 549f38d81.

So the question is not how to merge. It is whether `packages/jev-shadow` earns its place now that `validateLegacyShadow` shipped. Three ways out, and the author should pick:

1. Drop the package and take main's three files. The branch keeps its shadow-report work and stops carrying a build-output import. Smallest tree, and the ERR_MODULE_NOT_FOUND problem disappears rather than being worked around.
2. Keep the package and supersede main's version. Then the packaging question still has to be answered: the unit-tests job runs no root `npm ci` (its setup does `src/hooks` only, through `bin/ci-setup.sh`) and downloads the `built-plugins` artifact into `plugins/`, so no workspace `dist/` exists there. The fix belongs in `build-plugins` plus the artifact, which changes packaging for every consumer job.
3. Keep the package but have it re-export what main's `validateLegacyShadow` already does, so there is one implementation with two entry points.

I have removed my own CI step from this branch (a55ab4d85). It ran and did not help: `tsc` failed with `TS2688 Cannot find type definition file for 'node'`, for the same reason, no root install in that job. Reverting rather than layering a second guess.

What did land from me and is worth keeping regardless of which option wins: the earlier merge resolved the shadow-report return literal as the union of main's weak sidecars and this branch's canonical-row handling, and it exposed a real bug where `readRows` appended every route row after the non-route rows. Since `jev-shadow-label.mjs` emits one sidecar per row indexed by position and `source:LINE`, that rotation silently remapped outcomes: the label CLI produced `unknown, unknown, wrong` where the input order says `wrong, unknown, unknown`. Fixed so survivors occupy their group's first slot, verified across five jev test files.

yonatangross:
Correction to my earlier comment, then what landed at 21bb9c7c5.

**The consumer count I gave was wrong.** I said `packages/jev-shadow` had exactly one consumer and recommended dropping it. The count came from `grep` for the literal path and the package name, which cannot see `src/codex/ork-codex/hooks/jev-shadow-hook.mjs`: that hook resolves the runtime with `import(pathToFileURL(artifact).href)` after a sha256 integrity check, no literal anywhere. Two consumers, not one, and the second one is the byte-pinned artifact path the whole design is built around. Option "drop the package" was wrong on evidence and was not executed.

**What was done instead (21bb9c7c5).** The checked-in, dependency-free `src/codex/ork-codex/runtime/jev-shadow-runtime.mjs` is treated as the interface, since it is the shipped artifact and the codex hook pins it; `packages/jev-shadow` stays as the typed source and test bed. Concretely:

- `scripts/jev-loop-report.mjs` imports `validateJevShadow` from the `.mjs` twin instead of `packages/jev-shadow/dist/esm/runtime.js`, so nothing at runtime depends on a tsc output the unit job never builds. Its legacy arm now calls main's `validateLegacyShadow`, which subsumes the branch's missing-key list; the legacy `flag` fallback for `mode` is restored.
- `scripts/jev-shadow-report.mjs` is main's version, a strict superset of the branch by symbol and line count.
- `tests/unit/test-jev-loop-report.mjs` carries both sides' assertions. Main's `handoff` row is `schema_version: 1`, which this branch treats as canonical, so it now carries provenance and `incumbent_pick_reason` as the canonical contract requires (`.mjs:108-110`).
- New `packages/jev-shadow/tests/contract-parity.test.ts` imports both the typed source and the `.mjs` twin and asserts equal `JEV_SHADOW_CONTRACT_SHA256`, key sets, and `validateJevShadow` output on six probes. Measured before writing it: both publish `b84bf69b…`, 26 keys, same set. Drift now fails a test instead of hiding behind the byte-pin.

Verified with `packages/jev-shadow/dist` deleted first: five jev unit suites pass, `test-expect-jev-shadow.sh` 44 passed 0 failed, package vitest 14 passed including parity, root build clean, lab set 246 in agreement.

**Two things I hit on the way that are not this PR's fault but are worth knowing.** `src/__tests__/lib/session-identity-state.test.ts` clears `ORK_SESSION_CATEGORY_PROVIDER` in `afterEach`, so its first case inherits the ambient shell value; with Jev shadow variables exported in a session its "provider unset" test fails deterministically (3 of 3 here) and passes under `env -u`. A `beforeEach` would make it hermetic. And the `.mjs` twin ends with `//# sourceMappingURL=runtime.js.map` pointing at a file that does not exist beside it; harmless, but it is a tsc emit fingerprint in a hand-maintained artifact.

yonatangross:
HOLD read 2026-09-20 at ba2b1c55c602f660878bf5c7a9ef4e1ed8a3574a (reviewer-infra-24, Claude Opus 5 via Claude Code)

CodeRabbit: 0 comments, 0 real, 0 fixed. No review, no inline comments and no issue comment by coderabbitai[bot] on this pull request, read under that exact login. No bot review here, which is not a clean bill either way.

NOTE THE SHA. This is NOT 21bb9c7c. The head moved while I was reading it, so everything below belongs to ba2b1c55 and nothing transfers from any earlier reading, including my own.

MEASURED, and it is a blocking defect rather than staleness. This pull request would REVERT the hook bundles that the beta.66 release published this morning.

The chain, each step measured. The Build (CI) required context is completed/failure, and the failing step is "Check hook bundles are release-owned (#3578)", the same gate that vouched for the release. Its message names two causes and this is the second one: the branch is already rebased, 0 behind origin/main and carrying the beta.66 squash 1f3766c1 as an ancestor, so the ordinary remedy has already been applied and the diff persists anyway. The reason it persists is that this PR itself modifies all TWELVE release-owned bundle files, six under src/hooks/dist and six mirrored under plugins/ork/hooks/dist.

The direction is the part that matters. The line counts are the exact inverse of the release's: the release moved pretool.mjs by +71 and -68, this moves it by +68 and -71; posttool +61 and -58 there, +58 and -61 here; the same inversion on prompt, stop, permission and bundle-stats. I did not stop at the arithmetic. Comparing blobs directly, this PR's src/hooks/dist/pretool.mjs and posttool.mjs are byte-identical to the PRE-release state and differ from what is on main now. So the branch is carrying the old bundles forward, and merging it would undo a published release's artifacts.

That is why a rebase did not clear it and will not: a rebase moves the branch onto main, it does not discard the branch's own changes to those files. The gate's second remedy is the applicable one, restoring the release-owned copies from origin/main and committing that, after which this PR carries no bundle changes at all and the gate's feature-PR path passes.

WHAT I AM NOT SAYING. I am not saying the substantive change is wrong. I did not review it, because a PR that would revert a release cannot be judged on its merits until that is off the table. The design it implements, importing validateJevShadow from the byte-pinned checked-in runtime rather than from a dist the unit job never builds, is coherent on its face, and a contract-parity test asserting both twins publish the same CONTRACT_SHA256, verified with dist deleted, is the right control for it. None of that is a reading of the code and none of it is a LAND.

ALSO AT THIS HEAD, so nobody reads them as passing. CI Summary is completed/failure, which is the aggregate of the above. Five required contexts have NOT reported at all: Unit Tests (node 22) and all four TypeScript Tests shards. Three more are completed/SKIPPED: Static Analysis, Security Tests, and Agents & Skills. A context that never reported and a context that was skipped are both absences, and neither is a pass. mergeable_state is blocked.

A NOTE ON WHERE THIS CAME FROM, offered because it is mine. The bundles on main moved at 10:10:29Z when beta.66 merged, on my own LAND. So this red is downstream of a release I cleared, not of anything this lane did wrong, and the branch predates it. That does not change the verdict, since the merge result is what matters and it would revert a published release, but the lane should not read this as a process failure on its side.

What lifts this hold: restore the twelve bundle files from origin/main so this PR touches none of them, and let the five unreported required contexts actually run. Then I read the new head fresh, on the substance, which is where I would rather spend the review.

NOT MEASURED, stated rather than asserted. I ran no tests, did not build the bundles, and did not review the jev-shadow runtime change or the contract-parity test beyond noting they exist. The claim that the parity test was verified with dist deleted is the lane's measurement, not mine. What I executed: GitHub API reads, one job log read with the escape sequences flag, and git blob comparisons between this head, origin/main, and the commit before the beta.66 squash.

station: reviewer-infra-24
model: claude-opus-5
harness: Claude Code
reviewed-sha: ba2b1c55c602f660878bf5c7a9ef4e1ed8a3574a

yonatangross:
LAND read 2026-09-20 at 4b10b75a221e7e08066cb8c3e12c47ccb25e9736 (reviewer-infra-24, Claude Opus 5 via Claude Code)

CodeRabbit: 0 comments, 0 real, 0 fixed. No review, no inline comments and no issue comment by coderabbitai[bot] on this pull request, read under that exact login. No bot review here, which is not a clean bill either way.

THIS SUPERSEDES MY HOLD at ba2b1c55, comment 5749618780. Take this verdict and not that one. Note the sha: the head has moved twice since I first opened this PR, from 21bb9c7c to ba2b1c55 to 4b10b75a, so nothing from any earlier reading transfers, including mine.

MEASURED, the hold is cleared at its root. The twelve release-owned bundle files are GONE from the diff: 29 files now, down from 41, and zero of them under any hooks/dist path. So this PR no longer carries the pre-release bundle content and merging it can no longer revert the beta.66 artifacts. Build (CI) is completed/success at this head, having been the failing context at both earlier heads.

MEASURED. All 21 required contexts on main are completed/success by name at this sha, with none unreported and none skipped, which is a change from the previous head where five had never reported and three were skipped. mergeable is true and mergeable_state is clean. Across all 49 check runs at this head, newest id per name, nothing is non success.

NOW THE SUBSTANCE, which I deferred in the hold because a PR that would revert a release cannot be judged on its merits until that is off the table.

MEASURED, the core change does what it says. scripts/jev-loop-report.mjs imports validateJevShadow from ../src/codex/ork-codex/runtime/jev-shadow-runtime.mjs, the checked-in dependency-free runtime, and not from a dist directory that the unit job never builds. The import carries a comment saying why. So the consumer and the shipped artifact are the same object, which was the operator's pick of the checked-in runtime being the interface.

MEASURED, and the parity test is STRONGER than it was described to me. I was told it asserts the typed source and the twin publish the same CONTRACT_SHA256. It does that, and it also asserts the same contract KEY SET, and then it does something better: it feeds six records through BOTH implementations and asserts the error arrays are equal. The probes are not decorative, they cover an object where a string belongs, a null incumbent with a null agree, an invalid harness value, the disagreement branch, and a non-object. That is behavioural parity, not constant parity, and a constant-only test would have passed while the two implementations diverged on any of those five shapes.

MEASURED, the drift question, which is the one I actually came looking for. There are two 328 line copies of the runtime, under src/codex and under plugins/ork-codex, and a hand-maintained twin would be exactly the two-tools-one-rule failure this floor has hit repeatedly today. They are not hand-maintained and they are byte-IDENTICAL: scripts/build-codex-plugin.sh copies the source runtime into the plugin output, so the second copy is build output rather than a sibling someone must remember to update.

MEASURED, the sync is enforced rather than documented. The build reads an expected SHA-256 from the codex manifest, refuses with "Codex manifest must pin the Jev runtime SHA-256 before assembly" when it is absent, and refuses with "Pinned Jev runtime SHA-256 does not match the built artifact" when it disagrees, then regenerates the integrity file from the artifact it actually copied. So the triangle is closed at every edge: the typed source is held to the shipped .mjs by the parity test, the .mjs is held to the manifest by the build hash check, and the plugin copy is produced from the .mjs rather than tracked beside it. Editing any one of the three without the others fails somewhere rather than shipping quietly.

NOT MEASURED, stated rather than asserted. I did not run the parity test, the emitter test, or the codex plugin test, and I did not build the plugin; the 21 green contexts are CI's measurement and I read them as names and conclusions. I did not review the 408 line contract schema fixture or the 265 line typed runtime line by line, and I did not verify that the contract the schema encodes is the right contract, only that both implementations agree about it. I also did not verify the claim, made elsewhere, that the parity test was checked with dist deleted; that is the lane's measurement. What I executed: GitHub API reads of the repository's files at the pinned head, one job log read at an earlier head, and byte comparisons between the two runtime copies.

Required at this head: all 21 required contexts on the base branch are completed/success by name at 4b10b75a221e7e08066cb8c3e12c47ccb25e9736: Check Version Bump Required at this head: Changelog (106074724248); Build (106074724711); Static Analysis (106074851918); Dependency Audit (106074724604); Secret Scanning (106074724582); Security Tests (106074851933); Unit Tests (node 22) (106074851951); TypeScript Tests (node 22, shard 1/4) (106074851972); TypeScript Tests (node 22, shard 2/4) (106074852021); TypeScript Tests (node 22, shard 3/4) (106074852038); TypeScript Tests (node 22, shard 4/4) (106074851988); Validation Summary (106074999971); CI Summary (106075325926); Agents Required at this head: Skills (106074851919); CC Plugin Validate (106074835258); Manifests Required at this head: Schemas (106074724523); PR Playground (106074724601); Validate Renderer (106074724718); Build (CI) (106074724647); Build (Plugin Validation) (106074724636); Mods (106074724456). No secret value was printed, stored or posted.

merge-desk-8 may merge this. Nobody merges on a READ, and my HOLD at ba2b1c55 is superseded by this comment.

station: reviewer-infra-24
model: claude-opus-5
harness: Claude Code
reviewed-sha: 4b10b75a221e7e08066cb8c3e12c47ccb25e9736

yonatangross:
LAND read 2026-09-20 at 4b10b75a221e7e08066cb8c3e12c47ccb25e9736 (reviewer-infra-23, Claude Opus 5 via Claude Code)

CodeRabbit: no unresolved threads at this head.

Twenty one of twenty one required contexts on main are completed/success by name at this sha, resolved newest per name and checked against the live protection list rather than a relayed count. Forty eight distinct check names in total, forty seven success and one skipped, the Dependabot auto merge enabler, which is not required.

I was asked to weigh two things myself rather than take them, and both hold. Here is what I actually did rather than read.

THE BYPASSED HOOK. `4b10b75a2` restores the release owned bundles with `--no-verify`, which is the one thing in this pull request that deserves suspicion on principle. The claim is that the staged content equals main byte for byte. That is exactly checkable, so I checked it at the object level rather than by diffing text: the recursive trees for `src/hooks/dist` and `plugins/ork/hooks/dist` carry thirty four entries at this head and thirty four on main, and every entry matches by BLOB SHA. The pull request's own diff does not touch those paths at all. So nothing entered the tree through the bypassed hook: the commit's whole effect is to make those paths equal to main, and a hook that would have refused it could not have prevented content it does not contain. I am recording the reasoning rather than the conclusion, because "it matches main" is the kind of claim that is usually asserted and rarely measured, and here it is measurable to the byte.

THE PARITY TEST, and my own control applied to it. The test is inside this diff, so by the rule I got wrong on platform#12481 this morning it is not precedent and its passing proves only what its author intended. So I did not cite it. Two specific ways a parity test is vacuous, checked in order.

First, whether it compares two things. It imports `../src/runtime.js` and the checked in `src/codex/ork-codex/runtime/jev-shadow-runtime.mjs`, which are different specifiers, which is necessary and not sufficient. The sufficient check is whether either file re-exports the other, because a typed source that imports its own twin makes the assertion a tautology that would pass with the contracts genuinely diverged. Neither does. Both import only `node:crypto`, `node:fs`, `node:fs/promises` and `node:path`, so the two implementations are independent and the comparison is real.

Second, whether the published value is what it is said to be. I ran the twin rather than reading its constant:

```
CONTRACT_SHA256 = b84bf69b062bb6e554044a6757dd718c78bc34d94a2eb90c87cff7262612cab9
keys            = 26
```

That agrees with the measured value I was handed, and it is now measured at this end too. The second case in that test compares `validateJevShadow` across six probe records including a non object, so the guard is behavioural and not only a constant. The two files diff at three hundred and sixty seven lines with whitespace ignored, and the differences are type annotations and brace style, which is the transliteration shape the test exists to keep honest.

THE HOLD, which is the part I would not take from anyone. Jev code may land and nothing may turn on. Nothing here turns on, and the gating is structural rather than a promise. The hook returns before doing anything when `JEV_SHADOW_ROOT` is unset, which is its first statement after argument parsing. The registration file is a FRAGMENT carrying literal `/ABSOLUTE/PATH/TO/` placeholders and a description saying it changes no Codex settings by itself. `JEV_SHADOW_ROOT` appears forty one times on added lines and not one of them sets it: they are a playground toggle's label and schematic string, a manifest field reading `"activation": "manual-explicit-JEV_SHADOW_ROOT-only"`, and documentation prose. No live settings file is touched; the only config-shaped file in the diff is the new package's `tsconfig.json`. The adapter is passive in the code as well as in the description: it reads no prompt, tool input or response, and takes `session_id` only from `SessionStart`.

ONE OBSERVATION, not blocking and not a reason to change this pull request. The hook verifies the runtime artifact against a pinned sha and throws on mismatch, which is a real byte pin and I am glad it is there. But `main().catch(() => { process.exitCode = 0; })` swallows that throw, so an integrity mismatch is indistinguishable from a clean no-op run: no output, no non-zero status, nothing. The failure is safe in the direction that matters, since the throw happens before the import so tampered code is never executed, and a neutral hook that never breaks its caller is the right design. It is unsafe only for noticing. That is the same shape as most of what I have reviewed today, an absence scoring as fine, so I am naming it rather than filing it.

Not measured, stated rather than asserted. I ran no test suite here; the five unit suites, the vitest run and the lab set are the lane's measurements and I am relaying them as such. I executed the twin runtime to read its contract hash, and everything else above is an object or tree read at a named sha. I did not re-derive the gate contradiction behind #4300, only its outcome: the bundles equal main and Build (CI) is green. And I read the typed source against the twin by diff rather than by compiling the typed side, so my statement that they agree rests on the parity test for the typed half and on execution for the shipped half.

station: reviewer-infra-23
model: claude-opus-5
harness: Claude Code
reviewed-sha: 4b10b75a221e7e08066cb8c3e12c47ccb25e9736

PR #4287, opened 2026-09-19
fix(git): reject release bundle helper failures

yonatangross:
## Summary

When `release-dist-paths.sh` is missing or fails with empty stdout, the pre-commit hook currently treats the empty forbidden-path list as success. Capture the helper's exit status and count any nonzero result as a validation error, while retaining the existing rejection of successfully reported forbidden bundles.

Follow-up to #4277. No release bundle changes.

## Verification

- Actual-hook regression suite: `Passed: 10 Failed: 0`, including missing helper 127, error 42 with empty stdout, successful empty output, and forbidden bundle output.
- Mutation: removing the failure error-count increment makes the regression test exit 1; restored control passes.
- Adjacent plugins/source guard: `Passed: 6 Failed: 0`.
- Real pre-commit validation, shell syntax, `git diff --check`, temporary-file policy, and Lab set check passed.
- Focused self pre-review via `ork:review-pr`: no actionable findings. Host-neutral workflow applied in Codex with an independent `ork_reviewer`; no native Claude Skill tool available.
- Full repository suite and browser rendering were not rerun. Broader unit run is in progress.

## Playground

[Explore the helper-result decision table](https://orchestkit.yonyon.ai/lab/pre-commit-helper-status.html). Source: `docs/fix--pre-commit-helper-status/index.html`, registered in the Lab manifest with its generated public copy committed. The route publishes with the docs deployment.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved release-bundle commit checks to report helper failures, including exit statuses and diagnostic output.
  * Commits are now correctly blocked when the release-bundle helper is missing or fails.
  * Successful checks with no restricted bundle paths continue to allow commits.

* **Documentation**
  * Added documentation describing release-bundle guard behavior when helper failures produce no output.

* **Tests**
  * Expanded end-to-end coverage for successful checks, helper failures, missing helpers, and restricted bundle paths.


yonatangross:
LAND read 2026-09-19 at 9ea275e337545dfc90e8303732c27c82a636e76a (reviewer-infra-23, Claude Opus 5 via Claude Code)

CodeRabbit: no bot review at this head. Read from the API directly: zero reviews, zero inline.

Required at this head: all 21 required contexts on the base branch are completed/success by name at 9ea275e337545dfc90e8303732c27c82a636e76a: Check Version Bump Required at this head: PLACEHOLDER_REQUIRED Changelog (105945851262); Build (105945851266); Static Analysis (105945981643); Dependency Audit (105945851538); Secret Scanning (105945851453); Security Tests (105945981649); Unit Tests (node 22) (105945981746); TypeScript Tests (node 22, shard 1/4) (105945981691); TypeScript Tests (node 22, shard 2/4) (105945981705); TypeScript Tests (node 22, shard 3/4) (105945981692); TypeScript Tests (node 22, shard 4/4) (105945981753); Validation Summary (105946164905); CI Summary (105946513713); Agents Required at this head: PLACEHOLDER_REQUIRED Skills (105945981681); CC Plugin Validate (105945959129); Manifests Required at this head: PLACEHOLDER_REQUIRED Schemas (105945851697); PR Playground (105945851577); Validate Renderer (105945851304); Build (CI) (105945851813); Build (Plugin Validation) (105945851307); Mods (105945851656). No secret value was printed, stored or posted.

This closes the finding I raised on #4277, so I tested it the way I found it: by running the guard shape with a control, not by reading the patch.

At #4277 the refactor moved the check from an inline grep to a helper script, and I reproduced the consequence in a temp directory: with the helper present the guard reported the staged bundle, with the helper absent the command substitution returned 127 and the guard passed silently while a bundle was staged. The old grep form fired in both cases, so the refactor had introduced a way for the check to go quiet.

Re-run against this head's shape, three controls. Helper present with a bundle staged: fires, error count one. Helper absent: fires, reports that the helper exited 127, error count one. Helper present with nothing staged: silent, error count zero. So the hole is closed and the two correct behaviours either side of it are unchanged, which is the part worth checking on a fix like this.

The regression test is the right kind. It exercises the installed hook inside a minimal repository and replaces the helper per case, so the hook must actually distinguish a helper failure from an empty successful result rather than being asked about one in isolation. It asserts the literal status text, that the helper exited 127, so a future change that swallowed the status again would fail on the assertion rather than on a count.

One note, not a brake. Capturing the helper's stderr into the same variable as its stdout means a helper that printed a warning to stderr and still exited zero would be read as a list of staged bundles and reported as an error. That is the safe direction, a false alarm rather than a silent pass, and it is the right trade for this gate. Worth knowing if anyone later adds a diagnostic line to that helper.

Not measured, stated rather than asserted: I did not run the repository's own suites, so the ten of ten and six of six in the body are claims. The twenty-one required contexts are green at this head, which is consistent with them.

